PR #4, opened 2018-10-05
License should not be empty

ildyria:


strub:
@vbgl Is GPL2 the license we want?

vbgl:
Indeed, GPL is wrong. Now that we got rid of CompCert modules, we can elect the license we want.

Why don’t we also get rid of this `opam` file? Does the Jasmin team supports defective by design software development?

strub:
I use opam and find nix utterly slow on my computer. So let’s keep this
file that doesn’t interfere we other systems.

Regarding the licence, I trust your choice. Please, propose the one your
prefer.

Le ven. 5 oct. 2018 à 19:09, Vincent Laporte <notifications@github.com> a
écrit :

vbgl:
What about CeCILL-B?

strub:
Fine by me

PR #10, opened 2020-12-11
Add missing dependency to default.nix

jangilcher:
default.nix is missing a dependency 'yojson' to successfully build the compiler via make in the nix-shell


vbgl:
Thanks!

PR #21, opened 2021-05-06
New AVX2 instructions

MQuaresma:
- Add 10 new AVX2 instructions w/ Coq semantics and unit tests
**Notes**: The `VMOVLPD` and `VMOVHPD` instructions are only partially implemented (**i.e.** no support for memory to register operations)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,11 @@ DESTDIR  ?=
 PREFIX   ?= /usr/local
 BINDIR   := $(PREFIX)/bin
 LIBDIR   := $(PREFIX)/lib
-INSTALL  ?= install
-
+ifeq ($(UNAME_S),Darwin)
+  INSTALL  ?= ginstall
+else
+  INSTALL  ?= install
+endif
 # --------------------------------------------------------------------
 .PHONY: all build check clean dist distcheck
 
--- a/compiler/tests/success/pext.jazz
+++ b/compiler/tests/success/pext.jazz
@@ -1,0 +1,21 @@
+export
+fn test_pext32(reg u64 tmp)
+{
+  reg u32 a b c;
+
+  a = 0xFF;
+  b = 0x02;
+  c = #PEXT_32(a, b);
+  (u32)[tmp] = c;
+}
+
+export
+fn test_pext64(reg u64 tmp)
+{
+  reg u64 a b c;
+    
+  a = 0xFF;
+  b = 0x02;
+  c = #PEXT_64(a, b);
+  (u64)[tmp] = c;
+}
--- a/compiler/tests/success/pocnt.jazz
+++ b/compiler/tests/success/pocnt.jazz
@@ -1,0 +1,11 @@
+export
+fn test_popcnt16(){
+}
+
+export
+fn test_popcnt32(){
+}
+
+export
+fn test_popcnt64(){
+}
--- a/compiler/tests/success/popcnt.jazz
+++ b/compiler/tests/success/popcnt.jazz
@@ -1,0 +1,32 @@
+export
+fn test_popcnt16(reg u64 rp){
+  reg bool zf;
+  reg u16 t;
+
+  t = 0x01;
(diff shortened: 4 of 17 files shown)


vbgl:
Hi Miguel. Thanks for your hard work.

This PR adds many things, some of them are very nice and can be added as-is, some might need some cleaning, some deserve some discussion.

To start with the easy part, I gathered the changes related to the `POPCNT` instruction and did some cosmetic changes (expand the test case, fix the semantics — can you spot the change ? —, use the `of` syntax rather that the ugly back-tick when defining `Variant` datatypes). This resulted in that commit: 59e847f1016d372116c93b2fd2ec8f0143324d7e. Please double check that you’re happy with it and I’ll merge it into `master` and `glob_array3`. Then I’ll review more of this PR.

MQuaresma:
Hi Vincent.

Yes, I can see the changes in the test case and the semantics (somehow I missed that last part when rebasing the commit). Besides that, everything seems good with the commit!

vbgl:
Thanks for the feed-back. The `POPCNT` instruction is now available in `master` and `glob_array3`.

vbgl:
Please check the commit about VPERMD: 973c2689a1080cbc190b6eb4f1235b5d201df6e2
I’ve removed the `wsize` argument and the ability to add a size suffix (in the concrete syntax), as this instruction is only available for 256-bit vectors.
Also, I’ve amended the semantics: I put a “mod 8” where there was a “and 15”. I think its easier to reason about modulo than bit-level operations, and the old semantics was keeping one extra bit.

MQuaresma:
Commit 973c268 looks good go me too!

vbgl:
Thanks! Next in the queue: bf366d6ab229538554dbb06b02407668bc2e20af

MQuaresma:
bf366d6 looks good too!

vbgl:
Thanks. Next one is for VPCMPEQ & VPCMPGT: 1ced643cfdf209adf72576e062a7674fd0b54d39

MQuaresma:
1ced643 looks good too! The semantics are easier to understand like this!

vbgl:
Thanks again. VPCMPEQ and VPCMPGT are now in master & glob-array3.

Next commit is for RDTSC and RDTSCP: a54493682589a8887d49bd500337b08bd7e5d37c
According to our previous discussions, I’ve changed the semantics to be “assert false”. The compiler can be trusted to correctly compile programs using these instructions but no proof at all can be done on such programs.

vbgl:
In addition to the RDTSC commit (rebased on current master as 73819db03d04a7106731a71e36f91398faa917ff), there is the addition of PEXT: 61ad0fc2e898794ae6286ebcc3b1a672a0060d73 I’ve removed the `bits2z` function and replaced its use by `t2w ∘ in_tuple`; I think this is correct.

MQuaresma:
Both a544936 and 61ad0fc look correct!

vbgl:
Please double check commit 7d5028ed2ce871cfb9a83a33f00083179bc67380 that adds VPMADDUBSW and VPMADDWD. I’ve rewritten parts of the semantics. The documentation for VPMADDWD does not refer to saturation: it’s probably not relevant because no overflow can happen in the signed/signed case (right?).

Regarding the two last instructions mentioned in this PR: VMOVHPD and VMOVLPD, I’m a bit confused about what to do. These instructions have two forms: one for load, one for store. It seems that it would make sense to represent the two forms as two different instructions.

vbgl:
The VMOVHPD and VMOVLPD instructions have been added in f0595f3ec53b2b8e0d32b5556efc4599d64f32e4. It’s only the “store” variant (one source, one destination), 128-bit to 64-bit. The “load-and-merge” variants are left as future work.

If I’m not mistaken, all changes proposed in this PR have been incorporated into `master` (and `glob_array3`). I’ll thus merge this PR.

Thanks for your contribution!

PR #40, opened 2021-07-05
Proof of the new stack alloc pass

eponier:
Here it is!

Apart from the main changes (in stack_alloc.v, stack_alloc_proof.v and stack_alloc_proof_2.v), I pushed changes in common files (psem.v, memory_model.v, ...). These changes are probably more important to review than the main ones.

Here are a few questions and points to closely look at:
- there are two files stack_alloc_proof.v and stack_alloc_proof_2.v only because the initial file was becoming too big (cf. paragraph at the top of stack_alloc_proof_2.v). Should we keep two different files or should I merge them?
- I performed a renaming `pointer` -> `ptr` for the sake of uniformity with `Uptr` and `sptr` (introduced in stack_alloc_proof.v because I don't know where to put it). But is it worth the trouble?
- Among the changes I made to memory_model.v, I proved `top_stack_after_aligned_alloc` instead of assuming it and proving it in `memory_example.v` because it turned out the definition of `top_stack_after_alloc` was precise enough to make the proof directly. I also moved `top_stack_after_alloc_bounded` from memory_example.v to memory_model.v. Actually, I wonder whether there are other parts that could be proved instead of assumed (`ass_above_limit`, `ass_fresh`)?
- I added a `-1` to `alloc_stack_complete` and `top_stack_after_alloc_bounded` because this is true and more precise.
- there are now two proofs called `writeP_neq` (in memory_model.v and low_memory.v). The one in low_memory.v should be easier to use (and shadows the other one). File linear_proof.v was using `writeP_neq` from memory_model.v. I chose the simple solution of using the longer name `CoreMem.writeP_neq` to still use this one, but maybe the right fix would be to use the other `writeP_neq`.
- I changed the priority of some type class instances, so that they do not hide more natural instances.

I can of course explain some points if some choices seem questionable (actually, until the end of week, because I'm going on vacation on Saturday). But this is not urgent, so take the time needed to review this. It can wait until I am back from vacation if needed. I have already rebased this PR a few times, so I can do it again, the effort is small.

### DIFF
--- a/compiler/src/stackAlloc.ml
+++ b/compiler/src/stackAlloc.ml
@@ -62,7 +62,7 @@ let memory_analysis pp_comp_ferr ~debug tbl up =
   let cget_sao fn = get_sao (Conv.fun_of_cfun tbl fn) in
   
   let sp' = 
-    match Stack_alloc.alloc_prog crip gao.gao_data cglobs cget_sao up with
+    match Stack_alloc.alloc_prog false crip gao.gao_data cglobs cget_sao up with
     | Utils0.Ok sp -> sp 
     | Utils0.Error e ->
       Utils.hierror "compilation error %a@." (pp_comp_ferr tbl) e in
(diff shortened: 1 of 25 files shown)


eponier:
Remark: I squashed all my commits because the history was really dirty (lots of commits with meaningless titles in particular).

eponier:
Ok, I think I reverted all `pointer` -> `ptr` changes, and turned `sptr` into `spointer` (only a few occurrences actually). Should be ready.

vbgl (approved):

PR #41, opened 2021-07-08
Do not export global variables

vbgl:
Currently, Jasmin programs that use global variables cannot be dynamically linked (at least on Linux with GCC and GNU binutils). I don’t understand all the details, but the linker complains: “relocation R_X86_64_PC32 against symbol `g' can not be used when making a shared object”.

Maybe we could allow global data to be tagged as `export` to get the current behavior (name is accessible from other compilation units, but only static linking is possible).


eponier:
Hum, after a very quick research, this is linked with option `-fPIC` that we have to pass to `gcc` when compiling a shared library. I have never understood the details either, but when trying to produce a shared library, `gcc` itself asks to add option `-fPIC` si I always add it (I admit I do not understand why it is on by default if we have to use it systematically).

PR #45, opened 2021-07-30
Clean EC semantic for VPMOVMSKB instruction

MQuaresma:
Simplify `VPMOVMSKB` semantic in EC

### DIFF
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -374,13 +374,15 @@ abbrev [-printing] VPBLEND_16u16 = VPBLENDW_256.
 abbrev [-printing] VPBLEND_8u32 = VPBLENDD_256.
 
 (* ------------------------------------------------------------------- *)
+op msb (b: W8.t): bool = 128 <= (to_uint b).
+
 op VPMOVMSKB_128 (v: W128.t) : W16.t =
-  let vb = w2bits v in
-  W16.bits2w (mkseq (fun i => nth false vb (8*i + 7)) 16).
+  let vb = W16u8.to_list v in
+  W16.bits2w (map msb vb).
 
 op VPMOVMSKB_256 (v: W256.t) : W32.t =
-  let vb = w2bits v in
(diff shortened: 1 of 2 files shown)


eclib/JModel.ec
```diff
@@ -374,13 +374,15 @@ abbrev [-printing] VPBLEND_16u16 = VPBLENDW_256.
 abbrev [-printing] VPBLEND_8u32 = VPBLENDD_256.
 
 (* ------------------------------------------------------------------- *)
+op msb (b: W8.t): bool = 128 <= (to_uint b).
```
vbgl, line 377: I’m surprised that there is no such function yet in the `W8` theory. There is `ALU.SF_of` that has a weird name but seems to be what you need. It might be worth to define a general `msb` function it the `BitWord` theory.

MQuaresma: Yes, I was surprised too,  Didn't think of using `ALU.SF_of` though, but long term it may make more sense to add the `msb` function to the `BitWord` theory. I can include that in this PR...

vbgl: That would be nice, thanks.

eclib/JWord.ec
```diff
@@ -874,6 +874,8 @@ abbrev (`&`) = andw.
 abbrev (`|`) = orw.
 abbrev (`^`) = (+^).
 
+op msb (b: t): bool = modulus <= (to_uint b).
```
vbgl, line 877: Are you sure its `modulus` and not its half? You may also test if the signed interpretation is negative, or simply access the most significant bit (`b.[size - 1]`).

MQuaresma: You're right, I messed that up somehow... 
I went with the unsigned comparison, at least in the case of the Kyber proof, this make it easier to reason about the behaviour of the code. 
Maybe it makes sense to add some auxiliary lemmas that state the other properties you mentioned (i.e. if the `msb` is set, then the signed representation is negative or that the result of `msb` is equal to that of indexing the most significant bit directly). I can include those here too

PR #46, opened 2021-08-04
Prove to_sint bounds

MQuaresma:
Add `to_sint_cmp` lemma to define the bounds for the `to_sint` operator

### DIFF
--- a/eclib/JWord.ec
+++ b/eclib/JWord.ec
@@ -101,6 +101,11 @@ proof.
   by rewrite to_uintE bs2int_ge0 -(size_w2bits x) bs2int_le2Xs.
 qed.
 
+lemma to_sint_cmp (x : t) : min_sint <= to_sint x <= max_sint.
+proof.
+ by rewrite /to_sint /= /smod /=; smt.
+qed.
+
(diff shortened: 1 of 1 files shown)


eclib/JWord.ec
```diff
@@ -101,6 +101,11 @@ proof.
   by rewrite to_uintE bs2int_ge0 -(size_w2bits x) bs2int_le2Xs.
 qed.
 
+lemma to_sint_cmp (x : t) : min_sint <= to_sint x <= max_sint.
```
vbgl, line 104: I’m not very familiar with naming conventions but I could find the following lemmas in EasyCrypt theorems: [rg_log](https://github.com/EasyCrypt/easycrypt/blob/512f0f2bd8dc8c1132f3cd138a17d57d4db9c514/theories/algebra/Group.ec#L393) and [rg_asint](https://github.com/EasyCrypt/easycrypt/blob/512f0f2bd8dc8c1132f3cd138a17d57d4db9c514/theories/algebra/ZModP.ec#L43).

MQuaresma: Sure, I can do that, so it would become `rg_sint`? The naming here was based on the same lemma for the `to_uint` operator, which is called `to_uint_cmp`.

Issue #15, opened 2021-03-17
Assertion failed when calling a function with an array of wrong length


eponier:
When compiling the following program in the `glob_array3` branch
```
inline fn sum (stack u64[2] a) -> reg u64 {
  reg u64 r; 
  
  r = a[0];
  r += a[1];
    
  return r;
} 
  
export fn main (reg u64 x) -> reg u64 {
  reg u8 r;
  stack u64[4] a;
  reg u64 res;
  inline int i;
  
  r = 0;
  for i = 0 to 4 {
    a[i] = 3;
  } 
  
  res = sum(a);
  return res;
} 
```
we first get a warning `can not ensure that the type U64[4] is compatible with U64[2]` that is not fatal, and then `Fatal error: exception File "src/alias.ml", line 122, characters 2-8: Assertion failed`. The error is expected but could be clearer.

Note that if we try a more naive error:
```
stack u64[4] a;
stack u64[2] b;

a = b; // or b = a;
```
we get the same warning, but then a typing error. Maybe we should get the same typing error in the first case.

I must admit that the non-fatal warning, followed in both cases by a fatal error due to the same problem, is rather confusing. But maybe there are cases where the warning is triggered and the compilation still possible?

eponier:
@vbgl

vbgl:
What is this program supposed to mean? Is it ill-typed (in which case the warning should be an error)? Or does the `sum` function only receives the first half of the array `a` (in which case the compiler may implicitly introduce a slicing operation, i.e., change the call into `sum(a[0:2])`)?

eponier:
I would say that it is indeed ill-typed. If the user wants to pass only the beginning of `a`, it can write explicitly `sum(a[0:2])` that is only 5 characters longer. And if it is not done on purpose, it receives a proper error from the compiler. But you may known use cases where the automatic slicing introduction would be interesting.

eponier:
We can get the same error when the problem with non-equal types occurs in the result, e.g.
```
fn f(reg ptr u64[4] x) -> reg ptr u64[2] {
  return x;
}

export fn main (reg u64 x) -> reg u64 {
  stack u64[4] a;
  reg ptr u64[2] b;
  reg u64 res;

  b = f (a);
  res = b[0];

  return res;
}
```

eponier:
Benjamin suggested to be stricter in the typing pass. If we change the `subtype` function to accept only arrays of the same size, then we have a proper error.

eponier:
[...]

Issue #35, opened 2021-06-10
makeReferenceArguments: bad fresh id


eponier:
The program is admittedly ugly, but it still produces an error.

```
fn f(reg ptr u64[10] x) -> reg ptr u64[8] {
  return x;
}

export fn main (reg u64 x) -> reg u64 {
  stack u64[15] a;
  reg u64 res;

  a[0:8] = f (a[0:10]);
  res = a[0];

  return res;
}
```

eponier:
Another example:
```
fn f(reg ptr u64[4] x) -> reg ptr u64[2] {
  return x;
}

export fn main (reg u64 x) -> reg u64 {
  stack u64[4] a;
  stack u64[2] b;
  reg u64 res;

  b = f (a);
  res = b[0];

  return res;
}
```
This relies on `b` declared as `stack`. It does not work if `b` is declared as `reg ptr` (actually, we get another error, cf. #15 ), since then the `makeReferenceArguments` pass does not change the program too much.

eponier:
Working on it.

eponier:
Same comment as in another PR: https://github.com/jasmin-lang/jasmin/issues/15#issuecomment-918100424.

eponier:
Solved by https://github.com/jasmin-lang/jasmin/commit/dd734c1bded81a0bdd51e8816a95297aacac06b5.

PR #61, opened 2021-09-13
Strict array typing

eponier:
Until now, the typechecking of arrays was loose, in the sense that in some places arrays shorter than expected were accepted. But it appears that some compiler passes were not as lenient, leading to rather unexpected compilation failures.

Fixes https://github.com/jasmin-lang/jasmin/issues/15 https://github.com/jasmin-lang/jasmin/issues/35


eponier:
I targetted glob_array3, but maybe such changes should instead target master? Hum, it turns out that the typing was quite different in master, so I'm not sure about that.

eponier:
@bgregoir @vbgl Do we want this in the end?

PR #62, opened 2021-09-13
Arrays in pretyping

eponier:
A strange piece of code in pretyping was making a program fail to typecheck too early (i.e. before param expansion). Moreover it was also rejecting such an example:
```
fn f () -> reg u32 {
  reg u32 r;
  inline int r2;
  r = 17;
  r2 = r;
  return r;
}
```
by mistyping the equality `r2 = r`. So it seems it is really buggy.


eponier:
The patch is included in https://github.com/jasmin-lang/jasmin/commit/4b6609d6313044764a89db98983314f814c803f6.

PR #63, opened 2021-09-15
Pretyping: keep the error when OnlyParam

eponier:
We catch an exception and in some cases reemit a different one. I think it is better not to catch it in these cases (which means the original exception is displayed to the user).

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -582,9 +582,8 @@ let tt_var_global (mode:tt_mode) (env : Env.env) v =
   let lc = v.L.pl_loc in
   let x, k = 
     try tt_var mode env v, E.Slocal
-    with TyError _ -> 
+    with TyError _ when mode <> `OnlyParam ->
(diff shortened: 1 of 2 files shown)


eponier:
I added a test case.
Before:
```
"tests/fail/typing/only_param.jazz", line 3 (10-11):
typing error: unknown variable: `x'
```
After:
"tests/fail/typing/only_param.jazz", line 3 (10-11):
typing error: only param variable are allowed here
```
For me, it is clearer for the user to have the message : "this is a known variable, but you cannot use it here" rather than "variable not known".

**Comment by eponier** [comment]
But anyway, the same problems exist regarding global variables. For instance
```
u64 g = 2;
u64[g] g2 = {1};
```
gives `typing error: unknown variable: `g'`. I would rather expect a message "global variables cannot be used here" or again "only param variable are allowed here". But this is a bigger rework than what I want to do.

vbgl (approved):
Looks like an improvement. Even though the user experience might not be perfect yet.

PR #73, opened 2021-12-02
Add VPTEST and VPBLENDVB

MQuaresma:
Add 2 new AVX2 instructions w/ Coq semantics and unit tests.


bgregoir (changes requested):
Hi, 
   I think you forgot to patch eclib. 
   Can you do that before we merge ?

bgregoir (changes requested):
Same remark. It is good to add the semantic in eclib.

PR #78, opened 2022-02-11
pretyping: improve location information in warnings

### LINKED ISSUE #23: Bad location for a warning
In the following code
```
fn f () -> reg u64 {
  reg u64 r;
  r = 0;
  return r;
}

export fn main (reg u64 x) -> reg u64 {
  reg u64 r;
  f ();
  r = x;
  return r;
}
```
we get the warning `warning: at line -1, introduce 1 _ lvalues` (instead of line 9). This seems to be due to the absence of assignment. Indeed, in this other example
```
fn f () -> reg u64, reg u64 {
  reg u64 r;
  r = 0;
  return r, r;
}

export fn main (reg u64 x) -> reg u64 {
  reg u64 r;
  r = f ();
  return r;
}
```
we correctly get a warning mentioning line 9.

vbgl:
Fixes #23

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -10,8 +10,8 @@ module W = Wsize
 module T = Type
 
 (* -------------------------------------------------------------------- *)
-let loc_of_tuples locs =
-  List.fold_left L.merge L._dummy locs
+let loc_of_tuples default locs =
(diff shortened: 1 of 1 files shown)


eponier:
The fix seems correct to me: when there are no result variables, we use the location of the instruction.

PR #82, opened 2022-02-23
Reject programs that attempt to write a constant pointer during typing

### LINKED ISSUE #81: Reject modifications to constant pointers rather than ignoring them
The below program allocates a variable on the stack, sets it, and passes a pointer to it to a subprocedure, which modifies the pointer contents but does not return it. I would expect this program to be rejected with an error; instead the modification gets eliminated silently and a valid assembly program is emitted.

```
fn foo (reg ptr u64[1] px) {
  px[0] = 1;
}

  stack u64[1] x;
  reg u64 res;
  x[0] = 41;
  foo(x);
  res = x[0];
  return res;
}
```

The mutation of the stack seems to disappear in the `unrolling` pass:
```
(* After remove unused function *)


fn foo (Reg const ptr u64[1] px.136) -> () {
  
  px.136[u64 0 ] =(u64) ((64u) 1);
  return ();
}

fn main () -> (Reg u64) {
  Reg u64 res.134;
  Stack u64[1] x.135
  x.135 =(u8[8]) array_init(8);
  x.135[u64 0 ] =(u64) ((64u) 41);
   () = foo(x.135);
  res.134 =(u64) x.135[u64 0 ];
  return (res.134);
}


(* After unrolling *)


fn foo (Reg const ptr u64[1] px.136) -> () {
  
  
  return ();
}

fn main () -> (Reg u64) {
  Reg u64 res.134;
  Stack u64[1] x.135
  x.135 =(u8[8]) array_init(8);
  x.135[u64 0 ] =(u64) ((64u) 41);
   () = foo(x.135);
  res.134 =(u64) x.135[u64 0 ];
  return (res.134);
}


```

haselwarter:
Pointers that aren't returned are inferred to be constant; any attempt at
writing to such a pointer should be considered an error.

Closes #81.

### DIFF
--- a/compiler/tests/fail/typing/write_constant_pointer_subproc_array.jazz
+++ b/compiler/tests/fail/typing/write_constant_pointer_subproc_array.jazz
@@ -1,0 +1,17 @@
+fn foo (reg ptr u64[3] px) -> reg u64 {
+  px[0] = 1;
+  px[1] = 0;
+  px[2] = 8;
+  reg u64 res_x;
+  res_x = px[0];
+  return res_x;
+}
+
+export fn main () -> reg u64 {
+  stack u64[3] x;
+  reg u64 res;
+  x[0] = 41;
+  _ = foo(x);
+  res = x[0];
+  return res;
+}
--- a/compiler/tests/fail/typing/write_constant_pointer_subproc_var.jazz
+++ b/compiler/tests/fail/typing/write_constant_pointer_subproc_var.jazz
@@ -1,0 +1,17 @@
+fn foo (reg ptr u64[3] px) -> reg u64 {
+  reg u64 res_x;
+  stack u64[3] y;
+  px = y;
+  y[0] = 666;
+  res_x = px[0];
+  return res_x;
+}
+
+export fn main () -> reg u64 {
+  stack u64[3] x;
+  reg u64 res;
+  x[0] = 41;
+  _ = foo(x);
+  res = x[0];
+  return res;
+}
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -51,6 +51,7 @@ type tyerror =
   | PtrOnlyForArray
   | ArgumentNotVar      
   | BadVariableKind     of P.v_kind
+  | WriteToConstantPointer of S.symbol
   | PackSigned
   | PackWrongWS of int
   | PackWrongPE of int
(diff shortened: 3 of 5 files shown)


vbgl:
The first commit looks good. Indeed, it is wrong to write in a an array that is marked `const ptr`.

However, I do not understand the second commit. Here is a more straightforward test case:

~~~
export
fn main() -> reg u64{
  reg u64 a;
  stack u64[1] s;
  reg const ptr u64[1] p;
  p = s;
  p[0] = 1;
  a = s[0];
  return a;
}
~~~

(Note that I’ve just fixed the code of `test_ptr_fn_inline.jazz`)

haselwarter:
I wasn't aware of `const` as a keyword, and forced the typer to infer the pointers as `const` by passing them to a subprocedure without returning them. There should still be two tests, one for each path leading to the error. Actually, I'd keep my tests as well to ensure that `const` inference behaves as intended.

(I wasn't sure what the intention of `test_ptr_fn_inline.jazz` was, I'll rebase and leave it untouched)

compiler/tests/fail/typing/write_constant_pointer_subproc_var.jazz
```diff
@@ -0,0 +1,17 @@
...
+  y[0] = 666;
+  res_x = px[0];
+  return res_x;
+}
+
+export fn main () -> reg u64 {
+  stack u64[3] x;
+  reg u64 res;
+  x[0] = 41;
+  foo(x);
```
vbgl, line 14:
```suggestion
  _ = foo(x);
```

compiler/tests/fail/typing/write_constant_pointer_subproc_array.jazz
```diff
@@ -0,0 +1,17 @@
...
+  reg u64 res_x;
+  res_x = px[0];
+  return res_x;
+}
+
+export fn main () -> reg u64 {
+  stack u64[3] x;
+  reg u64 res;
+  x[0] = 41;
+  foo(x);
```
vbgl, line 14:
```suggestion
  _ = foo(x);
```

compiler/tests/fail/typing/write_constant_pointer_subproc_var.jazz
haselwarter: Done in  28c42a7, thanks.

compiler/tests/fail/typing/write_constant_pointer_subproc_array.jazz
haselwarter: Done in  28c42a7, thanks.

bgregoir:
I think the file Typing need also to be patch.
By the way, I don't no if this is taken into account in Typing.

PR #83, opened 2022-02-24
Pretty-print intermediate programs in correct Jasmin syntax

vbgl:
This PR tries to ensure that Jasmin programs printed by the compiler are indeed written in Jasmin syntax. In particular, the output of `-pcstexp` is always syntactically valid (and usually well-typed); this is now checked by the test-suite.

This is a breaking change (in particular printed programs are now displayed on `stdout` instead of `stderr`).

Feature requested by Tiago (@tfaoliveira).

### DIFF
--- a/compiler/src/printer.ml
+++ b/compiler/src/printer.ml
@@ -94,7 +94,7 @@ let string_of_op2 = function
 
 
 let string_of_op1 = function
-  | E.Oint_of_word sz -> F.sprintf "(int of u%d)" (int_of_ws sz)
+  | E.Oint_of_word sz -> F.sprintf "(int /* of u%d */)" (int_of_ws sz)
   | E.Osignext (szo, _) -> F.sprintf "(%ds)" (int_of_ws szo)
   | E.Oword_of_int szo
   | E.Ozeroext (szo, _) -> F.sprintf "(%du)" (int_of_ws szo)
@@ -130,7 +130,7 @@ let pp_ge pp_var =
 
 (* -------------------------------------------------------------------- *)
 let pp_glv pp_var fmt = function
-  | Lnone (_, ty) -> F.fprintf fmt "_{%a}" (pp_gtype (fun fmt _ -> F.fprintf fmt "?")) ty
+  | Lnone (_, ty) -> F.fprintf fmt "_ /* %a */" (pp_gtype (fun fmt _ -> F.fprintf fmt "?")) ty
   | Lvar x  -> pp_gvar_i pp_var fmt x
   | Lmem (ws, x, e) ->
     F.fprintf fmt "@[(%a)[%a@ +@ %a]@]"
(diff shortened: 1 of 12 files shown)


eponier:
I'm rebasing and then I'll test a bit.

eponier:
I'm writing down the problems I see:
- some "export" qualifiers seem to be missing
- param is printed as const

eponier:
I may be wrong, but if I recall correctly, Tiago wants to print the code just after parsing, with no change at all from the compiler. Maybe we should print "ast"?

tfaoliveira:
The idea would be to *print* the input program after `require`s are solved (similar to the option `-E` from `gcc` or `clang`, " -E  Stop after the preprocessing stage"). This is useful, for instance, to create single-file Jasmin implementations that can later be pruned for the purpose of recreating a particular behavior (essentially, debugging/sharing code). 

I cloned this branch to give it a try but I noticed that the `require` feature does not seem to be supported on it. It will be very useful if this can later be merged in `get-random`.

tfaoliveira:
I think it is the second option. But I will write some comments.
When did you mention `param` to be propagated does that mean that a given `param int  N = 2;` would be statically solved?

The original idea was that given a file `globals.jazz` defined as:
```
param int N = 2;
// possibly some code here
```
and a second file `impl.jazz` defined as:
```
require "globals.jazz"
inline fn sum1(reg u64[N] a) -> reg u64[N]
{ inline int i;
  for i=0 to N
  { a[i] += 1; }
  return a;
}
```
Then `jasminc -preprocessing-option impl.jazz -possibly-preprocessing-output-filename-flag output.jazz` would write in output.jazz the following:
```
param int N = 2;
// possibly some code here
inline fn sum1(reg u64[N] a) -> reg u64[N]
{ inline int i;
  for i=0 to N
  { a[i] += 1; }
  return a;
}
```

eponier:
So yes, it is the second option. And I think it means we should print really just after parsing.

vbgl:
Do you really need the `param` declarations to be preserved? It requires some more work to get this working…

vbgl:
My bad, this is really doable. Stay tuned.

vbgl:
Done. CI seems happy.

eponier:
Do we still need to disable debug for cstexp in the end?

vbgl:
Indeed. I’ve just dropped that commit.

compiler/src/printer.ml
```diff
@@ -243,17 +244,27 @@ let pp_var_decl pp_var pp_size fmt v =
...
+    let seen = ref Spv.empty in
+    let mark x = seen := Spv.add x !seen in
+    let is_seen x = Spv.mem x !seen in
+    List.iter mark fd.f_args;
+    fold_vars_fc (fun x () ->
+        if x.v_kind <> Const && not (is_seen x) then (
+          mark x;
+          F.fprintf fmt "%a;@ " pp_vd x)
+    ) () fd
+  in
```
eponier, line 257: This is the only really non-trivial part of this PR. Could you add a comment explaining why this is needed please?

eponier:
(sorry, playing with the github interface)

compiler/src/printer.ml
vbgl: Currently, the printer does not print the declaration of local variables (which is not correct). This snippet iters on all variables that are used in the function and prints its declaration (at most once, hence the `seen` set) unless it is a function argument (line 251) or a parameter (`v_kind <> Const`, line 253).

eponier: Thanks. Does this mean that the ideal fix is to directly print the local declarations, or is your fix already the right solution?

vbgl: What do you call “directly print the local declarations”? There are no declarations of local variables in Jasmin abstract syntax.

eponier: Oh, I didn't know that. Actually, I had the impression that in some places we have them and in some others we don't, but probably I'm wrong. Do you know why this choice was made? Does it make things simpler?

vbgl: When the set of local variables is needed (like here) we can compute it. Maybe this computation might be memoized?

At least we never worry about the consistency of this set (are all variables declared? are there variables that are declared but not used?).

eponier: Thanks for the explanation!

PR #92, opened 2022-03-25
add easycrypt semantics for BSWAP

fdupress:
BSWAP is used in libjade, but doesn't have easycrypt semantics defined. This adds some semantics.

### DIFF
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -132,6 +132,25 @@ op mulu64 (w1 w2 : W64.t) =
   (W2u32.zeroextu64 (W2u32.truncateu32 w2)).
 
 (* -------------------------------------------------------------------- *)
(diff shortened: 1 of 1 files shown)


fdupress:
Comments on definition style and useful lemmas to add are more than welcome.
(I don't particularly like the point-free style definition, for example.)

@vbgl You added the BSWAP instruction, so might be appropriate as a reviewer?
Someone with experience of dealing with proofs on extracted EC code might be a good second opinion.

vbgl (approved):
Kudos for using `/(\o)` as an argument in your proof!

fdupress:
Force-pushed to remove a spurious newline.

> Kudos for using /(\o) as an argument in your proof!

Thanks. I'm just sorry `\o/` is not a valid infix operator name.

PR #94, opened 2022-03-31
Relax semu condition in instruction descriptors

vbgl:
Monotony w.r.t. `value_uincl` is enough.


bgregoir (commented):
This is a good change. We should try to add more small simplification like this.

bgregoir (approved):

PR #95, opened 2022-04-01
Remove some dead code

vbgl:
Commits are independent but might be squashed together.


eponier (approved):
All the changes are ok, thanks! I wouldn't have changed `lang/word.v` because I find that these lemmas make sense even if they are not used directly. But I don't mind if they are removed. We should upstream most of this file to coqword as suggest in https://formosa-crypto.zulipchat.com/#narrow/stream/308530-Jasmin.3Adevelopment/topic/words anyway.

vbgl:
Alright. I’ve dropped the commit about `word`.

PR #97, opened 2022-04-04
MIT everywhere

eponier:
When I did the change of license, I forgot to update a bunch of places. I'll backport this to release-21.


eponier:
The other change is that coqword is apparently MIT too.

vbgl (approved):

PR #98, opened 2022-04-04
fix Let notation

Alizter:
We recently merged https://github.com/coq/coq/pull/15754 which removed support for `_` in `ident`, after a period of deprecation. Since Jasmin was recently added to the CI, the CI run for that PR missed the fact that the `Let` notation defined here is using ident. This PR changes `ident` to the correct `name`, which allows for underscores.


vbgl:
There are also two notations in `ssrmisc/oseq.v` (lines 115 and 119) that use `ident` and trigger a warning.

Alizter:
Feel free to push to this PR. The main reason for submitting this PR is to get jasmin working in the coq CI again.

vbgl (approved):

PR #102, opened 2022-04-07
Initial changelog

eponier:
I probably forgot some of the features (should we mention the `-lazy-regalloc` option somehow?), please say so if you remember one important thing that is not listed. I wasn't sure how long I should be. I think in theory I should be shorter and give pointers to the documentation, but since most of the doc is missing, I give more details.

The idea would be to merge that just before branching out branch `release-22`, I guess, except if @vbgl tells me this is not the right way to proceed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Here are the main changes of the release.
   declared with `inline fn` and export functions with `export fn`. This is a
   breaking change, since before `fn` was a synonym for `inline fn`. Unlike
   inline functions, they are proper functions. Unlike export functions, they are
-  internal. As such, they do not need to respect any standard calling conventions
+  internal. As such, they do not need to respect any standard calling convention
   and are therefore a bit more flexible.
 
 - **New storage modifiers `reg ptr` and `stack ptr`** to declare arrays.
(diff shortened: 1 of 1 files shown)


eponier:
It is also possible that some points are not detailed enough or inaccurate, I'll take another look tomorrow. Please don't merge too fast.

bgregoir (commented):
nice, we did a lot of work.

CHANGELOG.md
```diff
@@ -0,0 +1,84 @@
...
+being listed below, but also a lot of breaking changes. Please upgrade with
+care.
+
+Here are the main changes of the release.
+- **A new kind of function, subroutines,** in addition to inline functions and
+  export functions. They are declared with `fn` only, while inline functions are
+  declared with `inline fn` and export functions with `export fn`. This is a
+  breaking change, since before `fn` was a synonym for `inline fn`. Unlike
+  inline functions, they are proper functions. Unlike export functions, they are
+  internal. As such, they do not need to respect the calling conventions of C
```
vbgl, line 14:
```suggestion
  internal. As such, they do not need to respect any standard calling conventions
```

CHANGELOG.md
```diff
@@ -0,0 +1,84 @@
...
+  and are therefore a bit more flexible.
+
+- **New storage modifiers `reg ptr` and `stack ptr`** to declare arrays.
+  `reg ptr` is used to store the address of a stack array in a register. The
```
vbgl, line 18:
```suggestion
  `reg ptr` is used to store the address of an array in a register. The
```

vbgl (commented):
Thanks for curating this long list!

vbgl:
There is a new constant-time checker, also.

eponier:
@vbgl can you have a look on the lastest version, please?

eponier:
Thanks for your feedback. I merged your changes and removed the wrong item.

vbgl (approved):
This is way more than a change log: it’s also the documentation of the new features. Thanks!

PR #103, opened 2022-04-08
Improve ct checker

eponier:
The goal is to print the types using valid Jasmin syntax.


compiler/src/ct_checker_forward.ml
```diff
@@ -447,11 +447,11 @@ let get_annot ensure_annot f =
   let check_lvl lvl =
     match lvl with
     | Poly s ->
-      let diff = Svl.diff known s in
```
vbgl, line 450: Can you please show a test case that illustrate the wrong behavior (without this change)?

eponier: Done. It's easy to reproduce, you just need to introduce a security variable in the arguments and not to use it in the return type.

vbgl (approved):

PR #105, opened 2022-04-20
Disable GitHub Actions

vbgl:
There are a few issues that must be fixed

  - The OPAM cache is incomplete and corrupts as time elapses
  - The push-compiler-code job must not run on PRs (and other WIP branches)
  - A single CI system is enough

Currently, all CI pipelines fail because the CI system is broken.


eponier:
I think we can disable the CI without removing those files (either by removing the `on:` line or in the web interface). It depends whether we want to fix these files in the future or if you consider that we'll stick to Gitlab for a long time.

vbgl:
What you suggest is like removing dead code by wrapping it in comments.

Any attempt to fix the workflows can start by reverting this commit.

eponier:
Fair point, but what I would like to understand is whether if GitHub's CI is fixed (cc @maximedenes) you'll be happy switching or if you prefer Gitlab's CI anyway. In the second case, it seems useless to spend time on GitHub actions, and we should rather try to report the results of Gitlab's CI in GitHub PRs (cf. https://github.com/coq/coq/pull/15855#issuecomment-1079288691).

maximedenes:
> * The OPAM cache is incomplete and corrupts as time elapses

Do you have more info on this?

> * The push-compiler-code job must not run on PRs (and other WIP branches)

Ok, I can fix this.

> * A single CI system is enough

Sure, but integration with developers workflow really matters. There are two ways to achieve it. One is by switching to GitHub actions, since the main repo is on GH. The other is by developing and maintaining some glue between Gitlab's CI and GH (like we do for Coq, for historical reasons). Are you saying you prefer the later approach?

maximedenes:
> * The OPAM cache is incomplete and corrupts as time elapses

Did I fail to cache the same thing as what the Gitlab CI pipeline was doing, or do you observe the same problem with Gitlab CI?

vbgl:
> > * The OPAM cache is incomplete and corrupts as time elapses
> 
> Did I fail to cache the same thing as what the Gitlab CI pipeline was doing, or do you observe the same problem with Gitlab CI?

The behavior is the same on gitlab. The difference is that gitlab features a “clear runner cache” button so this issue is just a minor annoyance. On GH, it becomes a blocking bug.

maximedenes:
Ok, so can you clarify your expectations? There are no open issues for these problems AFAICT, just this revert PR. Also, I don't know what "A single CI system is enough" implies.

I'm fine with this PR being merged and forgetting about GH Actions. I'm also fine trying to fix the issues. What I'd like to avoid is wasting my time on it if anyway the third point means you'd like to stick with Gitlab CI.

vbgl:
Just remove `.gitlab-ci.yml` when enabling the GH workflows.

eponier:
Ok, so I think we agree on switching to GitHub actions if the limitations are removed. So let's merge this PR and create issues about the two limitations that were identified.

eponier (approved):
GitHub's CI needs a bit more work to be ready.

PR #107, opened 2022-04-22
Minor change in changelog

eponier:
I realized we are testing with option `-ptyping`, not `-pcstexp` so I changed the Changelog accordingly.


vbgl (approved):
Thanks.

PR #109, opened 2022-04-22
CI: refactor opam job

vbgl:
This should prevent the cache of the OPAM root to rot, as long as it stays in sync with the pinned nixpkgs.

(The update of nixpkgs is not related.)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,7 +33,7 @@ cache dependencies:
   stage: prepare
   extends: .common
   variables:
-    EXTRA_NIX_ARGUMENTS: --arg coqDeps true --arg ocamlDeps true --arg testDeps true --arg ecDeps true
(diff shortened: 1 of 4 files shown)


vbgl:
Yes, we need to clear the cache every time the ref to `nixpkgs` is updated.

eponier:
But also what is installed by opam is cached, isn't it? This means that if there is a new version of a package installed by opam, we need to clear the cache? Oh, maybe opam detects that it is outdated and installs the new version?

vbgl:
You never know what opam will do. It may reuse what is in the cache or pick a new set of versions.

If you want to use specific versions of your dependencies, you may want to use a package manager.

eponier:
For me the risk was that the versions of the packages installed by opam were frozen once in the cache. If that's not the case, it's good. This reduces the amount of manual clearings of the cache.

eponier (approved):
I trust @vbgl on this one.

PR #110, opened 2022-04-25
Opam file for the Jasmin compiler

vbgl:
To make it easier to just `opam install` the jasmin compiler.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -95,9 +95,8 @@ opam:
   cache:
     paths:
       - $OPAMROOT
-  before_script:
-  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'scripts/opam-setup.sh'
   script:
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'scripts/opam-setup.sh ALL'
   - >-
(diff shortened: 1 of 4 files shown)


eponier (approved):
Ok

eponier:
I've just realized, if I'm not mistaken, that we install with opam some packages that are already in the environment due to nix-shell. Isn't it a bit strange to install a dependency that is already there?

vbgl:
Can you be more specific about the dependencies that are installed twice?

eponier:
Sorry, I was not clear, and maybe I'm mistaken. All the commands are run using `nix-shell`, which provides an environment (described in default.nix) where the dependencies of Jasmin are available, e.g. coqword. So for instance when we run `opam pin --yes --no-action add coq-mathcomp-word https://github.com/jasmin-lang/coqword.git`, we have coqword already in the environment, so we ask opam to install it (not this command, but the following one) while it is already installed.

vbgl:
No. The nix shell only brings only git, pkg-config, perl, ppl, mpfr, and opam:

https://github.com/jasmin-lang/jasmin/blob/main/default.nix#L51

eponier:
Oh, I'm stupid, I saw that `coqDeps` was initialized using `!inCI`, but did not read `!` as not. This was, as expected, misunderstanding from my side.

PR #111, opened 2022-04-26
Fix install rules for macOS

eponier:
"install" on macOS does not use the same options as GNU's "install". We
choose a syntax that is valid both on macOS and GNU/Linux.


eponier:
Not sure this is the exact problem that was reported on Zulip, but this is the one I encountered on macOS 10.15. Maybe there are problems on macOS 11 and 12 since security restrictions were increased.

eponier:
Here is probably a better fix. (I've just noticed that I reverted some parts of 3c3becd1e54c4ee7f7a71b92e9c61f1c8764cc52, I'm really good at reverting @vbgl 's work it seems, so as written below, maybe we don't want to go in that direction)

What is done:
- install is replaced by install-sh everywhere
- I split the `install` and `uninstall` targets, so that the top-level Makefile just has to call the corresponding targets in the other Makefiles. This also means that each Makefile redefines variables like `PREFIX`, `INSTALL`, etc. so I'm not sure we want to go in that direction. I left the variables defined in the top-level Makefile, if we go that way we probably want them removed.
- `compiler/scripts/install-sh` is removed since there is also `scripts/install-sh`

eponier:
The simplest fix is just to use install-sh instead of install in the toplevel Makefile. I wanted also to clean up some things, but maybe the core fix was better.

eponier:
@vbgl What do you think? If the move is controversial, we can just replace install with install-sh and that's enough. But it would be good to have a release fully working on mac.

eponier:
@vbgl I'd like to have a fix for macOS before doing the release. Can you take a look, please?

vbgl (approved):
Looks good.

PR #117, opened 2022-04-29
eclib: addc & zlog2

vbgl:
Let’s keep the various branches in sync.

### DIFF
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -601,3 +601,7 @@ type leakage_t = [
 ].
 
 type leakages_t = leakage_t list.
+
+
+op zlog2 : int -> int.
+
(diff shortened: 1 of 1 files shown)


eponier:
What branch are these commits coming from?

vbgl:
They are new!

eponier:
So I fail to understand: what is kept in sync then?

vbgl:
https://github.com/jasmin-lang/jasmin/blob/constant-time-op/eclib/JModel.ec#L569-L588

eponier:
Oh ok, I wrote "commit", but I meant more generally "where does this code come from?".

eponier (approved):
Nice.

Issue #114, opened 2022-04-28
Add LGPL in the compiler release


eponier:
I think if we want to do things correctly, we have to distribute the LGPL along with the extracted files of the standard library in CIL/. And we should write that the license is not MIT but MIT+LGPL in LICENSE and opam.

What would be good is to automatize that.

PR #121, opened 2022-04-29
Better licensing for compiler-only distribution

eponier:
The problem is that currently there is no LICENSE file in the compiler tarball. And the situation of CIL is complex, since its contents is extracted from various sources.

This is a light fix, editing a bit the README and adding a few LICENSE files. Maybe we should specify for each file where it comes from, but this is tedious, so let's at least postpone that. I took the easy solution to have all this static. We could imagine adding these files while generating the tarball.

Fix https://github.com/jasmin-lang/jasmin/issues/114

### DIFF
--- a/compiler/CIL/LICENSE.coq
+++ b/compiler/CIL/LICENSE.coq
@@ -1,0 +1,502 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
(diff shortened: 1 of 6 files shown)


eponier:
@vbgl If you're okay with the principle and the phrasing, could you merge quickly please? Once it's merged, I can release 2022.04.0.

vbgl (approved):
Дякую

PR #122, opened 2022-04-29
Make the cache key depend on scripts/nixpkgs.nix

eponier:
This allows to use a different cache each time we update the nix
dependencies.


eponier:
This is failing but I don't know why.

eponier:
CI happy. @vbgl What do you think? The first commit changes the cache key automatically when we update the nix hash. The second commit is an ugly temporary hack to circumvent an opam limitation removed in the future release opam 2.2.

vbgl (approved):

PR #126, opened 2022-05-06
linear: simplify StoreLabel and make it portable across architectures

vbgl:
As of today, the semantics of the linear language is architecture-dependent. I think it is a mistake.


eponier:
This change to linear_sem was in my opinion one of the ugliest changes that we did to make Jasmin less architecture-specific, so I'm glad to see your PR. But let me first do the release.

eponier (approved):
Change very welcome.

Do you recall why it was like this before? Relying on LEA/MOV instead of directly write in the state?

vbgl:
IIRC, I found that it was elegant ;-)

eponier:
Fair enough!

PR #127, opened 2022-05-10
opam: mark easy-format-1.3.3 as conflicting

vbgl:
In CI (i.e., when dependencies are managed by nix), installation of
easy-format at version 1.3.3 fails. This OCaml library is a dependency
of `yojson` that is used in Jasmin compiler.

The cause of this failure is not really clear (might be a bug in the
opam description of the easy-format library), but this is a quick
workaround to fix the CI jobs.

### DIFF
--- a/compiler/opam
+++ b/compiler/opam
@@ -24,3 +24,6 @@ depends: [
   "ocamlbuild"
   "ocamlfind"
 ]
+conflicts: [
+  "easy-format" {= "1.3.3" }
+]
--- a/opam
+++ b/opam
@@ -31,3 +31,6 @@ depends: [
   "coq-mathcomp-algebra"
   "coq-mathcomp-word"
 ]
+conflicts: [
+  "easy-format" {= "1.3.3" }
+]


eponier:
For the record, the problem comes from an interaction between the opam sandbox and the non-standard filesystem architecture used in the CI. The opam sandbox uses a whitelist of paths. The paths in the whitelist are available (read-only or writable depending on the paths) and the rest of the filesystem is hidden from the build processes. `/home` is hard-coded in the whitelist, and the standard location of the opam root is `$HOME` which is standardly in `/home`. So, on most systems, the opam root is available in the sandbox. But not in our CI, since the opam root is `/builds/jasmin-lang/jasmin`. Which means that if the build process tries to call some opam command, it will get a failure because opam does not recognize a valid opam root. I don't know why the problem appears only with easy-format and only with version 1.3.3. Apparently, it is the only package calling opam inside its build process.

Actually, opam provides a way to add more paths available in the sandbox, using an environment variable called `OPAM_USE_PATH_RO`. But the problem is that the nix package of opam already uses this variable to make opam compatible with the filesystem architecture used by nix, overwriting any preexisting value, so we cannot use this env variable directly to add `/builds/jasmin-lang/jasmin` to the sandbox.

There are 3 ways of fixing this bug:
- fixing it for our setup only. The current PR fixes it for package easy-format which is the only problematic one for now. I have an alternative fix in PR https://github.com/jasmin-lang/jasmin/pull/122 which is more general but hacky.
- fixing it in the nix packaging of opam, so that we can use `OPAM_USE_PATH_RO`
- fixing it in opam ; actually, it is already fixed in the master branch of opam, but we have to wait for the release of 2.2 (the fix is simply that opam mounts nearly all the filesystem in the sandbox, instead of a given set of paths)

PR #129, opened 2022-05-11
Port PR #122 to release 21

vbgl:


eponier:
Note that the problem with easy-format.1.3.3 was [reported](https://github.com/ocaml-community/easy-format/issues/28) to the easy-format repository, so maybe they'll come with a quick fix and the ugly patch won't be needed anymore. Or we merge the patch, since we already have it and can remove it at anytime? What do you think?

vbgl:
Currently CI spuriously fail. I prefer to workaround issues so that CI is working.

eponier:
Ok, I'm merging then.

PR #131, opened 2022-05-12
Fix printing of while loops

vbgl:
Does this need a changelog entry?


eponier:
> Does this need a changelog entry?
Why not.

vbgl:
Done.

eponier (approved):
The first real changelog entry!

PR #132, opened 2022-05-12
ci: crytolib job really extends .common

vbgl:
This change ensures that the `before_script` variable defined in the `.common` job is not overridden

Currently, all `cryptolib` jobs (about 36 per pipeline) rebuild apron…


eponier (approved):
Indeed the `before_script` sections were not merged.

PR #133, opened 2022-05-13
Linear: simplify the types of expressions and of L-values

vbgl:
This is meant as a simplification. But maybe it is not. Any thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 
 # [unreleased]
 
+## Compiler
+
+- The linear uses simpler expressions: memory accesses cannot be nested (this
+  is now syntactically enforced).  See [PR
+  #133](https://github.com/jasmin-lang/jasmin/pull/133).
+
 ## Bug fixes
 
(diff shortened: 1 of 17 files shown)


bgregoir:
This change seems to be relatively large. Are we sure that there is a benefice ?

bgregoir (changes requested):
Improve error messages in linearisation, they should not be internal...

bgregoir (commented):
Load and Store should have type wsize & lea

bgregoir (commented):
Load and Store should have type wsize & lea
Give access to printLinear.pp_fexpr in Coq, this should allows to remove pexpr_of_fexpr

bgregoir (commented):
Load and Store should have type wsize & lea
Give access to printLinear.pp_fexpr in Coq, this should allows to remove pexpr_of_fexpr
rename fexpr -> lexpr
              lexpr -> lout | ldest 
              rexpr -> lin | lsrc

vbgl:
Thanks for the feedback. I’ve just rebased and will try to address the comments shortly.

vbgl:
This needs more thought.

PR #137, opened 2022-05-16
word: four lemmas

vbgl:


eponier (approved):
The lemmas were clearly missing.

PR #139, opened 2022-05-16
Export absolute PREFIX from toplevel Makefile

eponier:
In PR https://github.com/jasmin-lang/jasmin/pull/111, I broke the `install` rule when `PREFIX` is a relative path. My fix is simply to ask `make` to compute the absolute path first. I don't think it's a beautiful fix, if you have a better idea (including reverting the PR), please do suggestions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,11 @@
 # --------------------------------------------------------------------
 SED      ?= sed
 DISTDIR  := jasmin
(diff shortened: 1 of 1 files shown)


eponier:
(It breaks the export of the artifact of target `opam` in the CI, that's how I discovered the bug)

vbgl (approved):

PR #149, opened 2022-05-19
opam: disable sandboxing

vbgl:
This script is only meant to be run in contenerized CI jobs, thus there
is no need for an additional layer of sandboxing

### DIFF
--- a/scripts/fix_path_ro.patch
+++ b/scripts/fix_path_ro.patch
@@ -1,8 +1,0 @@
-@@ -87,6 +87,7 @@ add_dune_cache_mount() {
-   add_mount rw "$u_dune_cache" "$dune_cache"
- }
- 
-+OPAM_USER_PATH_RO="$OPAM_USER_PATH_RO:$OPAMROOT"
- # mount unusual path in ro
- if  [ -n "${OPAM_USER_PATH_RO-}" ]; then
-    add_mounts ro $(echo "${OPAM_USER_PATH_RO}" | sed 's|:| |g')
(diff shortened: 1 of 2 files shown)


eponier:
Hum, I don't know because that's clear that we don't need the security brought by sandboxing in our CI, but the good part is that we test that opam sandboxing does not get in the way of our build and install procedures. For instance, we wouldn't have discovered the problem with easy-format.1.3.3 if we had disabled sandboxing.

vbgl:
> For instance, we wouldn't have discovered the problem with easy-format.1.3.3 if we had disabled sandboxing.

For me, this is an argument of favor of the change in this PR.

Also, I have a VM in which CI jobs currently fail with:
> \# bwrap: Creating new namespace failed: Operation not permitted

eponier:
This time the problem was not our fault, but for the users it is the same, if at some point we distribute an opam package, we want to detect whether it is installable or not. And next time, the problem can come from our side.

eponier:
Do you have an idea why bwrap fails in one of the VM?

vbgl:
Probably I did not enable casual users to use bwrap.

eponier:
Okay, I found https://github.com/ocaml/opam/issues/4327 where they advise passing `--disable-sandboxing` as you did when you use a container. If we go that way, you can remove the hack about `OPAM_USER_PATH_RO`, because it was precisely due to sandboxing reasons that we have to use this hack.

eponier:
It seems you removed the `--disable-sandboxing` part.

vbgl:
Well spotted…

eponier:
Still not there

PR #151, opened 2022-05-19
Generic patches

eponier:
I extracted from the work that we did with Santiago the part that is not arm-specific. I rewrote the history a lot and try to make meaningful commits that all compile. In the end, they all compile, but not sure if they are meaningful.

To roughly sum up what has been done:
- x86_gen was made generic and added to asm_gen
- all the architecture-dependent data is packed in a record defined in arch_params
- the Coq part is now completely architecture-independent
- on the OCaml side, the syntax of programs is made parametric on the type of assembly instructions
- the file arch_full.ml adds to the arch_params defined in Coq some architecture-specific information that we need on the OCaml side but not on the Coq side
- some hardwired `U64` were replaced by `Uptr` but I'm pretty sure some of them remain


eponier:
The change is large, I don't know what would be the clever way to review that.

bgregoir:
I think the first point is to check that it does not break libjade

bgregoir (approved):
Cool CI passes

PR #155, opened 2022-05-23
CI/libjade: test extraction to EasyCrypt

vbgl:
This makes the `proof/` directory of libjade: extract jasmin programs to EasyCrypt and run EasyCrypt of the generated code.

cc @tfaoliveira

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -182,7 +182,7 @@ libjade:
   - coq
   - ocaml
   script:
-  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run './scripts/test-libjade.sh'
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run './scripts/test-libjade.sh src'
   artifacts:
     paths:
     - results/
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -170,7 +170,7 @@ check:
   - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run './compiler/jasminc.native -version'
   - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C compiler check-ci'
 
-libjade:
+libjade-compile-to-asm:
   stage: test
   variables:
     EXTRA_NIX_ARGUMENTS: --arg testDeps true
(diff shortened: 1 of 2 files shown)


vbgl:
>     * Does variable `$(NIX_BUILD_CORES)` indicates the number of available cores?

Yes.

> If so, do you think it is reasonable to run `make -j $(NIX_BUILD_CORES)` to compile the source code and perform the extraction as well?

I don’t know how this makefile can cope with parallelism.

>     * I think @strub and @fdupress want to be consumers of [ecfiles.tar.gz](https://github.com/formosa-crypto/libjade/blob/90cb7cb5554682c6a76395ea4034b7f56ba39c00/proof/Makefile#L82). If they confirm this interest, would it be possible to make it available in the form of an artifact?

Sure. It would be nice if the Makefile from `libjade` would put all artifacts in a directory so that we can directly point the CI to this directory.

>     * Given the last line of `test-libjade.sh`, `make $MAKELINE err`, do you think it is advantageous to separate `$(MAKE) err` from [proof](https://github.com/formosa-crypto/libjade/blob/90cb7cb5554682c6a76395ea4034b7f56ba39c00/proof/Makefile#L38) and [src](https://github.com/formosa-crypto/libjade/blob/90cb7cb5554682c6a76395ea4034b7f56ba39c00/src/Makefile#L34) Makefiles from the 'default' rules?

Of course. It would be simpler to just call `make` in the CI process.

tfaoliveira:
> 
> I don’t know how this makefile can cope with parallelism.
> 

Locally, I use `make -j4` for extraction and compilation without any issue. I guess it would be nice to check if it works on other environments as well before merging. Maybe this can be done in a separate pull to understand if it causes any issues. 

 
> Sure. It would be nice if the Makefile from `libjade` would put all artifacts in a directory so that we can directly point the CI to this directory.

Great, in this case, we can sync with them about this. Putting all artifacts in a directory should be easy. This can also be done separately.

> 
> > Given the last line of `test-libjade.sh`, `make $MAKELINE err`, do you think it is advantageous to separate `$(MAKE) err` from [proof](https://github.com/formosa-crypto/libjade/blob/90cb7cb5554682c6a76395ea4034b7f56ba39c00/proof/Makefile#L38) and [src](https://github.com/formosa-crypto/libjade/blob/90cb7cb5554682c6a76395ea4034b7f56ba39c00/src/Makefile#L34) Makefiles from the 'default' rules?
> 
> Of course, it would be simpler to just call `make` in the CI process.

I'm not sure if I understand. Currently, `make err` is being run in those Makefiles in the context of the 'default' rule. Did you face any issues in recovering the return code?

vbgl:
> I'm not sure if I understand. Currently, `make err` is being run in those Makefiles in the context of the 'default' rule. Did you face any issues in recovering the return code?

I was not clear. Sorry. Things might be simpler if either:

  - `make` never failed and we can call `make err` at the end to detect success or failure;
  - or `make` did all in a single call (extract, check, build artifacts, return error status).

tfaoliveira:
Great, once we figure out the requirements from the EasyCrypt team, we can do these adjustments (I would aim for the second solution, make doing everything, to simplify the process).

tfaoliveira (approved):

PR #156, opened 2022-05-24
lea_proof: cleaning

vbgl:


eponier (approved):
Removing section hypotheses is always good.

PR #157, opened 2022-05-25
clean spurious Require

vbgl:
Running `make CIL` no longer compiles any `_proof` file!


eponier:
Oups, I was too fast in splitting code and proof. Thanks, Vincent!

PR #158, opened 2022-05-25
allows to move mmx reg ptr to reg ptr

bgregoir:
this allows to write 
#mmx reg ptr u64[8] e;
reg ptr u64[8] x;
x = e;
e = x;

It also allows to write
#mmx reg u64 e;
reg u64 x;
x = e;
e = x;

without using x = MOVX(e)
Examples are provided in tests/success/movx.jazz


eponier:
I didn't take a close look at the code, but do is_regx have to be both in lower_params and compiler_params?

vbgl (approved):

PR #160, opened 2022-05-31
compiler/default.nix: add missing dependency on camlidl

vbgl:
When using a recent `nixpkgs` (i.e., strictly posterior to 21.11), building from sources the released version of Jasmin fails with:

> \+ ocamlfind ocamlopt -rectypes -linkpkg -g -package 'batteries, menhirLib, zarith, apron.octMPQ, apron.ppl, apron.boxMPQ, yojson' -I entry jasmin.cmx entry/jasminc.cmx -o entry/jasminc.native
> …/ld: cannot find -lcamlidl: No such file or directory
> collect2: error: ld returned 1 exit status

This PR fixes this issue (well, not for released versions…).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -155,6 +155,19 @@ tarball:
     paths:
     - compiler/$TARBALL.tgz
 
+build-from-tarball:
+  stage: test
+  variables:
+    TARBALL: jasmin-compiler-$CI_COMMIT_SHORT_SHA
+  needs:
+  - tarball
+  dependencies:
+  - tarball
+  script:
+  - tar xvf compiler/$TARBALL.tgz
+  - nix build -o out -f $TARBALL
+  - ./out/bin/jasminc -version
+
 check:
   stage: test
   variables:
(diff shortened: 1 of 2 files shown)


eponier:
What is it used for and why did we not see that before?

vbgl:
Are you sure there is no way to use `apron` without `camlidl`? If this is the case, then there is a bug in `nixpkgs`.

eponier:
I have the impression that it is at least not possible to have the ocaml bindings of apron without camlidl.

eponier:
But when I look at the [nix file](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/ocaml-modules/apron/default.nix#L39) describing apron in nixpkgs, camlidl is listed as a dependency, so really I don't understand what the problem is.

vbgl:
Dependencies are only transitive in Hell.

eponier:
? Anyway, for Jasmin, is camlidl needed for something else than as a dependency of apron?

vbgl:
I don’t understand the question. `camlidl` is needed to link the final `jasminc` binary.

bgregoir:
What is the status of this pull request ? It seems nothing change for a while.

eponier:
I've still not understood everything, but I'm convinced enough and I can reproduce the problem, so I'm merging.

eponier (approved):

PR #161, opened 2022-06-03
Example standalone Jasmin program

vbgl:
I’ve just been told that it is not always necessary to write wrappers in a foreign language to call a function written in Jasmin. Here is an illustration.


bgregoir (approved):
This is cool, maybe we should add this example in the wiki.

PR #162, opened 2022-06-07
Removed duplicate definition of subtype

dwRchyngqxs:
Definition of `subtype` and related lemmas in `proofs/lang/values.v` was ducplicated from `proofs/lang/type.v`.


bgregoir (approved):
nice catch!

PR #163, opened 2022-06-09
linear and one_varmap do not depend on arch anymore

bgregoir:
This is the first step, next step is to add calling convention in Coq. So we can wait for the merge.

PR #167, opened 2022-06-10
asm-gen: accept any non-bool register as callee-save

vbgl:


bgregoir (commented):
I did almost the same in the pull request #163...

bgregoir:
(only the first commit)

eponier:
Should we close this in favor of #163 then?

PR #169, opened 2022-06-14
More aggressive t_xrbindP + preserves names

dwRchyngqxs:
Changes to t_xrbindP:
- removes units (intended for assert)
- preserves names


dwRchyngqxs:
The one in `compiler/compiler_util.v` unfolds functions introduced in that file. I don't know if/how they can be factored.

proofs/arch/asm_gen_proof.v
```diff
@@ -1139,7 +1132,7 @@ Lemma assemble_i_is_label (li : linstr) (ai : asm_i) lbl :
   -> linear.is_label lbl li = arch_sem.is_label lbl ai.
 Proof.
   by (rewrite /assemble_i /linear.is_label ; case li =>  ii []; t_xrbindP)
-    => /= [????? <- | <- | ? <- | ? <- | _ _ _ ? _ <- | _ ?? _ <- | ???? <-].
+    => /= [ > _ <- | <- | ? <- | ? <- | ? _ ? _ <- | > _ <- | > _ <-].
```
eponier, line 1135: @vbgl do you see any problem using this new syntax `>` (available from Coq 8.10)

proofs/compiler/merge_varmaps_proof.v
```diff
@@ -651,10 +651,10 @@ Section LEMMA.
   Proof.
     move => s1 m2 s2 jj xs fn args vargs vs ok_vargs sexec ih ok_s2 sz ii I O t1.
     rewrite /check_instr_r /=; case heq : get_fundef => [ fd | //].
-    t_xrbindP => ? hces _ hal _ hra _ hargs _ hres hxs _ pre sim.
+    t_xrbindP => hces hal hra hargs hres hxs _ pre sim.
     have [ok_wrf] := checkP ok_p heq.
     rewrite /check_fd; t_xrbindP => D check_body _ hdisjoint _ checked_params.
-    move=> _ RSP_not_result _ preserved_magic ? hsaved checked_ra.
+    move=> hsaved checked_ra.
```
eponier, line 657: I think the names have not been updated accurately.

eponier (dismissed):
Modulo the few remarks, this is ok to me. The duplication of `t_xrbindP` was already there anyway.

proofs/arch/asm_gen_proof.v
vbgl: Can you please tell me where this feature is documented? So that I can learn something…

vbgl: Thanks!

The current Coq development is not compatible with versions earlier than 8.13 and 8.13 is broken. So there is no problem in using features that are not available before 8.14.

proofs/compiler/merge_varmaps_proof.v
dwRchyngqxs: The names aren't used, I replaced them with ?.

eponier:
This should be good now.

vbgl (approved):

PR #173, opened 2022-06-16
Reset the default `call_conv` in `main_compiler`

vbgl:
This makes the `Glob_options` module independent of `Unix`


bgregoir (approved):
nice

PR #174, opened 2022-06-17
Small improvements to the CT checker

vbgl:
The main change is that the checker no longer calls `printf` during the type-cheking.

The last commit can be dropped: when generating fresh names for “poly” security levels of arguments, it tries to use the name of the argument instead of `l0`, `l1`, etc. I find it more readable.

### DIFF
--- a/compiler/src/ct_checker_forward.ml
+++ b/compiler/src/ct_checker_forward.ml
@@ -411,7 +411,8 @@ let ty_lvals1 env xs lvl =
 (* -----------------------------------------------------------*)
 
 let get_annot ensure_annot f =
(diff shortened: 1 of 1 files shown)


bgregoir (commented):
I have the impression that the new code allows to have different vl with the same name.
(see commit: fresh level names are more readable)

vbgl:
I think the recursive call below prevent duplicate names:
https://github.com/jasmin-lang/jasmin/pull/174/commits/a3aa2dad585e7e012c4f2cafcb269ebce47202d3#diff-3b6cec615a99ad926af39d7e4e398d7ff96685051cb950f48e37e09c4a76752bR449

bgregoir:
Ok with this

PR #178, opened 2022-06-22
checking of constant/writable arguments

bgregoir:


compiler/src/pretyping.ml
```diff
@@ -1217,7 +1217,7 @@ let tt_lvalue pd (env : 'asm Env.env) { L.pl_desc = pl; L.pl_loc = loc; } =
 
   | S.PLVar x ->
     let x = tt_var `NoParam env x in
-    reject_constant_pointers loc x ;
+(*    reject_constant_pointers loc x ; *)
```
vbgl, line 1220: Maybe this line should be removed completely.

eponier:
This example is artificial, but do we want to accept such things?

```
fn f (reg ptr u64[1] r) -> reg u64 {
  reg u64 res;
  stack u64[1] s;

  r = s;
  r[0] = 1;
  res = r[0];

  return res;
}

  reg u64 res;
  stack u64[1] s;

  s[0] = 1;

  res = f(s);

  return res;
}
```

vbgl (changes requested):
This is not ready. The following program is now rejected:

~~~
fn zerofill(reg mut ptr u64[1] x) -> reg ptr u64[1] {
  x[0] = 0;
  return x;
}

export
fn main() -> reg u64 {
  stack u64[1] s;
  s = zerofill(s);
  reg u64 r;
  r = s[0];
  return r;
}
~~~

eponier:
With @bgregoir, we are not sure what to do exactly, so let's postpone this a bit.

vbgl:
IMHO, it is OK to reject the artificial example above.

bgregoir:
I did not see your last commit and I did the same patch. 
Anyway I think we try to ensure something that is verified later by the compiler without having all the information.
We can do a variation of zerofill example:

inline fn zerofill(stack u64[1] x) -> stack u64[1] {
  x[0] = 0;
  return x;
}

export
fn main() -> reg u64 {
  stack u64[1] s;
  reg const ptr u64[1] t;
  t = s;
  s = zerofill(t);
  reg u64 r;
  r = s[0];
  return r;
}

compiler/src/pretyping.ml
```diff
@@ -1595,18 +1594,39 @@ let check_lval_pointer loc x =
   | _ -> rs_tyerror ~loc (NotAPointer x)
 
 let mk_call loc is_inline lvs f es =
-  begin match f.P.f_cc with 
-  | P.Internal | P.Export -> ()
-  | P.Subroutine _ ->
+  let open P in
+  begin match f.f_cc with
+  | Internal -> ()
+  | Export -> if is_inline = E.InlineFun then rs_tyerror ~loc (string_error "call to export function needs to be inlined")
```
vbgl, line 1600: This test is wrong…

vbgl: Fixed!

vbgl:
Unless someone wants to implement a more precise check, I recommend to merge this PR as is.

bgregoir:
Ok with this.

bgregoir:
Apparently I can't merge it myself.

eponier:
You have to accept it first (i.e. add a positive review).

eponier (approved):
@bgregoir accepts his own PR

PR #179, opened 2022-06-24
default flag variable are automatically declared

bgregoir:
declare default flag variables allow to replace 
?{cf,zf,sf,of, "==" = b0, "<s" = c, "<u" = b1} = #CMP_64(x,y);
by 
?{ "==" = b0, "<s" = c, "<u" = b1} = #CMP_64(x,y);

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -1436,6 +1436,12 @@ let pexpr_of_plvalue exn l =
   | S.PLMem(ty,x,e) -> L.mk_loc (L.loc l) (S.PEFetch(ty,x,e))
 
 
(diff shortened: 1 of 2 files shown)


vbgl:
This seems indeed more convenient. However, there is a new kind of cryptic messages. The following program is rejected by the type-checker:

> typing error: the expression has type bool instead of u64

~~~
 fn test(reg u64 x) -> reg u64 {
  reg u64 _zf_;
  inline bool c;
  _zf_ = x;
  ?{ "==" = c } = #CMP(x, x);
  return _zf_;
}
~~~

eponier:
Is this PR abandoned?

PR #189, opened 2022-06-28
Followup to #188

eponier:
We use `of_var_e_bool` everywhere.

I think we could actually get rid of `of_var_e`, since there is only one usecase any more, and only with registers.


bgregoir (approved):
Yes, previous patch was stupid.

vbgl (changes requested):
Proofs are broken, right?

eponier:
Yes, sorry for that.

PR #204, opened 2022-07-11
Two trivial fixes

vbgl:
Currently, trying to `Require Import Uint63.` fails with:

> Error: Notation "_ >> _" is already defined at level 25 with arguments constr at level 25, constr at next level
while it is now required to be at level 30 with arguments constr at next level, constr at next level.

The first patch fixes this issue.

The second patch removes the single (real) use of `stk_fun_extra` as an `eqType`.

### DIFF
--- a/proofs/compiler/merge_varmaps_proof.v
+++ b/proofs/compiler/merge_varmaps_proof.v
@@ -1,6 +1,7 @@
 (*
 *)
 Require Import sem_one_varmap sem_one_varmap_facts merge_varmaps psem_facts.
+Require Import seq_extra.
 Import Utf8.
 Import all_ssreflect all_algebra.
 Import ssrZ.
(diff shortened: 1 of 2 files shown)


eponier:
What's the goal of the second patch? Does it prepare something else?

vbgl:
eqTypes are annoying: we have to defined and maintain their definitions (proofs included), it’s hard to make them abstract (they interact badly with extractions), etc.

I’ve two experiments in preparation that would benefit from the second patch:

  - use more efficient data-structures for `var_info` and `instr_info` (say a plain OCaml `int`);
  - add a generic (unspecified) annotation to loads and stores (it is super annoying to have to say that this generic `Type` is in fact an `eqType`).

eponier:
These experiments look interesting, I'm looking forward to the results!

eponier (approved):

PR #206, opened 2022-07-12
Pretend pexpr is not an eqType

vbgl:


bgregoir (commented):
I will be happy to accept this. My only question is should we wait until we are able to remove the eqType on expr ?

bgregoir (approved):

PR #211, opened 2022-07-19
Syscall_t: make independent of extracted code from Coq

vbgl:
Currently, the `Syscall_t` module depends on the `BinNums` module, that is extracted from Coq code.

But the extracted code, residing in the `CIL/` directory, depends on the `Syscall_t` module.

This PR breaks this circular dependency.

This hack should disappear as soon as there are more than one syscall.

### DIFF
--- a/compiler/src/alias.ml
+++ b/compiler/src/alias.ml
@@ -170,7 +170,7 @@ let assign_arr params a x e =
   | None, _ | _, None -> a
   | Some d, Some s -> merge_slices params a d s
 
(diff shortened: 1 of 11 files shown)


vbgl:
The problem is that you need to build everything at once. I wished I could compile only parts of Jasmin…

eponier:
And if we add another syscall, will that make this hack useless or not possible anymore?

vbgl:
Useless because the module `Syscall_t` will be removed altogether (see comment in `lang/extraction.v`).

PR #223, opened 2022-08-17
Adapt w.r.t. jasmin-lang/coqword#12.

ppedrot:
Hopefully this should fix the Coq CI.


vbgl:
Is there any way to write code that would be compatible with both versions of coqword?

ppedrot:
Not that I know of.

strub:
We could revert in coqword and de ide to change the namespace only for the next release.

vbgl (approved):
I guess that a new release of `coqword` would be appropriate indeed.

ppedrot:
Whatever you do, please do it swiftly so that we can get a green CI in Coq asap.

ppedrot:
Thanks!

PR #229, opened 2022-09-01
Fix 202

vbgl:


bgregoir (commented):
The case for Copn does not need to be changed to ?

vbgl:
I think the case of `Copn` is right: auxiliary variables are used only for the outputs, which is consistent with what `init_aux_i` does.

bgregoir (approved):

PR #243, opened 2022-09-12
Check after unrolling that for-loops and other “inline” instructions are gone

vbgl:
Side-effect of this PR: improve the printer. That was rather useful for debugging.


eponier (approved):
Seems good to me, thanks!

PR #244, opened 2022-09-15
Add `else if` construction

dsprenkels:
This patch add syntactic sugar for `else if` constructions. E.g.:

```jazz
if x == 0 {
  // (...)
} else if x == 1 {
  // (...)
} else {
  // (...)
}
```

is parsed as if it were: 

```jazz
if x == 0 {
  // (...)
} else {
  if x == 1 {
    // (...)
  } else {
    // (...)
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,19 @@
 - new option
 	"-call-conv {windows|linux}": select calling convention (default depend on host architecture)
 
+- Added syntactic sugar for `else if` blocks
+  ([PR #224](https://github.com/jasmin-lang/jasmin/pull/244)).
+  I.e., you can now use constructions like:
+  ```
+  if x == 0 {
+    // (...)
+  } else if x == 1 {
+    // (...)
+  } else {
+    // (...)
+  }
+  ```
+
 ## Improvements
 
 - Intrinsics present at source-level can no longer be removed by dead-code elimination
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@
 	"-call-conv {windows|linux}": select calling convention (default depend on host architecture)
 
 - Added syntactic sugar for `else if` blocks
-  ([PR #224](https://github.com/jasmin-lang/jasmin/pull/244)).
+  ([PR #244](https://github.com/jasmin-lang/jasmin/pull/244)).
   I.e., you can now use constructions like:
   ```
   if x == 0 {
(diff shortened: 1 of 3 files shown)


bgregoir (commented):
I am not sure this is the good patch.
I think the right way to do it is to allow single instruction without { } in the else 
branch (maybe also in the then branch, but I think this will create conflit (for the then)) 
Maybe I am wrong, so I will be happy to change my mind.

dsprenkels:
@bgregoir Do you mean that we should allow constructions like this?

```
i = 0;
if x == 0
  i = 10;
// etc.
```

I think, especially because there is no automated formatter yet for jasmin, that we should *not* do this. I think we want to prevent bugs to happen because of bad indentations. Using braces is _always_ clear, and I don't think we should mind the extra verbosity.

Also, this patch only restricts the allowed instruction to be another `if-else` block. We can always loosen this restriction later on.

bgregoir:
What do you means by bad indentation? 
Something like 
 if t 
    x = e;
y = z;
while what you want to write is 
 if t {
    x = e;
    y = z;
}

I ask the question, because the parser is independent from the indentation.

vbgl:
There is a need for delimiters: between the guard and the then branch, to correctly link else branches to the corresponding if, etc. I think the current syntax is good.

Do you have existing code to show in which this patch would allow to enhance the readability?

dsprenkels:
@bgregoir Yeah I think that it would be ok to always keep the explicit braces, s.t. it is always (to the reader) unambiguous what is included into the block. The programmer may have used poor indentation and it may look like `y = z;` is included in the `if`-block, while in reality it is not. If we always require braces to be present, then this will not be an issue.

@vbgl E.g., https://github.com/formosa-crypto/libjade/blob/bad7ac3373b635008c7302d6c0f488bccf0b0155/src/crypto_sign/dilithium/dilithium3/amd64/ref/rounding.jinc#L99-L108

CHANGELOG.md
```diff
@@ -56,6 +56,19 @@
 - new option
 	"-call-conv {windows|linux}": select calling convention (default depend on host architecture)
 
+- Added syntactic sugar for `else if` blocks
+  ([PR #224](https://github.com/jasmin-lang/jasmin/pull/244)).
```
vbgl, line 60:
```suggestion
  ([PR #244](https://github.com/jasmin-lang/jasmin/pull/244)).
```

vbgl (approved):
I’m not convinced by the example but I’m not against this change.

A possible improvement would be to tweak the pretty-printers to take advantage of this new feature.

vbgl:
Thanks for your contribution.

PR #245, opened 2022-09-16
opam: relax constraints on Coq & mathcomp versions

vbgl:


eponier (approved):
I was about to submit a similar patch. I wait for the CI but in the principle this is ok.

PR #247, opened 2022-09-20
OCaml Evaluator: accept arguments & instr-info

vbgl:
This slightly improves error messages

### DIFF
--- a/compiler/src/evaluator.ml
+++ b/compiler/src/evaluator.ml
@@ -140,20 +140,16 @@ let small_step1 spp s =
 let rec small_step spp s =
   small_step spp (small_step1 spp s)
 
-let init_state scs0 p fn m =
-  let f = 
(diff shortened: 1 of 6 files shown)


bgregoir (commented):
Maybe I misunderstood the goal of this pull request.
My understanding was that you want to accept arguments to functions when you start evaluation. But the parsing does not allow this.

vbgl:
That’s right, the `jasminc` frontend only calls the evaluator on functions without arguments; but I suspect that some users may want to evaluate functions with arguments (and Jasmin users can write their own front-end, can’t they?).

This just removes a silly assertion. Concrete syntax — if needed — can wait another PR.

bgregoir (approved):
Ok, I will merge the PR. I think we should add something in the TODO to extend the syntax.

PR #248, opened 2022-09-22
Cleaning in the `examples/` directory

vbgl:


bgregoir (commented):
Why did you change + by LEA in some test ?
Else I think it is ok.

vbgl:
In siphash: I’ve removed the `-lea` argument from the Makefile and added a `#LEA` instruction where necessary (I find it weird to systematically use LEA instead of ADD).

In test_var_alloc2, it’s to make it compile (currently, all is dead code so the test case does not make much sense).

bgregoir (approved):

PR #255, opened 2022-09-28
Fix opam

eponier:
Several improvements to opam files

### DIFF
--- a/compiler/opam
+++ b/compiler/opam
@@ -1,5 +1,6 @@
 opam-version: "2.0"
 name: "jasmin"
+version: "dev"
 maintainer: "Jean-Christophe Léchenet <jean-christophe.lechenet@inria.fr>"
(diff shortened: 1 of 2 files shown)


compiler/opam
```diff
@@ -1,5 +1,6 @@
 opam-version: "2.0"
 name: "jasmin"
+version: "dev"
```
vbgl, line 3: Can you please elaborate on why this change is important?

Should a similar patch be applied to the `opam` file for `mathcomp-word`?

eponier: Actually, it's more that I wanted to apply this patch to `mathcomp-word` and decided to apply it here too. I think it's better practice in general that when you pin the git repo, you get "dev" as version name.

vbgl (approved):

PR #261, opened 2022-10-04
Fix for Coq 8.16

vbgl:
If anyone has an explanation or a better fix, please share it!


eponier:
What? The instance doesn't have a name?

vbgl:
I can call it `toto` but Benjamin will prefer it to be called `foo`. So in the end I’ll pick your choice.

bgregoir (approved):

PR #264, opened 2022-10-05
Increase the lower bound on OCaml version

eponier:
Function `String.for_all` is used in `latex_printer.ml`, and it was introduced only in OCaml 4.13.

### DIFF
--- a/compiler/opam
+++ b/compiler/opam
@@ -16,7 +16,7 @@ install: [
   cp "_build/entry/jasminc.native" "%{prefix}%/bin/jasminc"
 ]
 depends: [
-  "ocaml" { >= "4.08" & build }
+  "ocaml" { >= "4.13" & build }
   "batteries" {>= "3.2"}
   "menhir" {>= "20160825" & build }
   "menhirLib"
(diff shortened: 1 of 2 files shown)


vbgl (approved):
Let’s fix the wiki too.

I guess we need to port that to the release-2022.09 branch.

vbgl (changes requested):
Please change `scripts/opam-setup.sh` to use the correct OCaml version for CI.

vbgl:
`String.for_all` comes from batteries.

vbgl:
Released version 2022.09.0 works fine with OCaml 4.08.1.

Current main branch uses `List.concat_map` (in the ARM asm printer), and therefore requires 4.10.

eponier:
Ok, I had a bit of trouble to understand everything. batteries comes with `String.for_all` from version 3.4.0, and OCaml has it too from version `4.13`. The easy fix is to raise the lower bound on batteries.

eponier:
cf. https://github.com/jasmin-lang/jasmin/pull/265

PR #265, opened 2022-10-05
Raise versions of OCaml and batteries

eponier:
This should be the right fix.


vbgl (approved):
Looks good. Would it make sense to test these lower bounds in CI?

eponier:
Not sure if it's worth the trouble. opam's CI is really good at spotting this kind of bug, so we test the bounds at least at each release.

PR #269, opened 2022-10-13
Utils: remove dead code

vbgl:


vbgl:
I think that you’re right that some code in that module has become obsolete due changes to the standard / batteries libraries.

However, that module is far from empty (even after this PR):

  1. there are a few types and utility functions defined there, used in the compiler, and not part of any “standard” library that I know (please prove me wrong);
  2. what’s at the bottom (starting at https://github.com/jasmin-lang/jasmin/blob/240aa708270979e96cbcb7ee255d3b28a4c308de/compiler/src/utils.mli#L147) is very specific to the Jasmin compiler and cannot be remove (though it may be moved to a better place).

PR #270, opened 2022-10-13
x86: remove support for explicit if-then-else in flag combinations

vbgl:
A comment says this should be removed…

Does this need an entry in the CHANGELOG?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 - Support ARMv7 (Cortex-M4) as target architecture
   ([PR #242](https://github.com/jasmin-lang/jasmin/pull/242)).
 
+## Other changes
+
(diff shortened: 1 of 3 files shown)


eponier:
Does it change anything for users?

cc @bgregoir who wrote that comment (commit 33d6b62dc11a5c37ed8feb92cfd87a28036e3438)

vbgl:
This can break legacy code (if the programmer explicitly wrote something like: `x = y if (sf ? !of : of)`, which can easily be written as `x = y if sf != of`, up to typo).

My belief is that no such legacy code exist.

eponier:
Ok, but it is preferable to have something written down in the changelog then, I think. The explanation can be short.

vbgl:
I’ve just added a line into the changelog.

eponier (approved):
Nice changelog message!

PR #271, opened 2022-10-19
Split allocation into code + proof

eponier:
I removed the functors and used typeclasses instead. I mostly copy-pasted the code, but without understanding everything, maybe there were opportunities to simplify the code that I didn't see.


bgregoir (approved):
Cool, we where waiting for this from a long time.

PR #272, opened 2022-10-24
Remove field id_wsize from instruction descriptors

vbgl:
This field is generally meaningless. It is heuristically used in lowering on x86 to check whether an immediate value fits within the instruction or should go through an intermediate register.

I believe that the current design is bad and that the cleaning done in this commit is a necessary step towards a better design (that is yet to be discovered).


eponier (approved):
No strong opinion, if you think that's the way to go, I'm fine with it.

vbgl:
Yes, I’m convinced that this field does not make any sense.

PR #275, opened 2022-10-26
Cut dependencies

eponier:
The goal was to reduce the dependencies of compiler.v and extraction.v as much as possible. The commits make more or less sense when taken in isolation, but I did that in a messy way, so the intermediate commits probably do not compile. It is probably hard to review for this reason, sorry for that.


eponier:
Commit https://github.com/jasmin-lang/jasmin/pull/275/commits/7caaa97e66460696523bafeee0edf89c3382ad86 removes sem from the dependencies of compiler, but introduce a bit of redundancy (we more or less clone `app_sopn`).
Commit https://github.com/jasmin-lang/jasmin/pull/275/commits/38efe17bf737ebd237c79e2960a7b2298031dd8a tries to isolate SemPexprParams. Thanks to a "Hint Cut" black magic I discovered recently, we don't really need `spp_of_asm_e` anymore. But it is useful to have a nice name on the OCaml side, so I reintroduced the function just to be able to call it from main_compiler.ml.

vbgl:
Looks good to me. But CI is not happy.

Do you think we should aim to remove the duplication of `app_sopn`?

vbgl (approved):
I’m going to squash all these changes into a single patch, unless somebody quickly objects.

eponier:
No, please do, the commits are not self-contained anyway. But let me have a look at the `app_sopn` stuff.

eponier:
I did a basic refactoring. Should be ready.

vbgl (approved):

PR #281, opened 2022-11-07
ssa: use `filter_map` rather than `flatten ∘ map`

vbgl:
This might improve compilation time of laaarge programs


eponier (approved):
I don't know the impact on compilation time, but this is prettier anyway.

PR #283, opened 2022-11-08
Make the `cast_w` constructor actually smart

vbgl:


eponier:
Tiago pushed some fix to libjade, I restarted the CI manually.

eponier:
What is the motivation of this change?

Note to self: maybe this allows to remove cast_word from stack_alloc.v

vbgl:
About the motivation: it seems that doing this “push cast” normalization during asm-gen (when `mk_lea` is used) is too late. The compiler can produce expressions that are already in the correct shape for asm-gen to work smoothly.

An other option would be to have a `LEA of lea` leaf in the language of expressions…

bgregoir (approved):

PR #284, opened 2022-11-09
Refactor printers

vbgl:
This removes the dependency between Printer and PrintLinear modules.

### DIFF
--- a/compiler/src/printCommon.ml
+++ b/compiler/src/printCommon.ml
@@ -1,0 +1,143 @@
+open Format
+open Utils
+open Prog
+open Wsize
+module E = Expr
+
+(* -------------------------------------------------------------------- *)
+
+let pp_string0 fmt str = fprintf fmt "%a" (pp_list "" pp_print_char) str
+
+(* -------------------------------------------------------------------- *)
+
+let pp_wsize fmt sz = fprintf fmt "%a" pp_string0 (string_of_wsize sz)
+
+(* -------------------------------------------------------------------- *)
+
+let string_of_signess s = if s = Unsigned then "u" else "s"
+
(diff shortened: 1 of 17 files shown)


compiler/src/printLinear.ml
```diff
@@ -11,9 +12,6 @@ module F = Format
 module Pr = Printer
```
eponier, line 12: This should go away, I guess?

compiler/src/printCommon.ml
```diff
@@ -0,0 +1,92 @@
...
+  | Ovlsl (ve,ws) -> asprintf "<<%s" (string_of_velem Signed   ws ve)
+
+(* -------------------------------------------------------------------- *)
+
+let pp_var tbl fmt x =
+  let y = Conv.var_of_cvar tbl x in
+  fprintf fmt "%s" y.v_name
+
+let pp_var_i tbl fmt x =
+  pp_var tbl fmt x.E.v_var
```
eponier, line 92: This is used only in printLinear.ml, is it worth to put it here?

eponier (commented):
2 comments

compiler/src/printLinear.ml
vbgl: Done. Thanks.

compiler/src/printCommon.ml
vbgl: This is in anticipation of a future change. Should I postpone this?

eponier (approved):

compiler/src/printCommon.ml
eponier: No, that's ok.

PR #285, opened 2022-11-09
Add VPAVG x86 intrinsic

vbgl:


bgregoir (approved):
The +1 is surprising but corresponds to the x86 semantic :-).
Should we add something in the change log ?

vbgl:
I’ve just added a changelog entry.

vbgl:
Regarding the +1, it’s a usual trick to get rounding to the nearest value.

PR #288, opened 2022-11-14
arm: immediate arguments are non-negative

vbgl:
This fixes some examples using `movt`.

This also corrects a terminology confusion between alloc and free…

### DIFF
--- a/compiler/src/conv.ml
+++ b/compiler/src/conv.ml
@@ -26,14 +26,8 @@ let z_of_int32 z  = z_of_cz (Word0.wsigned W.U32 z)
 let z_of_int16 z  = z_of_cz (Word0.wsigned W.U16 z)
 let z_of_int8 z  = z_of_cz (Word0.wsigned W.U8 z)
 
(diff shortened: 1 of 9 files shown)


sarranz (approved):
Maybe we could add a comment like this at the beginning
```
Immediate values are always nonnegative integers.
```
and then modify the first comment
```
Offset addressing:
     - A base register and an immediate offset (displacement):
       [<reg>, #+/-<imm>] (where [+] can be omitted).
     - A base register and a register offset: [<reg>, <reg>].
     - A base register and a scaled register offset: [<reg>, <reg>, LSL #<imm>].
```
so that we have exactly the same as the manual and it's explicit why displacement keeps the `Conv.z_of_word`.

vbgl:
Thanks Santiago. Can you please commit your suggestion to this branch?

sarranz:
> Thanks Santiago. Can you please commit your suggestion to this branch?

Done!

vbgl:
Thanks.

PR #290, opened 2022-11-15
Add ROL and ROR operators

sarranz:
Extend lexing and parsing with `<<r` and `>>r`.
Extend latex and EasyCrypt printing.
Define operators in `eclib/JWord.ec`.
Implement lowering for x86 and ARM-M4.


vbgl:
This definitely needs a changelog entry!

sarranz:
> This definitely needs a changelog entry!

That's right, I forgot.

vbgl (approved):
There is still time to fight about concrete syntax before the next release :-)

PR #295, opened 2022-11-16
Fix ARM-M4 control flow

### LINKED ISSUE #293: Fix ARM-M4 control flow
1. The `LR` register can, technically, be used as a general purpose one, but it also stores the return address after a function call, so the compiler should not use it.
2. System calls should use the `BL` instruction.
3. Indirect jumps should use the `BX` instruction.

sarranz:
Closes #293.

### DIFF
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -168,29 +168,28 @@ let pp_instr tbl fn (_ : Format.formatter) i =
       [ LInstr ("b", [ pp_remote_label tbl lbl ]) ]
 
   | JMPI arg ->
-      (* TODO_ARM: Review. *)
-      let lbl =
+      let r =
         match arg with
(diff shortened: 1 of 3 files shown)


compiler/src/pp_arm_m4.ml
```diff
@@ -168,29 +168,28 @@ let pp_instr tbl fn (_ : Format.formatter) i =
...
   | JAL _ ->
-      failwith "TODO_ARM: pp_instr jal"
+      failwith "Internal error: JAL"
 
   | CALL lbl ->
       [ LInstr ("bl", [ pp_remote_label tbl lbl ]) ]
 
   | POPPC ->
-      [ LInstr ("b", [ pp_register LR ]) ]
+      [ LInstr ("bx", [ "lr" ]) ]
```
vbgl, line 189: This looks wrong. IMHO, should be “pop pc”.

PR #302, opened 2022-11-30
Distinguish internal from external labels

vbgl:
External labels correspond to return addresses, or more generally to targets of computed jumps.

External labels may be encoded as a pointer and then stored in registers or in memory.

As the (unspecified) encoding of labels only depends of the set of external labels, adding fresh internal labels is guaranteed to preserve the encoding.


eponier (approved):
This allows to inject code after linear without problems with labels.

eponier:
Waiting for CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/749464240.

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/750652501

PR #304, opened 2022-12-01
Fix version name in changelog

eponier:
I bugged when looking at the changelog: where is the "2022.04.0" version? It turns out the old name "22.0" was still in use in the changelog.


vbgl (approved):

eponier:
I guess we should backport this one?

vbgl:
Right.

eponier:
Done.

PR #305, opened 2022-12-05
Remove some dead code

vbgl:


eponier (approved):
`disjoint_uincl_on` looks like it could be useful, but if it turns out it is not used...

PR #307, opened 2022-12-06
Make estate not depend on SemPexprParams

eponier:
The dependency was artificial. This requires to add some `(spp:=mk_spp)`, but it seems we cannot avoid those.


eponier:
cc @bgregoir

PR #316, opened 2022-12-16
AES-NI impl in Coq

bshvass:
Making a PR cf this [zulip thread](https://formosa-crypto.zulipchat.com/#narrow/stream/308530-Jasmin.3Adevelopment/topic/AES.20operations.20in.20.60x86_instr_decl.60).

The PR contains implementations of all the AES instructions except `MixColumns`, using the same semantics as in [eclib](../blob/main//eclib/AES.ec).
I'm not sure if I should include a 'hardcoded' implementation of `MixColumns`, since one actually using the correct finite field semantics would probably be more useful.


vbgl:
Thanks for your contribution. Could you please elaborate about the MixColumns issue: what is “the correct finite field semantics”?

bgregoir:
If I remember well MixColumns is a matrix multiplication, do implement it we need the field of the elements,
at least the mul and add. For add it is simple, it is a xor, for mul you need either a huge table (256*256) or to define it properly (mul of polynomial mod an other) this can be certainly done in Coq it is harder in EC. 
Maybe another solution (at least for EC) is to add table only for the needed entries: i.e the coef of the matrix: 1, 2 and 3.
e, b, d, 9 are then needed for the inverse. 
so 6 table will be needed ....
I propose to integrate this for the moment. But it will be good to properly define the galois field in Coq and then in EC at some time.

bgregoir (approved):

bgregoir (commented):
I did not this that mixcol and invmixcol are implemented by the identity, I will clearly prefer to backtrack this to a parameter. Or to provide the correct implementation.

bshvass:
Yes, as @bgregoir says, `MixColumns` should be implemented using the proper field, by interpreting the word as a vector of field element, doing the matrix multiplication, and then reinterpreting as a word. I've tried to use mathcomp for this, but had trouble instantiating the field with the proper minimal polynomial.
Note also that the hardcode solution (the big multiplication table) is not that bad, but I need to prove that `InvMixColumns` is an inverse to `MixColumns`, and this seems infeasible without the proper field definition, though I might be wrong.
Regarding backtracking to a parameter: I don't think this worked before either? I believe it broke compilation.
Thanks for the review.

bgregoir:
"Regarding backtracking to a parameter: I don't think this worked before either" the Jasmin evaluator was raising an error. 
I think it is better that returning a wrong result :-)

bshvass:
> "Regarding backtracking to a parameter: I don't think this worked before either" the Jasmin evaluator was raising an error. I think it is better that returning a wrong result :-)
Ah OK, I was just worried about the comment at https://github.com/jasmin-lang/jasmin/blob/main/proofs/compiler/x86_instr_decl.v#L929-L936. I'll fix it.

proofs/lang/waes.v
```diff
@@ -0,0 +1,238 @@
...
+
+Import ssrnat.
+
+Local Open Scope Z_scope.
+
+(* -------------------------------------------------------------- *)
+
+Definition Sbox (v1 : u8) : u8 :=
+  match toword v1 with
+  | 0 => wrepr U8 99    | 1 => wrepr U8 124   | 2 => wrepr U8 119   | 3 => wrepr U8 123   | 4 => wrepr U8 242
```
vbgl, line 25: I believe that `wrepr U8` can be written once before the match. This might make things clearer.

bshvass:
@strub indeed. 

For now, I've reverted to parameters. I also fixed some minor mistakes.

vbgl:
The `Parameter`s break extraction. Can you please define a sensible OCaml fallback for the unimplemented functions? Do not hesitate to ask if you don’t know how to do it.

vbgl:
Really, there is nothing wrong in asking for help…

bshvass:
> Really, there is nothing wrong in asking for help…

Haha thanks! I suppose this is not valid OCaml? Can you give a pointer?

vbgl:
What you’ve written extracts the unknown constants to a failure. This is correct and is the default behavior. But this is not convenient as such a failing constant will prevent the OCaml program to ever start (the uncaught exception is raised as soon as the program is loaded).

What we want instead is a lazy failure: the exception is raised only when the unimplemented function is called. I took the liberty to amend your commit.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/749453146

vbgl:
Thanks for your contribution!

PR #321, opened 2023-01-04
randombytes syscall : fixing types + ptr copy for returning

tfaoliveira:
I tested with the distribution package of libjade, and it appears to work.

### DIFF
--- a/compiler/syscall/jasmin_syscall.h
+++ b/compiler/syscall/jasmin_syscall.h
@@ -2,7 +2,7 @@
 #ifndef JASMIN_SYSCALL
 #define JASMIN_SYSCALL
 /* FIXME this need xlen to be Uptr */
-unsigned char * __jasmin_syscall_randombytes__(unsigned char *x, uint64_t xlen)
+uint8_t* __jasmin_syscall_randombytes__(uint8_t* x, uint64_t xlen)
 asm("__jasmin_syscall_randombytes__");
 
 #endif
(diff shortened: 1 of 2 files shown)


tfaoliveira:
I don't know how to address this line: https://github.com/jasmin-lang/jasmin/blob/cac58bdaf929eafa5f5ceb87eb264cdba9fa366a/compiler/syscall/jasmin_syscall.h#L4

bgregoir (approved):

compiler/syscall/jasmin_syscall.h
```diff
@@ -2,7 +2,7 @@
 #ifndef JASMIN_SYSCALL
 #define JASMIN_SYSCALL
 /* FIXME this need xlen to be Uptr */
-unsigned char * __jasmin_syscall_randombytes__(unsigned char *x, uint64_t xlen)
+uint8_t* __jasmin_syscall_randombytes__(uint8_t* x, uint64_t xlen)
 asm("__jasmin_syscall_randombytes__");
```
vbgl, line 6: What’s the meaning of this `asm` stuff?

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/738118512

compiler/syscall/jasmin_syscall.h
tfaoliveira: I think the goal is to ensure that the function name is preserved. Maybe this was an issue in some testing environment (but I don't precisely recall the exact reason nor wrote any notes about it).

https://gcc.gnu.org/onlinedocs/gcc/Asm-Labels.html#Asm-Labels

https://clang.llvm.org/docs/AttributeReference.html#asm

vbgl: Thanks for the explanations & links.

PR #286, opened 2022-11-10
CI: test compilation to ARMv7-m4

vbgl:
This highlights some issues in the compiler.

PR #323, opened 2023-01-09
Arm stack alignment

sarranz:
The architecture parameter that aligns the stack pointer for export functions is now an optional piece of code, instead of an instruction.
Since in some architectures (ARMv7 and RISC-V for instance) we need an extra register whether we store the stack pointer in a register or in the stack, the function that chooses where to store the stack pointer now excludes certain variables (the extra register in these cases).
The solution implemented in ARM is not optimal: if the value is in [0, 4095], it can be done in one instruction, if it fits in 16 bits in two, and otherwise we use what's implemented now. But what is implemented now always works and I'll add this optimization shortly.

### DIFF
--- a/proofs/compiler/linearization.v
+++ b/proofs/compiler/linearization.v
@@ -58,45 +58,70 @@ End E.
 
 Record linearization_params {asm_op : Type} {asmop : asmOp asm_op} :=
   {
-    (* Scratch register to compute addresses. *)
+    (* Scratch register used to set up stack. *)
     lip_tmp : Ident.ident;
 
-    (* Return a linear instruction that allocates a stack frame.
-       The linear instruction [lip_allocate_stack_frame rspi sz] increases the
-       stack pointer [sz] bytes.
+    (* Variables that can't be used to save the stack pointer. *)
+    lip_not_saved_stack : seq Ident.ident;
+
+    (* Return the arguments for a linear instruction that allocates a stack
+       frame.
+       The linear instruction derived from [lip_allocate_stack_frame rspi sz]
+       decreases the stack pointer [sz] bytes.
        In symbols, it corresponds to:
-               R[rsp] := R[rsp] − sz
+               R[rsp] := R[rsp] - sz
      *)
     lip_allocate_stack_frame :
       var_i    (* Variable with stack pointer register. *)
       -> Z     (* Amount of space to allocate. *)
-      -> lvals * sopn * pexprs;
+      -> copn_args;
 
-    (* Return a linear instruction that frees a stack frame.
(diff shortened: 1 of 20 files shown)


proofs/compiler/linearization.v
```diff
@@ -58,45 +58,70 @@ End E.
...
   {
-    (* Scratch register to compute addresses. *)
+    (* Scratch register used to set up stack. *)
     lip_tmp : Ident.ident;
 
-    (* Return a linear instruction that allocates a stack frame.
-       The linear instruction [lip_allocate_stack_frame rspi sz] increases the
-       stack pointer [sz] bytes.
+    (* Variables that can't be used to save the stack pointer. *)
+    lip_not_saved_stack : seq Ident.ident;
```
vbgl, line 65: Can you please elaborate on this?

sarranz: In ARM we need a scratch register `tmp` to compute the new stack pointer. If the `rsp` is to be saved in a register `r`, then we can't use `r` as the scratch register. This is the code I wrote to set up the stack pointer for this case:
```
r := R[rsp]
R[tmp] := (R[rsp] - sz) & - wsize_size ws
R[rsp] := R[tmp]
```
When register allocation tries to find a register to store the stack pointer, it should ignore `tmp`.
Maybe a list is not needed, I don't think we will ever need more than one register.

bgregoir: Being able to express that some variable can be introduced by linearisation is exactly the role of
extra_free_register. Unfortunately it is attached to instruction. I think your solution is ok.

vbgl: I still fail to understand when this is needed. I think the following holds: when an export function allocates a stack frame, then the stack pointer is saved in the stack.

vbgl:
(I’ve just rebased, as #286 has been merged)

proofs/compiler/linearization.v
bgregoir: I think this is used when the stack point is saved into a register, it is the register r. If regalloc set r to tmp then the code is incorrect.  Export function can save the stack pointer to the stack of to a register.
Maybe this was on overkill optimisation.

vbgl: Contraposition of what I wrote before: if the stack pointer is saved into a register, then no stack allocation occur.

bgregoir (approved):
I think the PR is ok. 
I also think that it introduce some complexity in the case of export function saving rsp in a register. For my point of view this "optimisation" is overkill and is only useful for simple (toy) examples.
Since this optim add a lot of complexity in the compiler and in the proofs, I think we should remove it. But let's discuss it and do it in another PR.

PR #324, opened 2023-01-10
DIV and MOD can be used in compound assignments

vbgl:
There seems to be no justification to the current limitation.


vbgl:
and CI is happy: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/743196482

PR #329, opened 2023-01-13
Removed useless hypothesis in Array.set_neqiE

AntoineSere:
A lemma had a useless hypothesis. This PR removes it. May break some EC proofs using Jasmin arrays.


vbgl (approved):
Thanks.

PR #334, opened 2023-01-18
Add x86 instructions CLFLUSH, LFENCE, MFENCE, SFENCE

vbgl:
Draft until there is agreement on the concrete syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
   `VPMUL`
   ([PR #276](https://github.com/jasmin-lang/jasmin/pull/276)),
   `VPAVG`
-  ([PR #285](https://github.com/jasmin-lang/jasmin/pull/285)).
+  ([PR #285](https://github.com/jasmin-lang/jasmin/pull/285)),
+  `CLFLUSH`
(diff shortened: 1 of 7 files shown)


vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/752997195

bgregoir (changes requested):
You should also add some dummy operators in eclib JModel. 
Else the extraction will fail.
Expect that I approve it

vbgl:
Producing valid EasyCrypt code is tough: `() <- LFENCE;` is not accepted, the l-value must be named. But the mechanism for generating fresh variables in `toEC.ml` is not ready to deal with variables of type `unit`.

A possible work around would be to skip those instructions in extraction (maybe produce some leakage for `CLFLUSH` when extracting for CT) as they have no input, no output, and no effect on the semantics. What do you think?

vbgl:
Extraction to EC no longer fails.
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/753977355

bgregoir (approved):
I am not fan of the current solution. But it is clearly correct.

PR #335, opened 2023-01-26
Tests: some cleaning

vbgl:
I wished all test cases could be automatically verified for safety. It’s not the case yet, but this PR is a step towards this goal.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/758179917 (failed…)

### DIFF
--- a/compiler/tests/success/cmove8.jazz
+++ b/compiler/tests/success/cmove8.jazz
@@ -1,51 +1,47 @@
 export
 fn f(reg u8 b, reg u64 x y) -> reg u64 {
   reg u64 r;
-  reg u8 z;
   reg bool c;
 
   r = x;
   c = b == 1;
   r = y if c;
-  r = x if c;
 
   c = x <s y;
   r = y if c;
-  r = x if c;
 
   c = x <u y;
   r = y if c;
-  r = x if c;
 
   c = x <=s y;
   r = y if c;
-  r = x if c;
 
   c = x <=u y;
   r = y if c;
-  r = x if c;
 
   c = x == y;
   r = y if c;
-  r = x if c;
 
   c = x >=u y;
   r = y if c;
-  r = x if c;
 
   c = x >=s y;
   r = y if c;
-  r = x if c;
 
   c = x >u y;
   r = y if c;
-  r = x if c;
 
   c = x >s y;
   r = y if c;
-  r = x if c;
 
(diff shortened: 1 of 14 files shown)


vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/758207824

compiler/tests/success/cmove8.jazz
```diff
@@ -1,51 +1,47 @@
 export
 fn f(reg u8 b, reg u64 x y) -> reg u64 {
   reg u64 r;
-  reg u8 z;
   reg bool c;
 
   r = x;
   c = b == 1;
   r = y if c;
-  r = x if c;
```
eponier, line 10: Why did you remove all these lines?

compiler/tests/success/divmod.jazz
```diff
@@ -6,6 +6,9 @@ reg u32 e, f;
 reg u16 g, h;
 
 r = 0;
+a = 1;
+x = a if x <=u 0;
+x = a if x >=u 32768;
```
eponier, line 11: Why did you add this?

compiler/tests/success/pointers/test_spill_loop.jazz
```diff
@@ -8,7 +8,7 @@ fn loop(reg u64 x) -> reg u64 {
    p = s;
    sp = p;
    y = 0;
-   while (y < x) {
+   while (y < 42) {
```
eponier, line 11: Why this change?

eponier (commented):
Some of these changes are obviously good: initialization of a previously non initialized var, aligned read or write to external memory. But I don't understand the other ones. What is the idea of those other changes?

compiler/tests/success/cmove8.jazz
vbgl: Two reasons: the official one is that these lines bring nothing, they are redundant and do not add anything to the test case; the officious one is that there are some exponential behaviors in the safety checker, and the shorter test case can be checked in some reasonable amount of time.

compiler/tests/success/pointers/test_spill_loop.jazz
vbgl: To ensure termination.

bgregoir (approved):

PR #336, opened 2023-01-26
ARM: fix pretty-printing of JAL & CALL asm instructions

vbgl:
There is no CALL instruction on ARM…

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/758228456


bgregoir (approved):
thank Vincent

Issue #338, opened 2023-01-31
Uncouple parameters of states and semantics


sarranz:
The `SemPexprParams` record we have now imposes unnecessary dependencies on many definitions, `asmOp` in particular.

PR #339, opened 2023-02-01
Remove some unnecessary @s and implicit parameters

sarranz:
This is part of the task about improving how we handle parameters (#338), as a separate PR to ease reviewing.

### DIFF
--- a/proofs/compiler/dead_calls_proof.v
+++ b/proofs/compiler/dead_calls_proof.v
@@ -15,6 +15,9 @@ Context
   {spp : SemPexprParams asm_op syscall_state}.
 
 (* -------------------------------------------------------------------- *)
+
+Section CALLS.
+
 Fixpoint i_Calls (i : instr) {struct i} : Sp.t :=
   let: MkI _ i := i in i_Calls_r i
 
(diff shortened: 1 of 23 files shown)


proofs/compiler/dead_calls_proof.v
```diff
@@ -86,15 +89,17 @@ Definition CallsE :=
...
+Let Pc i := forall c, Sp.Equal (c_calls c i) (Sp.union c (c_Calls i)).
+
 Lemma c_callsE c i : Sp.Equal (c_calls c i) (Sp.union c (c_Calls i)).
 Proof.
 move: c.
-apply (@cmd_rect _ _
-         (fun i => forall c, Sp.Equal (i_calls_r c i) (Sp.union c (i_Calls_r i)))
-         (fun i => forall c, Sp.Equal (i_calls c i) (Sp.union c (i_Calls i)))
-         (fun i => forall c, Sp.Equal (c_calls c i) (Sp.union c (c_Calls i)))) => /=
+apply: (cmd_rect (Pr := Pr) (Pi := Pi) (Pc := Pc)) => /=
```
vbgl, line 100: This is ugly. Can’t we just write `(cmd_rect Pr Pi Pc)`?

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/763933359

proofs/compiler/dead_calls_proof.v
sarranz: I can make `Pr`, `Pi` and `Pc` explicit, but then we will need `cmd_rect _ _ _` when we are not doing a partial application. My impression is that these two cases are more or less equally frequent.  I have no preference.

vbgl: I see, thanks for the clarification.

vbgl (approved):

PR #341, opened 2023-02-02
Compiler: perform two rounds of range splitting

vbgl:
The first round, before expansion of register arrays, does not apply to cells of register arrays. The second round fixes this limitation.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/765177635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,10 @@
   can be expressed using equality and disequality tests
   ([PR #270](https://github.com/jasmin-lang/jasmin/pull/270)).
 
+- The live-range-splitting transformation is run a second time after
+  expansion of register arrays
+  ([PR #341](https://github.com/jasmin-lang/jasmin/pull/341)).
+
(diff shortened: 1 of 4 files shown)


vbgl:
I’ve refactored a bit to move the “live-range-splitting” pass into a separate function.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/765470868

bgregoir (approved):

eponier:
I didn't take any look to what was done here. But a natural question arises. Is it not possible to run the pass only after reg array expansion? Or do reg array expansion earlier?

bgregoir:
Vincent tried but, it does not work. The first step, reduce the liveness of some variables and allows to have a better reg array expansion.

vbgl:
The first round of live-range-splitting removes renaming assignments between arrays (e.g., `a = b` where `a` and `b` are arrays). Such assignments prevent array-expansion from properly working.

PR #343, opened 2023-02-03
Checksafety: soundly handle ArrayInit

vbgl:
Currently, the safety checker crashes when it encounters an explicit `ArrayInit` “instruction”.

Note that some of the added code is not tested as I don’t know how to exercise it…

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,8 +59,9 @@
   ([PR #313](https://github.com/jasmin-lang/jasmin/pull/313);
   fixes [#310](https://github.com/jasmin-lang/jasmin/issues/310)).
 
-- Fix to the safety checker
-  ([PR #315](https://github.com/jasmin-lang/jasmin/pull/315);
+- Fixes to the safety checker
+  ([PR #315](https://github.com/jasmin-lang/jasmin/pull/315),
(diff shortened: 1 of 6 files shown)


vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/766482100

vbgl:
Hmm, I guessed wrong the PR number…

bgregoir (commented):
Hi Vincent,
    I am not able to understand this path. It is not du to the patch itself, but to my general understanding of this code.
    If I understand well, for array_init we should forget what we known about the array (or the cell of the array). In a sense we should recover the state that we have for local variables just after entering in a function. Are we sure that your patch ensure that ?

vbgl:
Well spotted. I’ve thus added a negative test case.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/778487126

eponier:
If you want to exercise the code a bit more, you can use `s = a[(int)r:1];`.

eponier:
Well, this exercises the second case of assign_slice, but this is not the main point of this PR. Exercising the other interesting cases seems impossible to me.

vbgl:
Then I’d recommend to merge this as-is.

eponier (approved):

PR #345, opened 2023-02-06
Split parameters of the semantics into Estate, SemPexpr and SemInstr

### LINKED ISSUE #338: Uncouple parameters of states and semantics

sarranz:
Closes #338.
Having the parameters as separate types instead of nested records avoids the problems `Hint Cut` solved.
My conclusions regarding the generalization of typeclasses in ssreflect tactics:
1. `have` is indeed supposed to generalize in our case, so we need `assert`. I don't know if `have!` (in `utils.v`) is a good idea.
2. `fsetdec` does indeed introduce unnecessary dependencies: we need to `clear` the section hypothesis we don't want.
3. When the typeclass is needed to infer the type of the term, it is mandatory to give it explicitly (this is why we sometimes need `sip := sip` even in `assert`).


vbgl (approved):

vbgl:
Rebased to fix merge conflicts.
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/786618698

PR #350, opened 2023-02-14
detect number of cores to make better use of modern machines

kcning:
This patch unifies the number of OCaml build jobs and the number of Make jobs by default.

Users retain the ability to use a different number of OCaml and JS build jobs, by setting environment variable OCAMLBUILD_JOBS and JSJOBS respectively.

The behavior is as follows:

- If the user does not specify -j at the root dir, then both jasmin/Makefile and jasmin/compiler/Makefile will be invoked with one single job.

- If the user specifies a number, e.g. -j10, at the root dir, then both aforementioned Makefiles will be invoked with 10 jobs.

- If the user passes -j but does not specify a number at the root dir, then both Makefiles will be invoked with as many parallel jobs as the number of detected cores.

- Invoking jasmin/compiler/Makefile in the subdir directly has the same behavior as above, except that the root dir will not be touched.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ PREFIX   ?= /usr/local
 PREFIX   := $(abspath $(PREFIX))
 export PREFIX
 
+# detect the given number of job slots
+MAKE_PID := $(shell echo $$PPID)
+JOB_FLAG := $(shell ps T | sed -n 's/.*$(MAKE_PID).*$(MAKE).* \(-j\|--jobs=*\) *\([0-9]*\).*/-j\2/p')
+JOB_NUM  := $(or $(patsubst -j%,%,$(or $(JOB_FLAG),-j1)),$(shell nproc))
 # --------------------------------------------------------------------
 .PHONY: all build check clean install uninstall dist distcheck
 
@@ -16,8 +20,8 @@ all: build
 
 build:
 	$(MAKE) -C proofs all
-	$(MAKE) -C compiler CIL
-	$(MAKE) -C compiler all
+	$(MAKE) -C compiler CIL JOB_NUM=$(JOB_NUM)
+	$(MAKE) -C compiler all JOB_NUM=$(JOB_NUM)
 
 check:
 	$(MAKE) -C eclib check
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -1,7 +1,11 @@
 # -*- Makefile -*-
 
 # --------------------------------------------------------------------
-OCAMLBUILD_JOBS  ?= 2
+MAKE_PID := $(shell echo $$PPID)
+JOB_FLAG := $(shell ps T | sed -n 's/.*$(MAKE_PID).*$(MAKE).* \(-j\|--jobs=*\) *\([0-9]*\).*/-j\2/p')
+JOB_NUM  ?= $(or $(patsubst -j%,%,$(or $(JOB_FLAG),-j1)),$(shell nproc))
+
+OCAMLBUILD_JOBS  ?= $(JOB_NUM)
 OCAMLBUILD_BIN   ?= ocamlbuild
 OCAMLBUILD_EXTRA ?=
 OCAMLBUILD_OPTS  := -use-ocamlfind -j $(OCAMLBUILD_JOBS)
@@ -15,7 +19,7 @@ OCAMLBUILD_OPTS += $(OCAMLBUILD_EXTRA)
 OCAMLBUILD := $(OCAMLBUILD_BIN) $(OCAMLBUILD_OPTS)
 
 # --------------------------------------------------------------------
-JSJOBS    ?= 2
+JSJOBS    ?= $(JOB_NUM)
 CHECKPY   ?=
 CHECK     := $(CHECKPY) scripts/runtest --jobs="$(JSJOBS)"
 CHECK     += config/tests.config


bgregoir (approved):
If no objection, I will merge soon.

eponier:
Is it always a good thing to run on as many cores as possible? I agree the hard-coded "2" is not great, but I would prefer a patch that allows to specify the number of jobs with the standard `-j`.

eponier:
And maybe `-j` without argument would mean "as many as possible".

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/778094953

kcning:
> Is it always a good thing to run on as many cores as possible? I agree the hard-coded "2" is not great, but I would prefer a patch that allows to specify the number of jobs with the standard `-j`.

It should be possible. I'll take a look.

vbgl:
Can’t you just get access to the value that has been given to the `-j` argument?

kcning:
To my knowledge that is unfortunately not possible: https://stackoverflow.com/questions/48863762/gnu-make-check-number-of-parallel-jobs

vbgl:
The accepted solution to the question you’ve just linked seems much simpler than what you wrote.

kcning:
That works in Make targets. A variable defined in targets has limited scope and does not survive. You can give it a try.

vbgl:
> You can give it a try.

Not interested, sorry.

kcning:
I can try if I can make this patch more palatable

Makefile
```diff
@@ -9,6 +9,9 @@ PREFIX   ?= /usr/local
 PREFIX   := $(abspath $(PREFIX))
 export PREFIX
 
+# detect the given number of job slots
+JOB_FLAG 	:= $(filter -j%,$(MAKEFLAGS))
+JOB_NUM 	:= $(or $(patsubst -j%,%,$(or $(JOB_FLAG),-j1)),$(shell nproc))
```
vbgl, line 14: This is dead code, isn’t it?

vbgl:
Also, this is fishy:
https://gitlab.com/jasmin-lang/jasmin/-/jobs/3780944468#L1291
The machine that produced this log has 4 cores.

kcning:
This appears a Make version problem. A simpler example for debug:
```
 $ tree .
.
├── Makefile
└── sub
    └── Makefile

2 directories, 2 files

 $ cat Makefile
# detect the given number of job slots
JOB_FLAG 	:= $(filter -j%,$(MAKEFLAGS))
JOB_NUM 	:= $(or $(patsubst -j%,%,$(or $(JOB_FLAG),-j1)),$(shell nproc))
.PHONY: all build

all: build

# NOTE: $(MAKE) does not pass through the -j option to recursive make calls.
# To run parallel jobs, pass -j explicitly.

build:
	@echo "at root" $(JOB_NUM)

	$(MAKE) -C sub
	
 $ cat sub/Makefile
# detect the given number of job slots
JOB_FLAG 	:= $(filter -j%,$(MAKEFLAGS))
JOB_NUM 	:= $(or $(patsubst -j%,%,$(or $(JOB_FLAG),-j1)),$(shell nproc))
.PHONY: CIL

all: CIL

CIL:
	@echo at sub $(JOB_NUM)

 $ make
at root 1
make -C sub
make[1]: Entering directory '/tmp/test/sub'
at sub 1
make[1]: Leaving directory '/tmp/test/sub'

 $ make -j10
at root 10
make -C sub
make[1]: Entering directory '/tmp/test/sub'
at sub 10
make[1]: Leaving directory '/tmp/test/sub'

 $ make -j
at root 8
make -C sub
make[1]: Entering directory '/tmp/test/sub'
at sub 8
make[1]: Leaving directory '/tmp/test/sub'

 $ make --version
GNU Make 4.4
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

With a older version of Make
```
# make
at root 1
make -C sub
make[1]: Entering directory '/home/user/test/sub'
at sub 1
make[1]: Leaving directory '/home/user/test/sub'

# make -j10
at root 1
make -C sub
make[1]: Entering directory '/home/user/test/sub'
at sub 1
make[1]: Leaving directory '/home/user/test/sub'

# make -j
at root 1
make -C sub
make[1]: Entering directory '/home/user/test/sub'
at sub 1
make[1]: Leaving directory '/home/user/test/sub'

# make --version
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

kcning:
Please test the simple example on the machine with 4 cores and provide the output and its make version.

vbgl:
Please ping me when this is ready.

kcning:
I'm debugging. You reported an error on your test machine, therefore I would like to know the make version and the output of this simple example on the machine in question. Would it be possible?

kcning:
On the absence of further information, I can only conclude that your suggestion is not a good idea, because it depends on the internal implementation of Make. Here MAKEFLAGS is processed differently for Make 4.4 and 4.3 on my machine. Even if I make it work for version 4.3 as well, it's possible that Make will change its behavior w.r.t MAKEFLAGS in the future. I will therefore revert back to my original patch set, which does not depend on MAKEFLAGS.

PR #355, opened 2023-02-22
ARM: mark the export functions as .global

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/785407903

### DIFF
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -213,13 +213,18 @@ let pp_brace s = Format.sprintf "{%s}" s
 
 let pp_fun tbl fmt (fn, fd) =
   let fn = Conv.string_of_funname tbl fn in
+  let head =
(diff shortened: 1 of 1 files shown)


sarranz (approved):
Note that this is a bit different from what we do for x86, where we print all the `.global` directives at the top, but obviously this is perfectly OK as well.

vbgl:
Ah, well spotted. I don’t know whether this is important. My main concern is that the linker can find the symbol…

sarranz:
My understanding is that there is no difference between these to options, and it's good for now.
I started modularizing the printer to share more code with x86, and this is one of the parts that I abstract from both printers (printing all `.global`s at the top). I don't know if anybody cares that we print directives this way until I'm done with that other PR. It's taking me a while because I can't figure out how to load global variables properly in ARM, so I think this is good.

vbgl:
Great. Let’s merge as is then.

PR #356, opened 2023-02-23
ARM: a few fixes

vbgl:
CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/786535786


sarranz (approved):
This is very good. One extra thing we might want to add to this PR:  `JMPI` should be printed as `bx r`, not as `b r`, since `b` is only for labels (my bad).

vbgl:
Does the compiler ever emits a JMPI?

sarranz:
Yes, but apparently it's not a good idea to do so. So maybe in future we will end up removing this.

PR #362, opened 2023-02-27
Safety checker: on-demand unrolling of while loops

vbgl:
While loops with a `#bounded` annotation are optimistically unrolled during safety analysis until the loop condition can be refuted.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/790348734

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@
   ([PR #372](https://github.com/jasmin-lang/jasmin/pull/372)).
 
 - The safety checker fully unrolls `while` loops annotated as `#bounded`
+  and does not attempt at proving termination of `while` loops annotated
+  with `#no_termination_check`
   ([PR #362](https://github.com/jasmin-lang/jasmin/pull/362)).
 
 ## Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@
   `jazz2tex` tool; the `-latex` command line flag is deprecated
   ([PR #372](https://github.com/jasmin-lang/jasmin/pull/372)).
 
+- The safety checker fully unrolls `while` loops annotated as `#bounded`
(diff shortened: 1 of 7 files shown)


vbgl:
Fixed & rebased on top of #365.

CI: https://gitlab.com/jasmin-lang/jasmin/-/pipelines/791636722

bgregoir (approved):
I am agree with those changes.
Some remarks:
  - #bounded may fail (while (i < k) j = 0 #bounded while(j < i) ...)
  - [compiler/src/safety/safetyPreanalysis.ml], annot is not any more discarded. I assume it is intentional.
  - Should we also add documentation in the weaky.

vbgl:
>     * #bounded may fail (while (i < k) j = 0 #bounded while(j < i) ...)

Can you please elaborate?

>     * [compiler/src/safety/safetyPreanalysis.ml], annot is not any more discarded. I assume it is intentional.

Yes. Otherwise the analyzer cannot see the annotations.

>     * Should we also add documentation in the wiki.

Sure, as soon as this gets merged.

bgregoir:
* #bounded may fail (while (i < k) j = 0 #bounded while(j < i) ...)
 Since the first loop is not unrolled, the value of i will be abstract and it will be not possible to fully unroll the iner loop.  It was just a remark. If the user ask to do something than can't work I don't think it is problem if the analysis fail.
* Yes. Otherwise the analyzer cannot see the annotations.
 I am stupid, it is obvious.

eponier:
> Other question: should the analysis report at the end (i.e., when printing “no safety violation”) include a note telling that termination of some loops has not been checked?

I think this is a good idea indeed to report the assumptions made by the checker. I think this can be concise, but it will remind the users that they put some annotation at some point (maybe they forgot about it).

vbgl:
Done (maybe not in the most elegant way, though).

PR #381, opened 2023-03-04
Simplify the lexer a little bit

eponier:
but that are interesting in themselves


eponier:
This was mainly to test the CI. There is nothing interesting in this PR (yet).

vbgl (approved):

PR #382, opened 2023-03-09
Register allocation: register banks depend on arch

### LINKED ISSUE #201: ARM: internal error with argument of wrong size
Calling the ARM-version of the compiler on the program
```
export
fn main (reg u64 state)  {
}
```
returns
```
"", line 1 (0) to line 3 (1):
  Variable allocation: state (defined at "test_gimli.jazz", line 2 (18-23)) and r0 (defined at "", line -1 (-1)) : type mismatch
```

The problem comes from the argument being `u64` and not `u32`. Maybe the program should be rejected, but it should be in a proper way. Moreover, locations of the function and of `r0` are wrong.

vbgl:
The classification of registers between general-purpose “word”-sized registers and “vector” registers depends on the architecture.

Fixes #201

### DIFF
--- a/compiler/src/regalloc.ml
+++ b/compiler/src/regalloc.ml
@@ -44,16 +44,15 @@ let fill_in_missing_names (f: ('info, 'asm) func) : ('info, 'asm) func =
 
 type kind = Word | Vector | Flag | Unknown of ty
 
-let kind_of_type =
+let kind_of_type reg_size =
   function
-  | Bty (U (U8 | U16 | U32 | U64)) -> Word
-  | Bty (U (U128 | U256)) -> Vector
+  | Bty (U sz) -> if Wsize.wsize_cmp sz reg_size = Datatypes.Gt then Vector else Word
   | Bty Bool -> Flag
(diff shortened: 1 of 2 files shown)


vbgl:
Where am I supposed to put the test case that is expected to fail?

(ping @sarranz)

sarranz:
> Where am I supposed to put the test case that is expected to fail?
> 
> (ping @sarranz)

If we want to merge this now, maybe we can create `compiler/examples/arm-m4/fail`. But I can push and PR the restructured test directories today, if that's preferable.

vbgl:
There is no hurry, but indeed it would be nice to have a more general testing infrastructure for ARM.

sarranz:
> There is no hurry, but indeed it would be nice to have a more general testing infrastructure for ARM.

I have only the five files we discussed on Zulip to sort out and it's done.

bgregoir (approved):

vbgl:
I may also just remove the test case: I’m not sure it has a lot of value…

PR #384, opened 2023-03-10
checksafety: heuristic for bounded loops

vbgl:
Variables that are involved in the guards of bounded loops need not be put in the relational domain.


eponier (approved):
I don't understand those files at all, but the diff seems to make sense.

PR #388, opened 2023-03-13
Restructure tests

sarranz:
I split the changes in two for reviewing: the first commit moves test cases to their new directory, the second one updates the testing scripts. Maybe we want to squash them before merging.


compiler/tests/success/arm-m4/intrinsic_tst.jazz
```diff
@@ -22,9 +26,7 @@ fn tst(reg u32 arg0, reg u32 arg1) -> reg u32 {
     n, z, c = #TSTcc(arg0, arg1, !z && n == v, n, z, c); // GT
     n, z, c = #TSTcc(arg0, arg1, z || n != v, n, z, c);  // LE
 
-    arg0 = 0 if n;
-    arg0 = 0 if z;
-    arg0 = 0 if c;
+    n, z, c = #TSTcc(arg0, 3, z, n, z, c);
 
     return arg0;
```
vbgl, line 31: Caveat: most of this program is dead code!

compiler/config/tests.config
```diff
@@ -15,20 +15,30 @@ bin    = ./scripts/check-cct
...
+okdirs = examples/**/x86-64 tests/success/**/x86-64
+kodirs = tests/fail/**/x86-64
+exclude = tests/fail/warning/x86-64
+
+[test-x86-64-print]
 bin = ./scripts/parse-print-parse
-okdirs = tests/success
+okdirs = tests/success/x86-64
 
-[test-extraction]
```
vbgl, line 27: Please change the CI job referring to this name (`.gitlab-ci.yml`, line 249).

vbgl (changes requested):
Should be good once CI is fixed.

compiler/tests/success/arm-m4/intrinsic_tst.jazz
sarranz: It does sound more robust to have both, I updated them.

compiler/config/tests.config
sarranz: Done.

compiler/config/tests.config
```diff
@@ -15,20 +15,30 @@ bin    = ./scripts/check-cct
...
+[test-x86-64-extraction]
 bin = ./scripts/extract-and-check
-okdirs = !examples !tests/success
-exclude = tests/success/noextract examples/arm_m4
+okdirs = examples/**/x86-64 tests/success/**/x86-64
+exclude = tests/success/noextract
```
vbgl, line 39:
```suggestion
exclude = !tests/success/noextract
```

The subdirectories of `noextract/` should be included too.

sarranz: That's right. All excludes in these test scenarios should be recursive, maybe? (There's just one other directory we exclude, `tests/fail/warning`; it doesn't have subdirectories, but well, we don't want to test warnings for anything we may add in future.)

sarranz: I meant if I should add bangs to the other three excludes in this file, not to change the script.

vbgl: What you propose makes sense.

sarranz: Done.

vbgl (approved):

PR #390, opened 2023-03-15
CI: split the Coq job

### LINKED ISSUE #377: compiler job in CI
It would be good to be able to test the compiler even if the proofs are not working.

vbgl:
Fixes #377.

With this change, no CI job (apart from itself) depend on the `coq-proof` one, which checks all Coq proofs. Therefore, even when proofs are (transiently) broken, other jobs (testing libjade, building the tarball, etc) can go on.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,17 +45,29 @@ cache dependencies:
     | xargs nix-store --query --requisites
     | cachix push jasmin'
 
-coq:
+coq-program:
   stage: prove
   variables:
     EXTRA_NIX_ARGUMENTS: --arg coqDeps true
   extends: .common
   script:
-  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs'
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs Makefile.coq'
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs -f Makefile.coq compiler/jasmin_compiler.vo'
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -45,27 +45,37 @@ cache dependencies:
...
+coq-program:
   stage: prove
   variables:
     EXTRA_NIX_ARGUMENTS: --arg coqDeps true
   extends: .common
   script:
-  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs'
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs Makefile.coq'
+  - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C proofs -f Makefile.coq compiler/jasmin_compiler.vo'
   - nix-shell --arg inCI true $EXTRA_NIX_ARGUMENTS --run 'make -j$NIX_BUILD_CORES -C compiler CIL'
```
eponier, line 56: I was convinced that you cannot just `make CIL` in `compiler` in a fresh clone, I thought that it was needed to do something in the `proofs` folder before, but after testing it is not true. Could we just `make -C compiler CIL`? Maybe the goal is to test separately the compilation of the Coq files and the extraction?

eponier (approved):
Really nice. It will allow to do some experiments without having to fix the proofs. I just have a question about the `coq-program` job.

.gitlab-ci.yml
vbgl: OMG, you’re right. Do you think the `jasmin_compiler` coq module should be kept?

vbgl: Thanks. I’ve got the same feeling. Let me try the simplification you’ve just suggested and see what happens.

vbgl: Alright. CI confirms your findings. Thanks.

PR #391, opened 2023-03-15
Fix borderline smt calls in JModel

strub:
JModel has a borderline (in term of timeout) `smt` call that do not pass anymore when moving to Why3 1.6.

This PR simplifies this `smt` call.


vbgl (approved):
Thanks.

PR #392, opened 2023-03-16
Move the safety analyzer out of the Arch modules

vbgl:
IIRC, this function has been put there only for convenience, without a good reason.


sarranz:
This is good, I have just a cosmetic comment: I feel like `ArchCoreWithAnalyze` belongs to `arch_full.ml`, maybe under the name `ArchTools` or similar, and it's the module where all architecture-dependent analyses go?

vbgl:
What would be the other uses of this module type? If there is no reuse, I’d rather keep the definition close to the single point of use.

sarranz:
Good point, let's keep it here.

sarranz (approved):

PR #393, opened 2023-03-16
Regalloc nicer errors

eponier:
2 improvements in the error messages:
- "unknown bank" in presence of unknown types (e.g. `reg int`); actually, this error was there before at some point, but then removed
- in case of two many args or rets, we distinguish between `reg` and `xreg`, and mention the maximal allowed number

2 questions:
- is "large" precise enough in the error messages?
- I kept the name "bug_201.jazz", I don't know if I should rename it "too_many_large_args.jazz"


compiler/tests/fail/register_allocation/x86-64/var_unallocated.jazz
```diff
@@ -0,0 +1,8 @@
+export fn f () -> reg u64 {
+  reg bool b; // maybe this should be rejected by typechecking
+  reg u64 res;
+  if (b) {
```
vbgl, line 4: Why do you want that the compiler rejects this program? IMHO, the test case should be removed.

eponier: The comment is outdated, but this is a program on which compilation fails, so it seems to me that it can be added to `tests/fail`. If you want a variant that passes the safety checker, we can have instead:
```
export fn f () -> reg u64 {
  reg bool b;
  reg u64 res;

  b = true;
  res = 1;
  if (b) {
    res = 2;
  }
  return res;
}
```

vbgl: I don’t want this program as an example on which the compiler must fail.

eponier: Personally, I find it good  to have some coverage of the errors the compiler can emit. If the problem is that you don't want these tests to be considered as specification, we can add a comment stating clearly at the beginning of the file "this is totally ok to change the behavior of the compiler on this file" or something like that.

vbgl: A problem is that when the behavior changes and this program becomes accepted by the compiler, the “coverage” property will be lost, and nobody will notice.

My point is that I believe that these tests are less useful than what you think.

That being said, I do not have a strong disagreement to keeping these tests, as long as removing them when needed is OK.

vbgl (approved):

compiler/tests/fail/register_allocation/x86-64/var_unallocated.jazz
eponier: I think that now we check that tests in `fail/` actually fail, don't we? If that's the case, then the test which suddenly starts to pass will not go unnoticed. Then one has to decide whether this test succeeding is a normal consequence of the changes or revealing that something fishy is going on. Just removing the test if it is not valid or even useful anymore is clearly OK.

Issue #386, opened 2023-03-13
Checksafety imprecise on constant operations


vbgl:
The safety checker fails on the following example, whereas it succeeds when the multiplication is replaced by its result.

~~~
export fn main(){
  stack u16[3 * 4] a;
  reg u64 i;
  i = 0;
  while (i < 3 * 3) {
    inline int k;
    for k = 0 to 4 {
      a[(int) i] = 0;
      i += 1;
    }
  }
}
~~~

Issue #385, opened 2023-03-10
Checksafety crash with int shift


vbgl:
Analyzing the following program produces a crash.

~~~
export fn main(reg u64 n) {
  n += 1 >> 3;
}
~~~

> Fatal error: exception Jasmin.SafetyAbsExpr.AbsExpr(AbsDom).Binop_not_supported(_)

The analyzer might perform some constant folding.

PR #394, opened 2023-03-19
fix a bug on the handling of #VPEXTR_8 and #VPEXTR_16

bacelar:
Attempt to fix a bug in the handling of #VPEXTR_8 and #VPEXTR_16, which are currently miscompiled into "vpextrq".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,10 @@
   fixes [#385](https://github.com/jasmin-lang/jasmin/issues/385)),
   [#386](https://github.com/jasmin-lang/jasmin/issues/386)).
 
+- Fix compilation and semantics of the `VPEXTR` and `VPINSR` instructions
+  ([PR #394](https://github.com/jasmin-lang/jasmin/pull/394);
+  fixes [#395](https://github.com/jasmin-lang/jasmin/issues/395)).
+
 ## Other changes
 
 - Explicit if-then-else in flag combinations is no longer supported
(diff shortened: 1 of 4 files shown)


bacelar:
> Thanks. It seems that the semantics of the instruction is also wrong: the value of the immediate argument is truncated depending to the given size. Currently, the semantics overlooks this detail.

right, that's true. I'll fix that and return back later. It appears that the same issue might also be present in other instructions (at first sight, it appears to be the case for the VPINSR instruction -- I'll  to look at it also...)

vbgl:
Many thanks. I’ll test a bit and add a CHANGELOG entry.

vbgl:
I thought I would be able to push myself to this branch. But I can’t. So I’ll bother you Bacelar: can you please add a changelog entry? Here is what I had:

~~~
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,10 @@
   fixes [#385](https://github.com/jasmin-lang/jasmin/issues/385)),
   [#386](https://github.com/jasmin-lang/jasmin/issues/386)).

+- Fix compilation and semantics of the `VPEXTR` and `VPINSR` instructions
+  ([PR #394](https://github.com/jasmin-lang/jasmin/pull/394);
+  fixes [#395](https://github.com/jasmin-lang/jasmin/issues/395)).
+
 ## Other changes

 - Explicit if-then-else in flag combinations is no longer supported
~~~

vbgl (approved):

PR #396, opened 2023-03-20
add support for PCLMULQDQ instructions

bacelar:
This PR adds support for the carry-less multiplication instructions (PCLMULQDQ and VPCLMULQDQ).

### DIFF
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -977,8 +977,8 @@ op clmulq (x y: W64.t): W128.t =
        (iota_ 0 64).
 
 op PCLMULQDQ (v1 v2: W128.t) (k: W8.t): W128.t =
- let x0 = v1 \bits64 (W8.to_uint k %% 2) in
- let x1 = v2 \bits64 (W8.to_uint k %/ 16 %% 2) in
+ let x0 = v1 \bits64 (b2i k.[0]) in
+ let x1 = v2 \bits64 (b2i k.[4]) in
  clmulq x0 x1.
 
 abbrev [-printing] VPCLMULQDQ_128 = PCLMULQDQ.
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -964,6 +964,30 @@ abbrev [-printing] VAESENCLAST      = AESENCLAST.
 abbrev [-printing] VAESIMC          = AESIMC.
 abbrev [-printing] VAESKEYGENASSIST = AESKEYGENASSIST.
 
+(* ------------------------------------------------------------------- *)
+(* PCLMULQDQ instructions *)
+(*
+| PCLMULQDQ
+| VPCLMULQDQ  of wsize
+*)
+op clmulq (x y: W64.t): W128.t =
+ let x128 =  W128.of_int (to_uint x) in
+ foldr (fun i r => (if y.[i] then x128 `<<<` i else W128.zero) `^` r)
+       W128.zero
+       (iota_ 0 64).
+
+op PCLMULQDQ (v1 v2: W128.t) (k: int): W128.t =
+ let x1 = v1 \bits64 (k %% 2) in
+ let x2 = v2 \bits64 (k %% 16 %% 2) in
+ clmulq x1 x2.
+
+abbrev [-printing] VPCLMULQDQ_128 = PCLMULQDQ.
+
+op VPCLMULQDQ_256 (v1 v2: W256.t) (k: int): W256.t =
+ let x1 = v1 \bits64 (k %% 2) in
+ let x2 = v2 \bits64 (k %% 16 %% 2) in
+ pack2 [ W128.zero; clmulq x1 x2].
+
 
 (* -------------------------------------------------------------------- *)
 
--- a/eclib/JModel.ec
+++ b/eclib/JModel.ec
@@ -976,18 +976,16 @@ op clmulq (x y: W64.t): W128.t =
        W128.zero
        (iota_ 0 64).
 
-op PCLMULQDQ (v1 v2: W128.t) (k: int): W128.t =
- let x1 = v1 \bits64 (k %% 2) in
- let x2 = v2 \bits64 (k %% 16 %% 2) in
- clmulq x1 x2.
+op PCLMULQDQ (v1 v2: W128.t) (k: W8.t): W128.t =
+ let x0 = v1 \bits64 (W8.to_uint k %% 2) in
+ let x1 = v2 \bits64 (W8.to_uint k %/ 16 %% 2) in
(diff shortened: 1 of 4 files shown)


eclib/JModel.ec
```diff
@@ -964,6 +964,30 @@ abbrev [-printing] VAESENCLAST      = AESENCLAST.
...
+*)
+op clmulq (x y: W64.t): W128.t =
+ let x128 =  W128.of_int (to_uint x) in
+ foldr (fun i r => (if y.[i] then x128 `<<<` i else W128.zero) `^` r)
+       W128.zero
+       (iota_ 0 64).
+
+op PCLMULQDQ (v1 v2: W128.t) (k: int): W128.t =
+ let x1 = v1 \bits64 (k %% 2) in
+ let x2 = v2 \bits64 (k %% 16 %% 2) in
```
bgregoir, line 981: This line is suspicious, since k %% 16 %% 2 = k %% 2
 I think k%%16 should be k%/ 16.
I that case (k%%2) will be bit 0 of k and k %/ 16 %2 bit 4 of k.
Also why k is an int I was expected a W8.t.
But this k should be an immediate it is maybe not important.

eclib/JModel.ec
```diff
@@ -964,6 +964,30 @@ abbrev [-printing] VAESENCLAST      = AESENCLAST.
...
+ clmulq x1 x2.
+
+abbrev [-printing] VPCLMULQDQ_128 = PCLMULQDQ.
+
+op VPCLMULQDQ_256 (v1 v2: W256.t) (k: int): W256.t =
+ let x1 = v1 \bits64 (k %% 2) in
+ let x2 = v2 \bits64 (k %% 16 %% 2) in
```
bgregoir, line 988: Same remark here

proofs/compiler/x86_instr_decl.v
```diff
@@ -1767,6 +1804,23 @@ Definition Ox86_VAESKEYGENASSIST_instr :=
    (check_xmm_xmmm_imm8 U128) 3 (PrimM VAESKEYGENASSIST)
    (pp_name_ty "vaeskeygenassist" [::U128;U128;U8]).
 
+(* PCLMULDQD instructions *)
+
+Definition Ox86_PCLMULQDQ_instr := 
+  mk_instr_pp "PCLMULQDQ" [:: sword U128; sword U128; sword U8] (w_ty U128)
+    [:: E 0; E 1; E 2] [:: E 0] MSB_CLEAR (@x86_VPCLMULQDQ U128)
```
bgregoir, line 1811: MSB_CLEAR should be MSB_KEEP.

proofs/compiler/x86_instr_decl.v
```diff
@@ -1767,6 +1804,23 @@ Definition Ox86_VAESKEYGENASSIST_instr :=
...
+    (check_xmm_xmmm_imm8 U128) 3 (PrimM PCLMULQDQ)
+   (pp_name_ty "pclmulqdq" [::U128;U128;U8]).
+
+Definition Ox86_VPCLMULQDQ_instr := 
+ (fun sz =>
+   mk_instr (pp_sz "VPCLMULQDQ"%string sz) [:: sword sz; sword sz; sword U8] (w_ty sz)
+       [:: E 1; E 2; E 3] [:: E 0] MSB_CLEAR (@x86_VPCLMULQDQ sz)
```
bgregoir, line 1818: MSB_CLEAR is ok here.

eclib/JModel.ec
bgregoir: In fact, I think the extraction will fail. Because 
In jasmin the last argument is a w8 and in EC the last argument is a W8.t.

bgregoir (commented):
I have added some comment.
You should also add something in the change log

vbgl:
Reading the manual, I understand that the 256-bit variant does two parallel multiplications. Is it true? I don’t have access to any hardware with this instruction.

bgregoir:
I did not understand that, maybe I have skip a part of the manual. Where did you understand that?

eclib/JModel.ec
bacelar: yes, sorry about that. The index computation was intended to be (_ %/ _ %% _) -- but the whole EC portion was a last minute addition, and definitely very unsatisfactory :-(. I'll properly revise and test it soon.

bacelar:
...but the description is, at best, misleading...
```
Instruction:		VPCLMULQDQ ymm1, ymm2, ymm3/m256, imm8
CPUID feature flag:	VPCLMULQDQ
Description:		Carry-less multiplication of one quadword of ymm2 by one quadword
			of ymm3/m256, stores the 128-bit result in ymm1. The immediate is
			used to determine which quadwords of ymm2 and ymm3/m256 should be
			used.     
```

bgregoir (commented):
I assume that your are convince that Vincent's comment is write (for ymm)

eclib/JModel.ec
```diff
@@ -976,18 +976,16 @@ op clmulq (x y: W64.t): W128.t =
        W128.zero
        (iota_ 0 64).
 
-op PCLMULQDQ (v1 v2: W128.t) (k: int): W128.t =
- let x1 = v1 \bits64 (k %% 2) in
- let x2 = v2 \bits64 (k %% 16 %% 2) in
- clmulq x1 x2.
+op PCLMULQDQ (v1 v2: W128.t) (k: W8.t): W128.t =
+ let x0 = v1 \bits64 (W8.to_uint k %% 2) in
```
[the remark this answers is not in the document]
bacelar: yes, makes sense. I'll change it.

bgregoir:
Agree with you, I let you make the change, and thus we can merge.

vbgl:
If the `sz` argument has a single valid value, then it might not be needed.

bgregoir:
Ok Vincent is right:
"This instruction computes in parallel 4 pclmul multiplications, i.e. carryless multiplications of binary polynomials of degree at most 63, stored in 4 quadwords in
512 bit registers, as seen above. Thus, four 128 bit results are stored in the zmm register ... "

bgregoir (approved):

bgregoir:
I have not opinion on: should we keep the 256 version. 
If someone has a strong opinion against it let them speak now or forever hold their peace.

bacelar:
> I have not opinion on: should we keep the 256 version.
> If someone has a strong opinion against it let them speak now or forever hold their peace.

no strong opinion on that either. I realised that what I've suggested does not do what I intended -- so just keep it unchanged. Lets hope there's someone with a Zen3 processor that would make a good use of it... :-)

PR #397, opened 2023-03-20
Migrate compiler build from ocamlbuild to dune

maximedenes:
Main motivations:
- It makes it easier to use Jasmin as a library (e.g. to call it from a language server)
- The ocamlbuild home page encourages switching to dune, so we can expect more long term support
- Unlike some more complex projects (e.g. Coq), this migration is fairly simple. Note that only the ocaml build is migrated, not the one of Coq proofs.

Points I wasn't sure about:
- What should appear in MANIFEST files? In particular, should all `dune` files be mentioned?
- What was required from "classic display" for emacs to jump to errors? I think it does by default on dune messages, but I haven't confirmed it.


maximedenes:
Sorry, I probably asked this before, but do I need to do something specific to trigger a CI pipeline for this PR? I re-read https://github.com/jasmin-lang/jasmin/wiki/CI-on-gitlab, but it doesn't seem to say and no pipeline has been created here AFAICT: https://gitlab.com/jasmin-lang/jasmin/-/pipelines

maximedenes:
Right, that's what I wanted to evaluate. Given the first errors, I have one more question: what are the respective roles of `default.nix` vs `compiler/default.nix` (`ocamlbuild` appeared in both).

maximedenes:
Hum this time the pipeline didn't trigger after 14 minutes, not very convenient. How is the mirroring done? Is this kind of delay expected?

maximedenes:
The `eclib` failure seems related to https://github.com/EasyCrypt/easycrypt/commit/9f46b97dbeee0cccd37d058721ee4f989a7bbe48

eponier:
The `eclib` failure was fixed in https://github.com/jasmin-lang/jasmin/pull/398.

maximedenes:
Replaced by #399.

PR #398, opened 2023-03-21
[CI] EasyCrypt: use why3 1.6.0

vbgl:
Fix after https://github.com/EasyCrypt/easycrypt/pull/352


eponier (approved):
I trust you on these nix-related things. But why do we need these patches? We use old versions of EasyCrypt with the last version of Why3?

vbgl:
It’s not old. It’s the latest up-to-date released version of EasyCrypt.

eponier:
If I understand correctly, this means that these patches will be removed at the time of the next release of EasyCrypt?

vbgl:
Yes.

PR #399, opened 2023-03-21
Migrate compiler build from ocamlbuild to dune

maximedenes:
Main motivations:

- It makes it easier to use Jasmin as a library (e.g. to call it from a language server)
- The ocamlbuild home page encourages switching to dune, so we can expect more long term support
- Unlike some more complex projects (e.g. Coq), this migration is fairly simple. Note that only the ocaml build is migrated, not the one of Coq proofs.

I wasn't sure about what was required from "classic display" for emacs to jump to errors (cf change in `Makefile`). I think it does work by default on dune messages, but I haven't confirmed it.

Replaces #397.


maximedenes:
@vbgl I'm a bit puzzled: the `check` job now systematically hangs in CI, after all the tests have run. So I must have done something wrong, but I can't reproduce it locally (if I launch the same nix command, tests pass and terminate on my machine).

Any idea? Note that I didn't reproduce with exactly the same nixpkgs version and cached dependencies.

eponier:
I can reproduce locally, which does not mean I understand what happens.

maximedenes:
Ok, I had got the `apron` dependencies wrong, which made the safety checker be less precise and loop. Now working locally, let's see if CI agrees. Thanks for the hint, @eponier.

vbgl (approved):
Thanks Maxime. We may fine-tune some details (later, when need arises), but this is a good move.

Closes #350.

eponier:
CI failed because .merlin files need to be removed from MANIFESTs as well. Btw, was ocaml-lsp added to default.nix on purpose?

maximedenes:
> CI failed because .merlin files need to be removed from MANIFESTs as well. Btw, was ocaml-lsp added to default.nix on purpose?

Oops! Fixed.

PR #400, opened 2023-03-23
Build the check-safety parts as a self-contained library

vbgl:
This is an other step towards making the safety checker an independent component.


compiler/src/dune
```diff
@@ -4,3 +4,21 @@
 (menhir
  (flags "--table" "--explain")
  (modules parser))
+
+(library
+  (name jasmin_support)
+  (wrapped false)
```
maximedenes, line 10: Question: what global names will these modules have? Something like `Jasmin.Location` for example, or just `Location`? Having `wrapped` set to true is usually recommended to avoid name clashes at linking time.

compiler/src/dune
```diff
@@ -4,3 +4,21 @@
...
+  (modules annotations fInfo iInfo location syscall_t)
```
eponier, line 11: Would it make sense to put these files used at extraction time in a separate folder?

vbgl:
Thanks for your comments. I’ve just added a commit that “wraps” the jasmin-support library. TBH, I don’t like it. What do you think?

compiler/src/dune
vbgl: The hardest would be to find a name for this folder.

eponier: I don't know, something like "coq_stubs"/"extraction_stubs"/"coq_glue"?

vbgl:
New attempt: #405. Your comments are welcome there.

bgregoir:
Should we now close this PR?

vbgl:
I’d rather close #405 first.

PR #403, opened 2023-03-31
Add a `-timings` command-line flag

vbgl:


bgregoir (commented):
I think it is a first good step. I have the feeling that it would be more informative to have the time taken by the transformation. I hope this is possible to do. It should be doable by storing the previous printed time.

vbgl:
I’ll add a print of the elapsed time.

vbgl:
I’ve just rebased and added display of elapsed time.

compiler/src/utils.ml
```diff
@@ -394,6 +394,45 @@ let hierror ~loc ?funname ~kind ?sub_kind ?(internal=false) =
...
+let pp_now =
+  let open Unix in
+  let timestamp = ref (-1.) in
+  let pp_elapsed fmt now =
+    let old = !timestamp in
+    if old >= 0. then begin
+      let diff = now -. old in
+      let h, m, s, c = hmsc diff in
+      Format.fprintf fmt "|";
+      let g = false in
```
bgregoir, line 422: This two line are strange, can we replace them by let g = h > 0

compiler/src/utils.ml
```diff
@@ -394,6 +394,45 @@ let hierror ~loc ?funname ~kind ?sub_kind ?(internal=false) =
       raise (HiError err))
 
 
+(* -------------------------------------------------------------------- *)
+(** Splits a time in seconds into hours, minutes, seconds, and centiseconds.
+  Number of hours must be below one hundred. *)
+let hmsc f =
```
bgregoir, line 400: This look like localtime, the diff is in the c?
Maybe also in no day but only our.

bgregoir (commented):
Small extra comment, but I think all we discussed before is here.

compiler/src/utils.ml
vbgl: sure.

compiler/src/utils.ml
vbgl: There is some timezone for with a localtime of 1s is 11:00pm. This is not what I want here…

compiler/src/utils.ml
vbgl: Done.

bgregoir (approved):

bgregoir:
Thank Vincent.

PR #404, opened 2023-04-01
reduce complexity of stack_alloc checker for global

bgregoir:
Previous version was certainly quadratic, new version should be more linear.

### DIFF
--- a/proofs/compiler/stack_alloc.v
+++ b/proofs/compiler/stack_alloc.v
@@ -1404,62 +1404,97 @@ Definition alloc_fd p_extra mglob (fresh_reg : string -> stype -> string) (local
       |} in
   ok (swith_extra fd f_extra).
 
-Definition check_glob (m: Mvar.t (Z*wsize)) (data:seq u8) (gd:glob_decl) := 
-  let x := gd.1 in
-  match Mvar.get m x with
-  | None => false 
-  | Some (z, _) =>
-    let n := Z.to_nat z in
-    let data := drop n data in
-    match gd.2 with
-    | @Gword ws w =>
-      let s := Z.to_nat (wsize_size ws) in 
-      (s <= size data) &&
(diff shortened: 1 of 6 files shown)


proofs/compiler/stack_alloc.v
```diff
@@ -1404,62 +1404,111 @@ Definition alloc_fd p_extra mglob (fresh_reg : string -> stype -> string) (local
...
+  | xO p, l => pdrop p (pdrop p l)
+  | _, _ => [::]
+  end.
+
+Definition zdrop (A:Type) z (l:list A) := 
+  match z with
+  | Zpos p => pdrop p l 
+  | _ => l
+  end. *)
+
```
vbgl, line 1420: Please remove these comments.

proofs/compiler/stack_alloc.v
```diff
@@ -1404,62 +1404,111 @@ Definition alloc_fd p_extra mglob (fresh_reg : string -> stype -> string) (local
...
+    let '(mvar, pos, data) := globals in
+    if (pos <=? p)%Z then
       if Z.land p (wsize_size ws - 1) == 0%Z then
         let s := size_slot v in
-        ok (Mvar.set globals.1 v (p,ws), p + s)%Z
+        match ztake (p - pos) data with 
+        | None => Error (stk_ierror_no_var "bad data 1") 
+        | Some (_, data) =>  
+        match ztake s data with
+        | None =>  Error (stk_ierror_no_var "bad data 2")
```
vbgl, line 1479: Can this message reach the end user? If so, please make it clearer.

proofs/compiler/stack_alloc.v
bgregoir: Think for the remarks. All is fixed.

vbgl (approved):
This makes the execution time on one example from 42s to 2s. Very nice.

PR #405, opened 2023-04-03
Refactoring of the OCaml source tree

vbgl:
This makes easy to compile the Jasmin source code to be used as a library without the check-safety components (hence without the `apron` dependencies).

Replaces #400.


maximedenes:
Looks nice! At least for a language server, it is nice to be able to select more finely the dependencies (Jasmin without the safety checker).

Question: do you think it would make sense as a next patch to have two separate binaries for the compiler and the safety checker, and get rid of the `jasminc` library (there would be only a `jasminc` executable)?

vbgl:
Just rebased to address merge conflicts.

Thanks Maxime for the comment. Yes, the `jasminc` library is definitely a temporary hack. I intend to make the safety checker available as a separate binary.

maximedenes:
One thing I realize only now and don't like so much about this patch, is that it puts generated file in the `src` folder (IIUC). Is there a way to avoid it?

vbgl:
The message of the first commit is rather explicit about this, indeed.

To keep things in separate folders although they are mutually dependent, I think it would be necessary to split the `src/` into two libraries (to break the dependency cycle). This raises the issues of naming, wrapping etc. as discussed in #400.

PR #406, opened 2023-04-03
Various fixes to the LATEX printer

vbgl:


compiler/src/latex_printer.ml
```diff
@@ -374,15 +374,13 @@ let pp_pitem fmt pi =
...
-
-let pp_prog fmt prog =
-  F.fprintf fmt "%a" (pp_list "\n" pp_pitem) prog
+  | Prequire (from, s) ->
+    let pp_from fmt =
+      Option.may (fun name ->
+          F.fprintf fmt "%a %s " kw "from" (L.unloc name)) in
+      F.fprintf fmt "%a%a " pp_from from kw "require";
+      List.iter (fun s -> F.fprintf fmt "%S " (L.unloc s)) s;
+      F.fprintf fmt eol
```
bgregoir, line 383: Why did you change this code? 
I have the impression that replacing "\"%s\"" by "%S" is equivalent to your change.
Did I miss something?

bgregoir (commented):
Else I am agree with the other change

compiler/src/latex_printer.ml
vbgl: Missing `kw` and `eol`; better layout of the output.
I’m no longer sure about the `%S`…

bgregoir (approved):

PR #408, opened 2023-04-07
Fixes to ARM parser & printer

vbgl:
I wished I could expand the test suite, but printing of intrinsics is not correct yet: shifts are ignored.


vbgl:
I’ve just rebased do address merge conflicts.

sarranz (approved):

PR #410, opened 2023-04-07
Cleaning

vbgl:
Removes a few functions from `Utils` (either dead or redundant with batteries).

In ToEC: use the `Arch.reg_size` argument to correctly convert pointers.

Also fix the makefile from the siphash example (broken in #388).


eponier (approved):
Nice cleaning! And you removed some strange patterns too!

There are other cleaning possibilities (module `List` for instance), but this will be for other PRs.

Should I squash?

vbgl:
Feel free to squash.

PR #412, opened 2023-04-17
x86 lowering: handle shifts with explicit masking

vbgl:
This is meant as a first step before the semantics of shifts changes.

With this change, the compiler is able to correctly compile current and future jasmin programs. This will allow to update existing jasmin programs before breaking changes are introduced to the compiler.

### DIFF
--- a/compiler/tests/success/x86-64/opsizes.jazz
+++ b/compiler/tests/success/x86-64/opsizes.jazz
@@ -6,9 +6,9 @@ u = v;
 x = x + s;
 x = x + 64;
 x = x + 64s s;
-x = x >> u;
-x = x >>s 64;
-x = x >>64s u;
+x >>= u & 63;
+x = x >>s 24;
+x >>64s= u & 63;
 return x;
 }
 
--- a/proofs/compiler/x86_lowering.v
+++ b/proofs/compiler/x86_lowering.v
@@ -243,6 +243,15 @@ Definition mulr sz a b :=
     end
  end.
 
+  Definition check_shift_amount sz e :=
+    match match e with
+    | Papp2 (Oland _) a b =>
+        if is_wconst U8 b is Some n
+        then if n == x86_shift_mask sz then Some a else None
+        else None
+    | _ => None end
+    with None => Some e | x => x end.
+
 (* x =(ty) e *)
 Definition lower_cassgn_classify ty e x : lower_cassgn_t :=
   let chk (b: bool) r := if b then r else LowerAssgn in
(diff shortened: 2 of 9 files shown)


compiler/tests/success/x86-64/opsizes.jazz
```diff
@@ -6,9 +6,9 @@ u = v;
 x = x + s;
 x = x + 64;
 x = x + 64s s;
-x = x >> u;
-x = x >>s 64;
-x = x >>64s u;
+x >>= u & 63;
+x = x >>s 24;
```
eponier, line 10: Why `64` -> `24`? And why not write `x >>s= 24`?

proofs/compiler/x86_lowering.v
```diff
@@ -243,6 +243,15 @@ Definition mulr sz a b :=
...
 
+  Definition check_shift_amount sz e :=
+    match match e with
+    | Papp2 (Oland _) a b =>
+        if is_wconst U8 b is Some n
+        then if n == x86_shift_mask sz then Some a else None
+        else None
+    | _ => None end
+    with None => Some e | x => x end.
+
```
eponier, line 254: This function never returns `None`. The last line could become `with None => e | Some x => x end`.

eponier (approved):
Two comments, but otherwise it is good!

proofs/compiler/x86_lowering.v
eponier: Hum, I guess this is because of the change you want to make immediately after in PR https://github.com/jasmin-lang/jasmin/pull/413?

compiler/tests/success/x86-64/opsizes.jazz
vbgl: 1. Because 64 is bigger than 63.
2. I kept the syntax that was there before, without the compound assignment.

proofs/compiler/x86_lowering.v
vbgl: > Hum, I guess this is because of the change you want to make immediately after in PR #413?

You got it.

proofs/compiler/x86_lowering_proof.v
```diff
@@ -465,6 +465,44 @@ Section PROOF.
...
+  Qed.
+
+  Lemma check_shift_amountP sz e sa s z w :
+    check_shift_amount sz e = Some sa →
+    sem_pexpr gd s e = ok z →
+    to_word U8 z = ok w →
+    Sv.Subset (read_e sa) (read_e e) ∧
+    exists2 n, sem_pexpr gd s sa >>= to_word U8 = ok n & ∀ f (a: word sz), sem_shift f a w = sem_shift f a n.
+  Proof.
+   Ltac eop := move => > /Some_inj -> -> /= ->; split; [ reflexivity | eexists ].
```
bgregoir, line 485: Curiosity : Does the tactic eop disappear outside of the proof?

compiler/tests/success/x86-64/opsizes.jazz
bgregoir, line 10: I think it is good to add this line.
I also think we should add a line close to the previous  one, i.e : x = x >>s 64 & 63;

proofs/compiler/x86_lowering_proof.v
vbgl: Unfortunately no. But it disappear in the next PR…

compiler/tests/success/x86-64/opsizes.jazz
vbgl: Are you sure?

bgregoir: I think constant prop should remove the &63. I is good to check that nothing bad occurs in that case.

bgregoir: I agree that it is really an artificial program, but we can imagine something like:
   inline int i;
   for i = 0 to 128 {
       t[i] = t[i] >>s i & 63;
   }

bgregoir: Cool, thank

proofs/compiler/x86_lowering.v
vbgl: (Meanwhile, I have understood your comment and indeed, it was a nice suggestion. But the next PR is simpler if this function already returns an option.)

PR #413, opened 2023-04-17
Fix source semantics of shifts

vbgl:
This depends on:
  - [x] #412
  - [x] https://github.com/formosa-crypto/libjade/pull/59


bgregoir (approved):
I think we should wait for the merge of the two related PR before merging this one.
Is it right?

vbgl:
The second commit must not be merged. It is just here to show that the proposed changes to `libjade` are enough.

PR #414, opened 2023-04-19
jasminc: Add a `-slice f` command line argument

vbgl:
Slicing removes from a program functions and global variables that are not reachable from the selected functions.

This feature has been requested in the context of safety checking: it allows to analyze a single entry-point within a larger program.

Nonetheless it may be used for different purposes and in particular it makes the already available `-checkCTon` and `-ec` arguments somehow redundant.

cc @tfaoliveira

### DIFF
--- a/compiler/src/slicing.ml
+++ b/compiler/src/slicing.ml
@@ -1,0 +1,59 @@
+open Utils
+open Prog
+
+type keep = { vars : Sv.t; funs : Sf.t }
+
+let with_var k x = { k with vars = Sv.add x k.vars }
+let with_fun k f = { k with funs = Sf.add f k.funs }
+
+let inspect_gvar k { gs; gv } =
+  match gs with Slocal -> k | Sglob -> with_var k (L.unloc gv)
+
+let rec inspect_e k = function
+  | Pconst _ | Pbool _ | Parr_init _ -> k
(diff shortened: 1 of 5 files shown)


compiler/src/slicing.ml
```diff
@@ -0,0 +1,54 @@
...
+let slice fs (gd, fds) =
+  let k =
+    List.fold_left
+      (fun k fd ->
+        let n = fd.f_name in
+        if Sf.mem n k.funs then inspect_stmt k fd.f_body
+        else if List.exists (String.equal n.fn_name) fs then
+          inspect_stmt (with_fun k n) fd.f_body
+        else k)
+      { vars = Sv.empty; funs = Sf.empty }
```
bgregoir, line 50: I think the code will be simpler if funs (line 50) is initialised with something like "Sf.of_list (List.map (fun fn -> (find fn in fds))".
It has two advantages, line 47 and 48 can be removed. And we can use find to produce a proper error message if the user gives a name that does not exists.

vbgl: Right. I’ll give it a try.

bgregoir (approved):

compiler/src/slicing.ml
vbgl: Done. Thanks for the suggestion.

PR #415, opened 2023-04-19
ARM: intrinsics with shifts

vbgl:
This PR implements a few things; I can split it if required:

 - support for intrisics with shifted arguments using operator syntax, e.g., `#ADD(x, y >> 2)`
 - use this syntax in the tests and drop support for the old syntax `#ADD(x, y, #LSR(2))`
 - test the pretty-printer for ARM in the test-suite and on CI

Dropping support for the old syntax enables me to implement extraction to EasyCrypt in a simple way.


sarranz (approved):
I believe that with these changes also partly fix the problem about shift amounts being too large :) (when the amount in the source is less than 256).

vbgl:
The proof of lowering is much cleaner now. Thanks Santiago.

PR #416, opened 2023-04-20
ARM lowering: assignment of large (truncated) integers

### LINKED ISSUE #375: Bitwise negation of an immediate does not work
This example gives an internal error:
```
export
fn main() -> reg u64 {
    reg u64 x;
    x = !1;
    return x;
}
```
both in x86 and arm-m4, but at different passes.

vbgl:
Fixes #375

### DIFF
--- a/compiler/tests/success/arm-m4/bug_375.jazz
+++ b/compiler/tests/success/arm-m4/bug_375.jazz
@@ -1,0 +1,6 @@
+export
+fn main() -> reg u32 {
+    reg u32 x;
+    x = !1;
+    return x;
+}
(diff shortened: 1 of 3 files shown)


bgregoir (commented):
Hi Vincent, can you explain what was the problem?
The path suggest that 1 was viewed has a (word_of_int U64 1). I wonder if the patch is the good one. I do not understand why we obtain a U64 in this example. Maybe the good patch should be in pretyping. 
Also 375, say that we also have a problem in x86. It is still true ?

vbgl:
Pretyping infers the missing size annotation for the bitwise negation: `256u`; then inserts a word-of-int at that siz.

vbgl:
Regarding x86-64, I don’t know what the issue might have been. The given program compiles fine for x86 in all version I’ve tested.

bgregoir (approved):
Thank for the explanation.

PR #417, opened 2023-04-20
toEC: enable extraction of programs for arm-m4

vbgl:
The EasyCrypt model of ARM only declares operators of the right type. No semantics yet.


bgregoir (commented):
We can discuss of it this week. 
I think for a first step it is ok.
I wonder if we should rename JModel into JModel_x86 (but this is not backward compatible).
The next step will be to define the semantic of the operator that are assumed for the moment.
Maybe we can do a generic theory for the variation of the operator. ADDS ADD ADDcc ...

vbgl:
I’ve moved the old `JModel` into a new `JModel_x86`, and rebased.

I propose to postpone the definition of the actual semantics to an other PR.

bgregoir (approved):

PR #420, opened 2023-04-21
x86: fixes for instruction BT

vbgl:
- There is no 8-bit variant for this instruction
  - The generated assembly is now well-formed


bgregoir (approved):
Thank

PR #424, opened 2023-05-02
Change sopn.implicit_arg to var

sarranz:
From outside Coq we don't need to distinguish `reg` and `flag` implicit arguments.

### DIFF
--- a/proofs/lang/sopn.v
+++ b/proofs/lang/sopn.v
@@ -11,13 +11,8 @@ Local Unset Elimination Schemes.
 
 (* ----------------------------------------------------------------------------- *)
 
-Variant implicit_arg : Type :=
-  | IArflag of var
-  | IAreg   of var
-  .
-
 Variant arg_desc :=
-| ADImplicit  of implicit_arg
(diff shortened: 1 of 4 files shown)


proofs/lang/sopn.v
```diff
@@ -12,9 +12,7 @@ Local Unset Elimination Schemes.
 (* ----------------------------------------------------------------------------- *)
 
 Variant implicit_arg : Type :=
-  | IArflag of var
-  | IAreg   of var
-  .
+| IArgument of var.
```
vbgl, line 15: Do we really need to keep this type?

vbgl: (I mean: `implicit_arg` may just be replaced by `var`.)

sarranz: You're right, I'll remove it.

sarranz: Done.

vbgl: Thanks. Let’s wait for CI to complete.

vbgl (approved):

PR #427, opened 2023-05-03
Prune unused local functions

vbgl:
Local functions that are not reachable from any “export” functions are
removed from the program after inlining.

The compiler notoriously exhibits weird behaviors when local functions
are never called.


eponier (approved):
Should we issue a warning in some contexts?

vbgl:
IMHO, that’s a good task for a linter…

PR #428, opened 2023-05-03
ARM: minimal support for global data

vbgl:


compiler/src/pp_arm_m4.ml
```diff
@@ -33,8 +33,28 @@ let pp_reg_address_aux base disp off scal =
...
+  let m = Hashtbl.create 137 in
+  ( (fun z ->
+      match Hashtbl.find m z with
+      | n -> n
+      | exception Not_found ->
+          let n = Format.asprintf ".jasmin_global_%d" (fresh ()) in
+          Hashtbl.add m z n;
+          n),
+    fun () ->
+      Hashtbl.fold (fun z n a -> (z, n) :: a) m []
```
sarranz, line 53: Is this different from `Hashtbl.to_list`?

sarranz (approved):
Note that if we are in the code segment we can use `label + imm` as an address, I don't know if we want to do that. It is good as it is, though.

compiler/src/pp_arm_m4.ml
vbgl: Thanks. I was looking in the implementation from the stdlib…

vbgl:
Where is this `label + imm` documented? This might make the code simpler…

vbgl:
What is the correct syntax?

> jasminN1sbaj/jasmin.s:8:2: error: unsupported relocation on symbol
        adr     r1, .glob_data+4

vbgl:
Ok, that was easy and is indeed much simpler.

sarranz:
> Where is this `label + imm` documented? This might make the code simpler…

Nowhere I could find. This should be in the assembler's documentation, but I can't find it in any of them.

PR #429, opened 2023-05-03
fixed bug in to_matrix, waes

cmester0:
The `to_matrix` function in waes has an index error, this fixes that.


vbgl:
Thanks for the fix. This seems correct indeed.

vbgl (approved):

PR #430, opened 2023-05-05
Handle assignments before linearization

vbgl:
Ensure (and enforce) that there are no assignment left when reaching the linearization pass.

  - Lowering must handle all existing assignments
  - Passes that come after should not introduce new assignments.

Does this need a changelog entry?


bgregoir (approved):

vbgl:
Thanks for the feedback. Let’s merge then.

PR #432, opened 2023-05-05
Cleanup and move architecture utils

sarranz:
General cleanup (prefer `Scheme Equality`) and move some more definitions to `arch_utils` for when I come back to OTBN. Should I use `t_eq_axiom` in other files? (or remove it?)

### DIFF
--- a/proofs/arch/arch_decl.v
+++ b/proofs/arch/arch_decl.v
@@ -186,9 +186,7 @@ Scheme Equality for msb_flag.
 
 Lemma msb_flag_eq_axiom : Equality.axiom msb_flag_beq.
 Proof.
-  move=> x y;apply:(iffP idP).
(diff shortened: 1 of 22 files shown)


bgregoir (commented):
I hope this will not do too many conflicts with #423...
Also by t_eq_axiom is a tactic and not a lemma ?

sarranz:
> I hope this will not do too many conflicts with #423... Also by t_eq_axiom is a tactic and not a lemma ?
I can wait until we merge #423 if that helps, these changes are not hard to reproduce.
And that's right, it should be a lemma. I'll make this changes and use it everywhere.

vbgl (commented):
This looks right.

vbgl (approved):

PR #433, opened 2023-05-05
ziota: implement without going through nat

vbgl:
Well-founded recursion is so beautiful!

### DIFF
--- a/proofs/lang/utils.v
+++ b/proofs/lang/utils.v
@@ -3,7 +3,6 @@ From mathcomp Require Import all_ssreflect.
 From Coq.Unicode Require Import Utf8.
 Require Import ZArith Setoid Morphisms CMorphisms CRelationClasses.
 Require Import xseq oseq.
-Require Psatz.
 From mathcomp Require Import word_ssrZ.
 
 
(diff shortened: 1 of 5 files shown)


proofs/lang/utils.v
```diff
@@ -1559,7 +1559,42 @@ Notation zify := (Zcmp_le, Zcmp_lt, pify, (rwR2 (@ZleP), rwR2 (@ZltP))).
...
+Proof. case: x => // p; red; Lia.lia. Qed.
+
+Fixpoint ziota_rec (first z: Z) (H: Acc (Zwf 0) z) : seq Z :=
+  let: Acc_intro REC := H in
+  match z as z' return (∀ x : Z, Zwf 0 x z' → @Acc Z (Zwf 0) x) -> Zwf0_pred_t z' -> seq Z with
+  | Zpos p => λ REC h, first :: ziota_rec (Z.succ first) (REC (Z.pred p) h)
+  | _ => λ _ _, [::]
+  end REC (Zwf0_pred z).
+
+Definition ziota p z : seq Z :=
```
bgregoir, line 1577: Why 2 here, and not 10 of 100 ?

bgregoir: In fact it is used only internally and not for the extraction.

bgregoir (approved):

PR #434, opened 2023-05-08
Fix typos, remove dead code in runtest

sarranz:
1. Remove unused imports.
2. `parse.error` does not exist.
3. `_parse_timeout` is never used.
4. `ok   = [x for x in group if     x.success]` is never used.
5. `sp.kill()` does not exist.
6. `def gather(obj, scenario):` never uses `scenario`.

### DIFF
--- a/compiler/scripts/runtest
+++ b/compiler/scripts/runtest
@@ -1,7 +1,7 @@
 #! /usr/bin/env python3
 
 # --------------------------------------------------------------------
-import sys, os, errno, re, glob, shutil, itertools, logging
+import sys, os, re, glob, itertools, logging
 import subprocess as sp, time, datetime, socket, yaml
 import collections as cl, configparser as cp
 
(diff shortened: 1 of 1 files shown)


vbgl:
Thanks for this PR. I wonder if it makes sense to actively maintain this file as opposed to just taking the updates from the [EasyCrypt repository](https://github.com/EasyCrypt/easycrypt/blob/main/scripts/testing/runtest).

sarranz:
That could help, yes. However, they have most of these issues there, and even some others, for instance
```
scripts/testing/runtest:162:13: undefined name 'parse'
scripts/testing/runtest:329:13: undefined name 'logging'
scripts/testing/runtest:666:21: undefined name 'TIMING_DIR'
scripts/testing/runtest:159: unused function '_parse_timeout' (60% confidence)
scripts/testing/runtest:473: unused variable 'exc_type' (100% confidence)
scripts/testing/runtest:473: unused variable 'exc_value' (100% confidence)
scripts/testing/runtest:473: unused variable 'traceback' (100% confidence)
```
I can fix them and PR there, and then this repo pulls from there.

There must be a library for testing the output of an executable on some arguments, but I couldn't find any satisfactory answers. If we think it's worth some more digging, I can come up with something.

vbgl (approved):

PR #436, opened 2023-05-09
arm lowering: handle left rotation by an immediate (and other goodies)

vbgl:


vbgl:
CI failure is serious, let me fix it in a different PR.

vbgl:
Rebased on top of #437.

compiler/examples/gimli/arm-m4/gimli.jazz
```diff
@@ -32,43 +27,57 @@ fn gimli (reg u32 state) {
...
+fn gimli (reg u32 io) {
+  stack u32[12] state;
+  inline int i;
+  for i = 0 to 12 {
+    state[i] = (u32)[io + 4 * i];
+  }
+  state = gimli_ref(state);
+  for i = 0 to 12 {
+    (u32)[io + 4 * i] = state[i];
+  }
```
sarranz, line 82: We probably want
```
export
fn gimli (reg u32 io) {
  stack u32[12] state;
  inline int i;
  reg u32 tmp;

  for i = 0 to 12 {
    tmp = (u32)[io + 4 * i];
    state[i] = tmp;
  }

  state = gimli_ref(state);

  for i = 0 to 12 {
    tmp = state[i];
    (u32)[io + 4 * i] = tmp;
  }
}
```
?
Or should the compiler figure this out?

vbgl: We want the whole function to disappear and just export `gimli_ref`.

sarranz: I see. But this is not happening in either x86 or ARM, and in ARM the failure comes from moving between memory locations. At which point should the compiler realize this? Stack allocation?

vbgl: In x86, the intermediate register is introduced in lowering, which is too late IMHO.

What I’ve been dreaming of is a pass before lowering that introduces all required intermediate moves. This might be optional (enabled in “easy mode”, triggering warnings in “jasmin mode” , failing loudly in ”Peter’s mode”).

sarranz (approved):

compiler/examples/gimli/arm-m4/gimli.jazz
eponier: Maybe we need to write all these dreams down at some point. Describe the ideal architecture of the compiler. I think also of the split of linearization. And specify precisely the modes and what they enforce.

PR #439, opened 2023-05-11
Minor cleaning of arch definitions

vbgl:


bgregoir (approved):
Relate to this, I don't understand why we have this Lowering_params. 
My impression is that it is used for two purposes:
     - Provide some option to lowering 
     - Provide some fresh variables
 For the options we can use a record, some option may be ignore by the lowering depending of the arch. For the fresh variables, why we don't simply require a function for variable creation as in the other passes ?

PR #440, opened 2023-05-11
subtype and value_uincl require equal size for array

bgregoir:
WArray.cast fail if the size are not equal. subtype and WArray.uincl and value_uincl require equal size for array.


vbgl (approved):
Nice simplification.

PR #445, opened 2023-05-15
Clean generation of fresh variables

vbgl:


vbgl:
There is an issue with make-reference-arguments, but I’ll leave it as future work for now.

bgregoir (approved):

PR #447, opened 2023-05-15
Integrate SSLH

sarranz:
The source-level analysis is in `compiler/src/sct_checker_forward.ml` and `compiler/src/constraints.ml`, it's the same as in `glob_array3_slh`.

The lowering of SLH operators is in `proofs/compiler/slh_lowering.v` and `proofs/compiler/slh_lowering_proof.v`. Each architecture provides the mappings `shp_lower : SLHop -> arch_SLHop` and `to_asm : arch_SLHop -> arch_op` (as described in `slh_lowering.v`) and thus we prove that both preserve semantics.

In order to compile some of the operators we need several assembly instructions. Therefore now using `arch_extra.to_asm` on an extra instruction can return several assembly ones. The proof in  `asm_gen_proof.v` was updated to keep this in mind, using `asm_pos`: we prove that if in the linear program a label is at position `p`, in the compiled program it is at position `asm_pos p`.

`sopn` was restructured: now we have three constructors, `Oasm` (as before), `Opseudo_op` (things like `Ocopy`), and `Oslh`.
`sopn` now depends on `PointerData` and `MSFsize`, this is the reason for a lot of small changes in the OCaml side (particularly in the printers). We currently pass `X86_decl.reg_size` as `MSFsize`.


compiler/src/prog.ml
```diff
@@ -115,6 +115,7 @@ and ('len,'info,'asm) gstmt = ('len,'info,'asm) ginstr list
 type ('len,'info,'asm) gfunc = {
     f_loc  : L.t;
     f_annot: Annotations.f_annot;
+    f_user_annot : Annotations.annotations;
```
vbgl, line 118: What is this new field?

eclib/JModel_x86.ec
```diff
@@ -1,7 +1,8 @@
 (* -------------------------------------------------------------------- *)
 require import AllCore BitEncoding IntDiv SmtMap Ring List StdOrder Bool.
-(*---*) import CoreMap Map Ring.IntID IntOrder.
-require export JModel_common JArray JWord_array JMemory AES.
+(*---*) import CoreMap Map Ring.IntID IntOrder .
+require export JModel_common JArray JWord_array Jslh JMemory AES.
+export SLH64.
```
vbgl, line 5: What happens when a theory `require import` both `JModel_x86` and `JModel_m4`?

proofs/lang/wsize.v
```diff
@@ -193,3 +193,8 @@ Class PointerData := {
   Uptr : wsize;
 }.
 
+(* -------------------------------------------------------------------- *)
+Class MSFsize :=
+  {
+    msf_size : wsize;
+  }.
```
vbgl, line 200: Can someone please elaborate why this is needed? Is it conceptually different than the architecture `reg_size`?

bgregoir: I am agree with you, there are the same. but MSFsize, are used at place where reg_size does not exist.
But I don't know you we don't replace msf_size by Uptr.

eclib/JModel_x86.ec
bgregoir: Santiago, will fix this: we remove export SLH64 and SLH32 from JModel_x86 and JModel_m4.
The extraction will add a import SLH64 or SLH32 in the header depending on the architecture

proofs/lang/wsize.v
sarranz: Solved.

eclib/JModel_x86.ec
sarranz: Done.

eclib/Jslh.ec
```diff
@@ -0,0 +1,36 @@
+require import CoreInt JWord.
+
+abstract theory SLH.
+  type msf_t.
+  op of_int : int -> msf_t.
+
+  abbrev [-printing] init_msf : msf_t = of_int 0.
+
+  (* FIXME ec *)
```
sarranz, line 9: This seems to be fixed in EasyCrypt's main. Should I use the `abbrev`?

eclib/Jslh.ec
```diff
@@ -0,0 +1,37 @@
...
+  (* abbrev [-printing] move_msf (msf:msf_t) : msf_t = msf. *)
```
sarranz, line 10: This is solved in the main branch of EasyCrypt. Should I change this?

vbgl: I’d rather wait for the next release of EasyCrypt.

vbgl:
Can I squash all these commits into a single one?

compiler/tests/success/slh/x86-64/annot.jazz
```diff
@@ -0,0 +1,14 @@
+fn f(#transient reg u64 x) -> #secret reg u64 {
```
sarranz, line 1: Do I understand correctly that these tests are for the checker and not for the lowering of SSLH? (We never reach that pass because there are not export functions.)

vbgl:
I’ve just rebased to address merge conflicts.

I’ve also started to write test cases to trigger the various error cases of the sct-checker.

compiler/tests/success/slh/x86-64/annot.jazz
vbgl: Fixed.

vbgl (approved):

vbgl (approved):

PR #450, opened 2023-05-17
Refactor tests that execute Jasmin programs

vbgl:
This mostly moves things around, except the last commit which marks the `exec` directive as deprecated.

I’m not aware of any use of `exec` beyond unit tests (that are rewritten without this directive in this PR).


vbgl:
So I’ve removed the last commit, waiting for a proper replacement for the features currently available through `exec`.

bgregoir (commented):
I am sorry, but I don't understand this pull request. 
It seems that this remove all the tests for the exec functionality.
What did I missed?

vbgl:
Those tests are not removed. They are now gathered in the new `compiler/tests/exec/exec.ml` file. There is a small improvement: the tests now check that they produce the expected results.

bgregoir (approved):

PR #451, opened 2023-05-17
x86: fix semantics of SHR, RCL, and RCR

vbgl:
The manual reads: “For the SHR instruction, the OF flag is set to the most-significant bit of the *original* operand.”

I know that RCL and RCR do not compute the `CF` flag correctly, but I still do not understand the details of the issue…


vbgl:
The rotation operators have now been reworked, also in EasyCrypt.

bgregoir (approved):

PR #452, opened 2023-05-21
Varmap2

bgregoir:
The goal is to remove the depend map Fv.t.
This allows to generalise the semantic, in particular sem and psem can be fully unified.
Normally this should allows to extend array_expansion: 
    replace reg array argument and return on function call.
It should also allows to remove all the truncate stuff:
    we can use sem up to regalloc, after than I think the truncate stuff are not used
    anymore.


proofs/lang/varmap.v
```diff
@@ -0,0 +1,1178 @@
...
+  move=> [ | [w]] -> /=; auto.
+  rewrite orbF; right; eauto.
+Qed.
+
+Definition truncatable wdb ty v :=
+ match v, ty with
+ | Vbool _, sbool => true
+ | Vint _, sint   => true
+ | Varr p _, sarr p' => p == p'
+(* FIXME: echange l'ordre : (ws' <= ws)%CMP || sw_allowed *)
```
vbgl, line 56: I don’t understand this line.

vbgl (approved):

PR #453, opened 2023-05-22
Check ARM modified immediate constants before printing (and some fixes)

sarranz:
Before printing instructions that use "Modified immediate constants" we check that the immediate argument is in range. 

This is to both avoid printing (some) invalid instructions, and avoid printing instructions for which the Jasmin model differs from the spec (because using certain immediates sets the carry flag). In some cases we can accept the instruction if we force the `W` encoding, which requires that the instruction does not set flags (this is the case for `MOV`, `ADD`, and `SUB`).

The check of whether an immediate sets the carry flag is in `arm_decl.v`, so that eventually it will get used in `arm_instr_decl.v`. For the time being we use it only in `arm_lowering.v` to lower assignments of integers properly.
I also changed the stack alignment to use `BIC` instead of `AND` (which was incorrect but converted into a `BIC` by the assembler).

I added `arm_facts.v` to show that the way we classify immediates corresponds to the spec. We don't use this but I expect that we will once we add the immediate constraints to `arm_instr_decl.v`.

### DIFF
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -160,9 +160,9 @@ let pp_arm_mnemonic =
     String.lowercase_ascii mn in
   hash_to_string_core to_string
 
-let pp_mnemonic_ext (ARM_op (mn, opts)) args =
+let pp_mnemonic_ext (ARM_op (mn, opts)) suff args =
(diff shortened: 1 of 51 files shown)


vbgl (approved):
Very nice. Thanks.

PR #454, opened 2023-05-22
Do not use BatIO.to_input_channel

vbgl:
Its documentation reads:

> This function is extremely costly and is provided essentially for
> debugging purposes or for reusing legacy libraries which can't be
> adapted. As a general rule, if you can avoid using this function, don't
> use it.

Its implementation has a known bug that leads to exhaustion of available file descriptors and filesystem space (see:
https://github.com/ocaml-batteries-team/batteries-included/blob/134a8873ebdef76117978b61062c8682e3cdac35/src/batIO.ml#L710).

### DIFF
--- a/compiler/src/parseio.ml
+++ b/compiler/src/parseio.ml
@@ -6,7 +6,7 @@ module L = Lexing
 
 (* -------------------------------------------------------------------- *)
 let lexbuf_from_channel = fun name channel ->
-  let lexbuf = Lexing.from_channel channel in
+  let lexbuf = Lexing.from_function channel in
     lexbuf.Lexing.lex_curr_p <- {
         Lexing.pos_fname = name;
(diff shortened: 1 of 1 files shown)


eponier:
Actually, I have a patch somewhere doing a bit more than that. Let me find it.

eponier:
I did not find my patch. I think the gist of it was the removal of `traditional2revised`. I pushed something. There is probably some cleaning to be done.

eponier:
I did some cleaning. What do you think?

vbgl:
Looks good. Thanks. I’m going to apply the first commit to other release branches. What about the others?

eponier (approved):

eponier:
> Looks good. Thanks. I’m going to apply the first commit to other release branches. What about the others?

Functionally, it does not change anything, so I think you can backport only the first one.

PR #457, opened 2023-05-22
Regenerate instruction identifiers on conversion

vbgl:
Programs entering the “coq” side always get fresh identifiers


bgregoir (commented):
I have no strong opposition against this. But I think we want to be able to remove the need of Conv. I have the impression that this does not goes into that direction. 
My question is, did we really need to do that at each translation step?
Another solution will be to add a function to refresh the uid_loc in instructions, this function
can be used when needed.

vbgl:
Thanks for the feedback. I’ll figure out something different.
(context: current implementation of make-reference-arguments is fishy and would benefit from such freshness guarantees)

PR #458, opened 2023-05-22
eclib: do not attempt to model the parity flag

vbgl:
Current model is wrong, and there are no known use of this flag.

### DIFF
--- a/eclib/JWord.ec
+++ b/eclib/JWord.ec
@@ -898,6 +898,8 @@ abbrev (`^`) = (+^).
 
 op msb (b: t): bool = 2^(size - 1) <= (to_uint b).
 
+op lsb (w: t): bool = w.[0].
+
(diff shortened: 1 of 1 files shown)


bgregoir (approved):
I am perfectly find with this.
My impression is that we should go one step further and remove this flag in the 
Coq development too. Since this may break some existing Jasmin development we should discuss of this with user first.

eponier:
Just for the record, Santiago [removed](https://github.com/jasmin-lang/jasmin/pull/346/files) `DF` earlier this year and [asked](https://formosa-crypto.zulipchat.com/#narrow/stream/308530-Jasmin.3Adevelopment/topic/.E2.9C.94.20Should.20we.20remove.20.60PF.60.20and.20.60DF.60.3F) about `PF`.

PR #464, opened 2023-05-26
Compiler: define `v_kind` in Coq

vbgl:
This way, a few classification functions on variables can be directly implemented instead of assumed as external oracles.


bgregoir (commented):
this look nice.

vbgl:
I think I’m done with this PR.

bgregoir (approved):
I think at some point is_reg_array should be implemented in a similar way.
But I would like to keep the code of array_expension parametric on is_reg_array.
Thank a lot for this change

PR #468, opened 2023-06-01
Safety checker: minimal support for flag combinations

vbgl:
Current behavior is to crash (`assert false`).


bgregoir (approved):
We should take attention to this. If we extend the safety checker to arm.

PR #469, opened 2023-06-01
Liveness: improve for register allocation

### LINKED ISSUE #455: Register allocation bug with instrinsics
The following fails
```
export
fn main(reg u64 y) {
    reg u64 x;

    // Commenting either of the following two lines makes compilation succeed.
    x = #MOV(y);  // Using x = y; makes compilation succeed.
    x = 1;

    x = y if x == 0;
    [x] = x;
}
```
with `internal compilation error in function main: Variable allocation: y (defined at "../bug.jazz", line 2 (16-17)) and RDI (defined at "", line -1 (-1)) : variable already set`. Removing any of the four instructions makes compilation go through, as well as using an assignment instead of `MOV`.

vbgl:
Fixes #455

### DIFF
--- a/compiler/safetylib/domains/safetyProduct.ml
+++ b/compiler/safetylib/domains/safetyProduct.ml
@@ -487,7 +487,7 @@ module PIMake (PW : ProgWrap) : VDomWrap = struct
   (* We compute the dependency heuristic graph *)
   let pa_res = Pa.pa_make PW.main (Some PW.prog)
 
-  (* We compute the reflexive and transitive clojure of dp *)
+  (* We compute the reflexive and transitive closure of dp *)
   let dp = trans_closure pa_res.pa_dp
 
   (* We are relational on a variable v iff:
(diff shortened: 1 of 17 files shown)


eponier:
I don't get the idea of the second commit. I have the impression that we are less precise than before, and it gives better results?

vbgl:
Yes, I don’t really understand why we had this special case…

vbgl:
Without this change, when there is `x = y;` with `y` live afterwards and `x` dead, we tell the register allocation that it is OK to allocate both `x` and `y` to the same register. Unfortunately, the checker does not know about the semantics of move instructions.

Maybe I could fix that instead…

vbgl:
With this patch, the compiler would generate a dead `mov`, e.g., `movq %rdi, %rax`. Without this patch but with an improved checker, we would generate a silly dead `mov`, e.g., `movq %rdi, %rdi`.

Maybe it’s better to just crash…

vbgl:
I think this change is good: it simplifies code that is unnecessarily complex.

compiler/safetylib/domains/safetyProduct.ml
```diff
@@ -578,11 +578,8 @@ module PIDynMake (PW : ProgWrap) : VDomWrap = struct
...
   let ssa_main, pa_res =
     (* FIXME: code duplication! dirty hack *)
-    let is_move_op =
-      X86_arch_full.X86_core.aparams.ap_is_move_op
-    in
     let asmOp = Arch_extra.asm_opI X86_arch_full.X86_core.asm_e in
-    FSPa.fs_pa_make asmOp is_move_op PW.main
+    FSPa.fs_pa_make asmOp PW.main
 
   (* We compute the reflexive and transitive clojure of dp *)
```
bgregoir, line 584: Not so frequent that I am able to find a typo in a comment (clojure -> closure) :-)

bgregoir (approved):

compiler/safetylib/domains/safetyProduct.ml
vbgl: Fixed!

bgregoir (approved):

PR #471, opened 2023-06-02
CI: update nix & nixpkgs

vbgl:
This will make possible to use Coq 8.17 & mathcomp 1.17


bgregoir (approved):
I am not really competent for this, but I assume you know what you do ...

PR #472, opened 2023-06-05
opam: install EasyCrypt libraries

vbgl:
Two questions to whoever may review this:

 1. Is this way of selecting multiple files (`*.ec`) valid there?
 2. Is is a right place for installing easycrypt libraries?


vbgl:
About 1: the answer is no, but calling `sh` to expand the `*.ec` seems correct.

I’ve updated this PR to also fix the `opam` file: a few dependencies were missing or without a proper lower bound.

The minimal version of OCaml is also raised due to https://github.com/jasmin-lang/jasmin/pull/454/commits/ace848a9085cf50c8117c2c6d1545edfdc50b785

bgregoir (approved):

PR #474, opened 2023-06-07
Reg. allocation: remove “extra free registers”

vbgl:
These registers where use to compute the return address before we used the CALL instruction.


vbgl:
Do you have something specific in mind?

bgregoir (approved):

PR #477, opened 2023-06-13
ARM/lowering: High-level syntax for 32→64-bit multiplication

vbgl:
This is much more complex than expected as the order of the destinations is different for the jasmin operator and the ARM instruction.


bgregoir (approved):
Much better like that :-)

PR #480, opened 2023-06-16
ARM: add instructions MLA and MLS

vbgl:
Lowering introduce these instructions when compiling expressions of the form: x ± y * z.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 
 # [unreleased]
 
+## New features
+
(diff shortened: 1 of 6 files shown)


bgregoir (approved):
The arm documentation says 
  "r15 cannot be used for any of Rd, Rm, Rs, or Rn."
It seems to apply also to MUL. I don't think we already take this kind of constraint into account. But it seems doable.

bgregoir:
Since r15 is the PC, it is not so relevant...

PR #484, opened 2023-06-21
Make truncate_word opaque and add a fast path

vbgl:
There is nothing to do when source and target sizes are the same (only a type-cast to have more fun).

This improves the execution speed of the interpreter.

### DIFF
--- a/proofs/arch/asm_gen_proof.v
+++ b/proofs/arch/asm_gen_proof.v
@@ -231,7 +231,7 @@ Proof.
   t_xrbindP => hbrip.
   case ho: lea_offset => [ // | ] _ <- /=.
   case: lom => _ _ hrip _ _ _.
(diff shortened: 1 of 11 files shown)


bgregoir (commented):
Maybe a stupid question, can truncate_word_spec contains only 2 cases, such that the proofs will be simpler? I means a case for <= and one for type_error.

bgregoir (commented):
This is certainly the idea of the lemma truncate_word_le

bgregoir:
I think it is nice. 
I don't understand why you say that is make the definition opaque ?

vbgl:
There is a `Global Opaque truncate_word` at some point. It’s not really opaque, but you can no longer explicitly unfold it.

bgregoir (approved):

PR #488, opened 2023-06-28
x86: VPMULHxx instructions accept a single size suffix

vbgl:


bgregoir (commented):
I am not sur to understand the goal of this PR. 
I think there is two different thing.
   1/ remove the assert (raising a type error) in the semantic (and maybe also some stupid polymorphism)
   2/ change the syntax at the source level
I fully agree with 1, I am less sur for 2. My impression is that the previous syntax was more 
meaning full. Maybe this is imposed by the change on 1.

vbgl:
The problem is that it is possible to write the following program:
~~~
export fn test(reg u256 x) -> reg u256 {
  x = #VPMULH_4u64(x, x);
  return x;
}
~~~

The jasmin compiler is happy to compile it and produces invalid assembly.

vbgl:
I’ll try something else.

bgregoir:
I think the right way to fix this, is to externalise this type checking stuff,
this will allows the type checker to make this test.

PR #493, opened 2023-07-10
Expose pretyping API needed by the language server

maximedenes:


compiler/src/pretyping.mli
```diff
@@ -14,6 +14,14 @@ module Env : sig
   val add_from : 'asm env -> string * string -> 'asm env
   val dependencies : 'asm env -> string list list
 
+  val enter_file : 'asm env -> Annotations.pident option -> Location.t option -> string -> ('asm env * string) option
+  val exit_file : 'asm env -> 'asm env
+
+  module Funs : sig
+    val push : 'asm env -> (unit, 'asm) Prog.pfunc -> Prog.pty list -> 'asm env
+    val find : Annotations.symbol -> 'asm env -> ((unit, 'asm) Prog.pfunc * Prog.pty list) option
```
vbgl, line 22: The `Annotations.symbol` type is fishy. Should it just be `string`?

maximedenes: That's coming from the implementation (`pretyping.ml`). Do you mean that something is wrong there?

https://github.com/jasmin-lang/jasmin/blob/ec958d2a021a41a0a752e3bf2cc84dec8c91895b/compiler/src/pretyping.ml#L263

vbgl: Then everything’s right.

vbgl (approved):
I’ll merge unless CI fails.

maximedenes:
Thanks!

PR #514, opened 2023-07-12
Fix wsmul

vbgl:
The behavior was wrong on negative results.

The `wsmulP` lemma is never used, but good to have, I believe.

### DIFF
--- a/proofs/lang/word.v
+++ b/proofs/lang/word.v
@@ -650,7 +650,7 @@ Definition wumul sz (x y: word sz) :=
 
 Definition wsmul sz (x y: word sz) :=
   let n := wsigned x * wsigned y in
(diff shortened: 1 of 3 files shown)


proofs/lang/word.v
```diff
@@ -719,6 +719,49 @@ Lemma wsignedE sz (w: word sz) :
   wsigned w = if msb w then wunsigned w - wbase sz else wunsigned w.
 Proof. done. Qed.
 
+Lemma wsigned_wrepr sz z :
```
eponier, line 722: For consistency with `wunsigned_repr`, I would call it `wsigned_repr` unless we consider the current naming is wrong.

vbgl: Agreed.

eponier (approved):

PR #518, opened 2023-07-12
Enable inspection API in parser

maximedenes:
The inspection API is useful for the language server to be able to derive a concrete syntax tree (for e.g. semantic highlighting, refactorings, etc) as part of the parsing phase.

It will also help generating nicer error messages without manually crafting messages. But that part is waiting on a license clarification on the Menhir side (some relevant code we'd need to reuse is currently GPL 2).

Note: the changes in `parser.mly` are required because Menhir doesn't seem to support module definitions when it generates the `.mli` files for the interpreter API.

### DIFF
--- a/compiler/src/dune
+++ b/compiler/src/dune
@@ -4,7 +4,7 @@
  (modules lexer))
 
 (menhir
- (flags "--table" "--explain")
+ (flags "--table" "--explain" "--inspection")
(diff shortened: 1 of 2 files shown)


vbgl (approved):

maximedenes:
Thanks!

PR #520, opened 2023-07-13
Fix semantics of SHLD_16, SHRD_16, and VPSxLV

vbgl:
When the (masked) shift amount is larger than the operand size, the result is undefined. This is soundly approximated in the semantics by throwing a type error.

I still have to check whether SHLD has the same issue.

Found by random testing.


bgregoir (approved):

vbgl:
Caveat: the 16-bit variants of VPSLLV and VPSRLV belong to the instruction-set extension AVX-512 byte-and-word instructions (BW).

Issue #507, opened 2023-07-11
Support for more System Calls


Rixxc:
This is a **feature request** to support more system calls in Jasmin.

Currently Jasmin only supports the getrandom system call. For more complex application (e.g. crypto agents) it is necessary to perform systemcalls to interact with the operating system. Desired system calls for a crypto agents are:

- For interaction with the filesystem:
  - open, close
- For interactions with the filesystem and to comunicate with other programs (e.g. via pipes)
  - read, write, (sync?), (lseek?)

Rixxc:
An example program could look like this:

```
fn main()
{
    reg int fd, size;
    reg ptr u8[1337] buf;
    
    fd = #open('testfile', 'r');
    size = #read(fd, buf);
    #close(fd);
    fd = #open('testfile2', 'w');
    size = #write(fd, buf, size);
    #close(fd);
}

vbgl:
This suggest a new feature request: literal string constants.

PR #521, opened 2023-07-13
Add escaping to jasmin strings

Rixxc:
This pull request adds escaping to jasmin strings.

- Quotes are escaped using "\\""
- Backslashes are escaped as "\\\\"
- Non printable chars can be escaped using their hex values, e.g. "\x00"

### DIFF
--- a/compiler/src/lexer.mll
+++ b/compiler/src/lexer.mll
@@ -191,17 +191,13 @@ and comment lvl = parse
   | [^'\n']          { comment lvl lexbuf }
   | eof              { unterminated_comment (L.of_lexbuf lexbuf) }
 and read_string buf = parse
-  | '"'       { STRING (B.contents buf) }
-  | '\\' '\\' { Buffer.add_char buf '\\'; read_string buf lexbuf }
-  | '\\' '\"' { Buffer.add_char buf '\"'; read_string buf lexbuf }
-  (** | '\\' ['x' 'X']  { read_escaped_character buf lexbuf } *)
+  | '"'       { STRING (Scanf.unescaped (B.contents buf)) }
+  | '\\' '\"' { Buffer.add_string buf "\\\""; read_string buf lexbuf }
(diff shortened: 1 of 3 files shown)


Rixxc:
This pull request belongs to #507

Rixxc:
I've also added test cases. Please include them into your test pipeline before merging.

vbgl:
This looks overly complex. Wouldn’t a simple regular expression be enough?

Rixxc:
I've tried something like `| '"' (_* [^'\\'] | "" as s)  '"' { STRING (Scanf.unescaped s) }` which kind of works.

The problem is when you try to define multiple strings, like this:
```
u8[3] test = "\x00\\\"";
u8[4] test2 = "Test";

export fn main() -> reg u8
{
    reg u8 res tmp;
    res = 0x00;

    tmp = test[0];
    res = res ^ tmp;

    return res;
}
```

Then when trying to match the first string `_*` greedily matches until the end of the second string. I don't know how to work around that issue without specifying `_*` as being non greedy, which ocamllex does not seem to support.

Rixxc:
One way to overcome that issue would be to use this regex: `| '"' ([^'\n']* [^'\\'] | "" as s)  '"' { STRING (Scanf.unescaped s) }`.
This works, as loing as the strings are defined on different line. But it would also lead to unexpected errors if an user tries to define multiple strings on the same line.

I think the most idealistic way is to enter a seperate string parsing mode, as I've done in my solution, one the parser detects the start of a string.

vbgl:
I’ve just tried: `'"' (([^'"']|"\\\"")* as s) '"'`, and it seems promising.

An other thing that should be handled in this PR is the LATEX printer (needs two rounds of escaping!).

bgregoir:
I think this PR is not anymore necessary, due to #532.
I am correct ?

bgregoir:
@Rixxc?

Rixxc:
Yes, I think so too.

PR #534, opened 2023-07-19
Some test cases are relevant to all target architectures

vbgl:
This is a small improvement to the test suite: removes duplication; enable more test cases with ARM target; ensure some test cases are not empty.


bgregoir (commented):
Hi Vincent, 
    I think this is globally a good change.
    I do not understand why you have remove some test.
    For example, file1 and file2. The goal was to test that the require works and that double require does not pose a problem.

bgregoir:
Sorry, I am stupid...
You simply moved some files ...

bgregoir (approved):

PR #536, opened 2023-07-20
Pretyping.tt_prim: export & simplify

vbgl:
There are no good reasons to hide this function, right?


basavesh:
There is a good reason to expose it :D Looks good to me. Thanks!

basavesh (approved):

PR #549, opened 2023-07-31
Changelog cleaning

vbgl:

PR #550, opened 2023-07-31
Array-copy: no ArrayInit when LHS is a ptr

### LINKED ISSUE #499: Stack-allocation failure when using `#copy`
Compiling the following program fails:

~~~
fn aux(reg ptr u32[2] out in) -> reg ptr u32[2] {
  reg u32[2] x y;
  x = #copy_32(in);
  y[0], y[1] = x[0] * 32u x[1];
  out = #copy_32(y);
  return out;
}
export fn test(reg u32 t) -> reg u32 {
  stack u32[2] a b;
  inline int i;
  for i = 0 to 2 { a[i] = t; }
  b = aux(b, a);
  reg u32 res;
  res = 0;
  for i = 0 to 2 { t = b[i]; res |= t; }
  return res;
}
~~~

When the second `#copy` is (manually) expanded into a for loop, compilation goes through.

vbgl:
Fixes #499.

~Currently a draft as it depends of #549 to avoid conflicts. (This did not work as expected…)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@
 - Fix printing to EasyCrypt of ARMv7 instruction `bic`
   ([PR #554](https://github.com/jasmin-lang/jasmin/pull/554)).
 
+- Fix expansion of `#copy` operators when target is marked as `ptr`
+  ([PR #550](https://github.com/jasmin-lang/jasmin/pull/550);
+  fixes [#499](https://github.com/jasmin-lang/jasmin/pull/499)).
+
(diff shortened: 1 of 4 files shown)


vbgl:
The problem is that expansion of `#copy` inserts an `ArrayInit(…)` that later confuses the stack-allocation pass: killing an array pointer makes stack-alloc forget its associated region.

bgregoir (approved):

vbgl:
Rebased to address merge conflicts (in changelog).

PR #551, opened 2023-08-04
Add option to inline functions that are called only once

sarranz:
Discussing with Peter, we realized that
> There's actually a somewhat legit reason to have a function in libjade that is called just once and not inlined: say, we have a function that is used by both Dilithium and Kyber, so it's in some high-level ```common``` directory. If this function is used 3x by Dilithium, but only 1x by Kyber, we can either inline it everywhere (resulting in a significant code-size increase in Dilithium) or not inline it everywhere (leading to Kyber having a non-inlined function that is used just once).

(This particular example is easy to solve with the `inline` annotation, but this makes it easier when this happens in many different places.)

### DIFF
--- a/compiler/src/glob_options.ml
+++ b/compiler/src/glob_options.ml
@@ -140,6 +140,7 @@ let print_strings = function
   | Compiler.Inlining                    -> "inline"   , "inlining"
   | Compiler.RemoveUnusedFunction        -> "rmfunc"   , "remove unused function"
   | Compiler.Unrolling                   -> "unroll"   , "unrolling"
+  | Compiler.InlineSingleCalls           -> "inlinesinglecalls", "inline single calls"
   | Compiler.Splitting                   -> "splitting", "liverange splitting"
   | Compiler.Renaming                    -> "renaming" , "variable renaming to remove copies"
   | Compiler.RemovePhiNodes              -> "rmphi"    , "remove phi nodes introduced by splitting"
@@ -169,6 +170,8 @@ let symbol2pass =
 
 let set_sct_comp_pass s =
  sct_comp_pass := symbol2pass s
+
+let inline_single_calls = ref false
 
 let print_option p =
   let s, msg = print_strings p in
@@ -228,6 +231,9 @@ let options = [
     "-ATT", Arg.Unit (set_syntax `ATT), "use AT&T syntax (default is AT&T)"; 
     "-call-conv", Arg.Symbol (["windows"; "linux"], set_cc), ": select calling convention (default depend on host architecture)";
     "-arch", Arg.Symbol (["x86-64"; "arm-m4"], set_target_arch), ": select target arch (default is x86-64)";
+    "-inline-single-calls",
+       Arg.Set inline_single_calls,
+       ": inline functions that are called only once"
   ] @  List.map print_option Compiler.compiler_step_list @ List.map stop_after_option Compiler.compiler_step_list
 
 let usage_msg = "Usage : jasminc [option] filename"
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -1526,12 +1526,15 @@ let check_lval_pointer loc x =
   | P.Lvar x when P.is_ptr (L.unloc x).P.v_kind -> () 
   | _ -> rs_tyerror ~loc (NotAPointer x)
 
-let mk_call loc is_inline lvs f es =
+let mk_call loc inline lvs f es =
   let open P in
   begin match f.f_cc with
   | Internal -> ()
-  | Export -> if is_inline <> E.InlineFun then rs_tyerror ~loc (string_error "call to export function needs to be inlined")
-  | Subroutine _ when is_inline <> E.InlineFun ->
+  | Export ->
+    if not inline then
+      let err = string_error "call to export function needs to be inlined" in
+      rs_tyerror ~loc err
+  | Subroutine _ when not inline ->
     let check_lval = function
       | Lnone _ | Lvar _ | Lasub _ -> ()
       | Lmem _ | Laset _ -> rs_tyerror ~loc (string_error "memory/array assignment are not allowed here") in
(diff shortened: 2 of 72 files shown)


sarranz:
Should I update the CHANGELOG?

eponier (commented):
I don't know what to think about the PR. I let @vbgl and @bgregoir voice their opinion. I just suggested two fixes.

Why is it not possible to insert that code in the inlining pass directly? Because it is too early in the compilation pipeline and const prop has not been runned yet?

sarranz:
> I don't know what to think about the PR. I let @vbgl and @bgregoir voice their opinion. I just suggested two fixes.
> 
> Why is it not possible to insert that code in the inlining pass directly? Because it is too early in the compilation pipeline and const prop has not been runned yet?

Yes. Also, inline functions that make a function call don't know whether to inline it or not: they might be used inside a for loop in some other file that requires them.

compiler/src/glob_options.ml
```diff
@@ -228,6 +233,9 @@ let options = [
     "-ATT", Arg.Unit (set_syntax `ATT), "use AT&T syntax (default is AT&T)"; 
     "-call-conv", Arg.Symbol (["windows"; "linux"], set_cc), ": select calling convention (default depend on host architecture)";
     "-arch", Arg.Symbol (["x86-64"; "arm-m4"], set_target_arch), ": select target arch (default is x86-64)";
+    "-inline-single-calls",
+       Arg.Unit set_inline_single_calls,
```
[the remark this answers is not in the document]
sarranz: Done, thanks.

bgregoir (commented):
I have not a strong opinion on this new option. 
A solution will be to simply add some warning for this, thus the user can decide to add an option for forcing the inline, maybe we should fix this first.
For the implementation, I do not understand why we need a new proof here. 
I think the good way to do it is to generalise inline to take a function (deciding if the call should be inline or not, given a call instruction). Then both version can be proven using this more general version.

sarranz:
I just pushed a different implementation suggested by Benjamin.

On the usefulness of this PR:
Benjamin pointed out that if we have warnings and ask the user to write `#inline` annotations, the following unhappy scenario could happen

1. User uses `f` just once
2. Compiler issues warning.
3. User adds `#inline`.
4. Compiler issues no warning.
5. User uses `f` again.
6. Compiler issues warning again for the new use of `f` (remember we want to check this after Unrolling).

And there are cases where it is just impossible to get the desired behaviour with the annotation:

I have two `make` compilation jobs, one that sets `param int N = 1;` and one that sets `param int N = 1000;` for a program containing

```
for i = 0 to N {
    f();
}
```
I would need to manually add `#inline` and recompile.

Another scenario
```
fn g() {
    f();
}

// I would like to inline both g and f here, but I can only do g.
fn h0() {
    #inline
    g();
}

// I don't want to inline either one here.
fn h1() {
    g();
    g();
}
```

vbgl:
I think the first commit should be merged. The second one needs more discussion.

bgregoir (commented):
You did more that what I had in mind, I was not thinking to remove the inline info in the call instruction. But, I think the change make sens. 
Vincent, you want to discuss more of this?

bgregoir:
I think more change are needed before merging.
The main thing is to remove the file inline_single_call and inline_single_call_proof.
I will point more explanation in the code.

proofs/compiler/inline.v
```diff
@@ -33,7 +33,10 @@ Section INLINE.
 Context
   {wsw : WithSubWord}
   {asm_op syscall_state : Type}
-  {asmop:asmOp asm_op}.
+  {asmop:asmOp asm_op}
+  (is_inline : instr_info -> funname -> option instr_info)
```
bgregoir, line 37: To be able to remove the two new files, the idea is to replace the 
is_inline argument 
Inductive whattoinline := 
  | UseAnnot 
  | UseName of seq funname.
(side note, I think the branch for cryptoline will use an InlineAll) 

The inlining take then an argument of time whattoinline.
Furthermore, the idea to perform dead_call also in inlining 
as you did in inline_single_call, It will allow to remove the call to dead_call in compiler.

compiler/src/pretyping.ml
```diff
@@ -1594,14 +1597,7 @@ let rec tt_instr pd asmOp (env : 'asm Env.env) ((annot,pi) : S.pinstr) : 'asm En
...
       let es  = tt_exprs_cast pd env (L.loc pi) args tes in
-      let is_inline = 
-        match Annot.ensure_uniq1 "inline" Annot.none annot with
-        | Some () -> E.InlineFun
-        | None -> 
-          match f.P.f_cc with 
-          | FInfo.Internal -> E.InlineFun
-          | FInfo.Export | FInfo.Subroutine _ -> E.DoNotInline in
-      let annot = Annot.consume "inline" annot in
+      let is_inline = P.is_inline annot f.P.f_cc in
```
bgregoir, line 1600: I think if the function is annotated with inline, then the pretyping should 
add the annotation to the call.
So the is_inline function can simply look the annotation.

vbgl:
I’ll give a try to making inlining optional in a different PR.

vbgl:
Also please move the first commit to a different PR so that its technical details can be discussed independently.

sarranz:
> Also please move the first commit to a different PR so that its technical details can be discussed independently.

I created  #571

vbgl:
> I have two `make` compilation jobs, one that sets `param int N = 1;` and one that sets `param int N = 1000;` for a program containing
> 
> ```
> for i = 0 to N {
>     f();
> }
> ```
> 
> I would need to manually add `#inline` and recompile.

In this scenario, your “make compilation jobs” could as well insert the relevant `#inline` annotations.

I would like to also note that this kind of made-up scenario is a clear sign that you have no real motivation for this PR.

> 
> Another scenario
> 
> ```
> fn g() {
>     f();
> }
> 
> // I would like to inline both g and f here, but I can only do g.
> fn h0() {
>     #inline
>     g();
> }
> 
> // I don't want to inline either one here.
> fn h1() {
>     g();
>     g();
> }
> ```

In this second case, I do not understand why you would like not to inline `f`.

vbgl:
Considering the amount of merge conflicts and the lack of consensus, I’d suggest to close this PR.

PR #557, opened 2023-08-31
Implement is_reg_array in Coq

eponier:
Remove one more field from compiler params. And some cleaning too. `is_reg_array` still exists on the OCaml side (and is used), so it's not 100% clear to me that this PR is worth it, but every field removed from compiler params is a small victory.

### DIFF
--- a/compiler/src/compile.ml
+++ b/compiler/src/compile.ml
@@ -197,8 +197,6 @@ let compile (type reg regx xreg rflag cond asm_op extra_op)
     tokeep
   in
 
-  let is_reg_array x = is_reg_arr (Conv.var_of_cvar x) in
-
   let warn_extra s p =
     if s = Compiler.DeadCode_RegAllocation then
       let fds, _ = Conv.prog_of_csprog p in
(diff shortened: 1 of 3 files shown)


vbgl:
I’m not sure about the last commit: it might be good to let the heuristic/oracle/classifier of `remove_init`abstract.

eponier:
I forgot to remove `is_reg_array` from `compiler_params`, that's now done.

eponier:
> I’m not sure about the last commit: it might be good to let the heuristic/oracle/classifier of `remove_init`abstract.

It depends whether you think this can change in the future. If it might, then it's probably better to keep it abstract. Feel free to drop the last commit in this case.

vbgl:
I’ve merged the first two commits. Regarding the last one, implementing `is_reg_array` in Coq is definitely a good idea.

About making the `remove_init` pass less general, maybe Benjamin would like to comment.

eponier:
Oh, I had not understood. You mean we can both keep `remove_init` generic, but instantiate it on the Coq side. This is a good compromise indeed.

eponier:
I implemented the compromise.

vbgl (approved):

PR #558, opened 2023-09-01
Small improvements

vbgl:
- Dead-code: use an option rather than a list
 - Conv: use map instead of fold-right

### DIFF
--- a/proofs/compiler/dead_code.v
+++ b/proofs/compiler/dead_code.v
@@ -22,12 +22,15 @@ Section ASM_OP.
 
 Context `{asmop : asmOp}.
 
-Definition dead_code_c (dead_code_i: instr -> Sv.t -> cexec (Sv.t * cmd))
+Let ocat (o: option instr) (c: cmd) : cmd :=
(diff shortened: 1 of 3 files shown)


proofs/compiler/dead_code.v
```diff
@@ -22,12 +22,15 @@ Section ASM_OP.
 
 Context `{asmop : asmOp}.
 
-Definition dead_code_c (dead_code_i: instr -> Sv.t -> cexec (Sv.t * cmd))
+Let ocat (o: option instr) (c: cmd) : cmd :=
```
bgregoir, line 25: I think ocat is already defined in oseq: ocons.
For curiosity what is the goal of the Let instead of a global definition?

vbgl: No. `oseq.ocons` is something else.

`Let` ensures the definition is inlined at the end of the section.

bgregoir: You right, ocons is really conter intuitive.

bgregoir (approved):

PR #565, opened 2023-09-12
Generalize eval_cond to variables instead of flags

sarranz:


bgregoir (commented):
Hi Santiago, what is the objective of this change ?
My impression is that it restrict how eval_cond can be evaluated, it can only depend on flag.
I think this can be problematic for risc-V. If i remember well there is not flag in risc-V and condition depend directly on registers.

sarranz:
> Hi Santiago, what is the objective of this change ? My impression is that it restrict how eval_cond can be evaluated, it can only depend on flag. I think this can be problematic for risc-V. If i remember well there is not flag in risc-V and condition depend directly on registers.

My intention was to move from `eval_cond` taking a flag lookup function `rflag_t -> exec bool` to one that maps both flags and registers to values (x86 and ARM then use it composed with `to_var (T := rflag)` and RISC-V with `to_var (T := reg)`). But it sounds like I did something terribly wrong and achieved the opposite?

bgregoir:
No my bad, I read the PR in the wrong direction ....

proofs/arch/arch_decl.v
```diff
@@ -680,5 +681,5 @@ Canonical rflagv_eqType := EqType _ rflagv_eqMixin.
 Class asm (reg regx xreg rflag cond asm_op: Type) :=
   { _arch_decl   :> arch_decl reg regx xreg rflag cond
   ; _asm_op_decl :> asm_op_decl asm_op
-  ; eval_cond   : (rflag_t -> exec bool) -> cond_t -> exec bool
+  ; eval_cond   : (var -> exec value) -> cond_t -> exec bool
```
bgregoir, line 684: use the type: 
   (asm_typed_reg -> value) -> cond_t -> exec bool
 instead
  maybe "exec value" 
But I think value will work.

proofs/arch/arch_sem.v
```diff
@@ -13,7 +13,8 @@ Relation_Operators
 sem_type
 syscall syscall_sem
 label
-arch_decl.
+arch_decl
+arch_extra.
```
bgregoir, line 17: arch_sem should definitively depend on arch_extra. 
This is our trusted base and should remain as simple as possible

proofs/arch/arch_extra.v
```diff
@@ -319,3 +319,52 @@ Global Instance asm_opI : asmOp extended_op :=
     sopn.prim_string := get_prime_op }.
 
 End AsmOpI.
+
+Section OF_TO.
```
bgregoir, line 323: This is not necessary if you do the change proposed in asm_decl.

proofs/arch/arch_sem.v
```diff
@@ -80,7 +81,13 @@ Notation rflagmap := RflagMap.map.
 (* -------------------------------------------------------------------- *)
 Section SEM.
 
-Context {syscall_state : Type} {sc_sem : syscall_sem syscall_state} `{asm_d : asm} {call_conv: calling_convention}.
+Context
+  {syscall_state : Type}
+  {sc_sem : syscall_sem syscall_state}
+  `{asm_d : asm}
+  {call_conv: calling_convention}
+  {atoI : arch_toIdent}
```
bgregoir, line 89: Also not necessary with the proposed change

proofs/arch/arch_sem.v
```diff
@@ -130,13 +137,24 @@ Notation asm_result := (result error asmmem).
...
+  | ABReg f => of_rbool (asm_flag xm f)
+  end.
+
+Definition xm_get_var (xm : asmmem) (x : var) : exec value :=
+  Let tr := o2r ErrType (asm_typed_reg_of_var x) in
+  ok (get_typed_reg_value xm tr).
 
 Definition eval_cond_mem (s : asmmem) (c : cond_t) :=
-  eval_cond (st_get_rflag s) c.
+  eval_cond (xm_get_var s) c.
```
bgregoir, line 157: Just use get_typed_reg_value instead of xm_get_var.
So xm_get_var can be removed.

proofs/arch/arch_sem.v
```diff
@@ -232,8 +250,8 @@ Definition eval_asm_arg k (s: asmmem) (a: asm_arg) (ty: stype) : exec value :=
 
 Definition eval_arg_in_v (s:asmmem) (args:asm_args) (a:arg_desc) (ty:stype) : exec value :=
   match a with
-  | ADImplicit (IAreg r)   => ok (Vword (s.(asm_reg) r))
-  | ADImplicit (IArflag f) => Let b := st_get_rflag s f in ok (Vbool b)
+  | ADImplicit (IAreg x) | ADImplicit (IArflag x) =>
+      xm_get_var s (to_var x)
```
bgregoir, line 254: Why this change is needed ?

proofs/compiler/compiler_proof.v
```diff
@@ -538,6 +538,20 @@ Proof.
...
+  typed_reg_of_vari x = ok tr
+  -> var_of_asm_typed_reg tr = v_var x.
+Proof.
+  move: x => [] ?? /= h.
+  symmetry.
+  apply: asm_typed_reg_of_varI.
+  by case: (asm_typed_reg_of_var _) h => // ? [->].
+Qed.
+
+Lemma isSomeP A (o : option A) :
```
bgregoir, line 551: This lemma should be move to utils.

proofs/compiler/compiler_proof.v
```diff
@@ -538,6 +538,20 @@ Proof.
   exact: ok_callee_saved.
 Qed.
 
+Lemma typed_reg_of_variI x tr :
```
bgregoir, line 541: This lemma should be moved

proofs/compiler/x86.v
```diff
@@ -1,50 +1,65 @@
...
+  x86_instr_decl.
 
 Set   Implicit Arguments.
 Unset Strict Implicit.
 Unset Printing Implicit Defensive.
 
-Definition x86_eval_cond (get : rflag -> result error bool) (c : condt) :=
+Section WITH_PARAMS.
+
+Context {atoI : arch_toIdent}.
```
bgregoir, line 18: I think it will not be necessary anymore

bgregoir (commented):
So to conclude the PR goes in the good direction, but some change are needed.

PR #568, opened 2023-09-13
safety checker: avoid not tail-rec list concatenation

vbgl:
Note that the order of the elements in `if_conds` does not matter: this list represents a set.


vbgl:
Documentation of `append` is rather clear: it is tail-rec.
https://ocaml-batteries-team.github.io/batteries-included/hdoc2/BatList.html#VALappend

eponier:
I always forget that we use batteries.

PR #569, opened 2023-09-14
less restrictive version of infer_msf during compilation

bgregoir:
This should fix the problem discussed on zulip (Jasmin:SLH).


vbgl:
There is already a similar test case: `tests/fail/slh/x86-64/function_kills_msf_1.jazz`. Maybe this file should be moved to the `tests/success` directory.

bgregoir:
I think the fail/slh is for the type system, here what I want is to ensure that the compilation success even if the program is not speculatively well typed.

vbgl:
Compilation of `tests/fail/slh/x86-64/function_kills_msf_1.jazz` now succeeds.

PR #570, opened 2023-09-14
Generalize eval_cond to variables instead of flags

sarranz:


proofs/arch/arch_sem.v
```diff
@@ -233,7 +240,7 @@ Definition eval_asm_arg k (s: asmmem) (a: asm_arg) (ty: stype) : exec value :=
 Definition eval_arg_in_v (s:asmmem) (args:asm_args) (a:arg_desc) (ty:stype) : exec value :=
   match a with
   | ADImplicit (IAreg r)   => ok (Vword (s.(asm_reg) r))
-  | ADImplicit (IArflag f) => Let b := st_get_rflag s f in ok (Vbool b)
+  | ADImplicit (IArflag f) => ok (get_typed_reg_value s (ABReg f))
```
bgregoir, line 243: Are you sure the code is equivalent. I think st_get_rflag throw an exception if the flag in None.

sarranz: Sorry, yes there is a difference. I believe it would fail in
```
  Let vs  := eval_args_in s args idesc.(id_in) idesc.(id_tin) in
  Let t   := app_sopn _ idesc.(id_semi) vs in
```
but I'll change it back.

bgregoir (approved):

sarranz:
I hadn't pushed the unfolding of `st_get_rflag` in `asm_gen_proof`, sorry about that.

bgregoir:
So I can merge now (after completion of the merge)?

sarranz:
Yes

PR #571, opened 2023-09-15
Generalize Inlining with an oracle, remove inline_info

sarranz:


compiler/src/checkAnnot.ml
```diff
@@ -82,15 +82,20 @@ let check_no_for_loop (_, fds) =
       check_no_for_loop ~funname:f_name.fn_name f_body)
     fds
 
+let is_inlinable ir =
+  match ir with
+  | Ccall _ | Cif _ -> true
+  | _ -> false
+
```
vbgl, line 89: I don’t understand this part: if an assignment or an `Copn`  is annotated “inline”, there will be no error. Why?

compiler/src/prog.ml
```diff
@@ -463,8 +463,10 @@ let rec has_call_or_syscall_i i =
 
 and has_call_or_syscall c = List.exists has_call_or_syscall_i c
 
-let has_annot a { i_annot ; _ } =
-  List.exists (fun (k, _) -> String.equal (L.unloc k) a) i_annot
+let has_annot a { i_annot } = Annotations.has_symbol a i_annot
```
vbgl, line 466: Please keep the `; _` part (or use the dot syntax, as in `x.i_annot`).

compiler/src/checkAnnot.ml
sarranz: The idea is to move all inline releated stuff to annotations on call sites. The problem is that several passes propagate the `instr_info` of some instructions (Inlining uses the call site's `instr_info` for the assignments before and after the body, Array Initialization uses it to add `ArrayInit` assignments).
One solution would be to add a `consume_inline` field to `InstrInfo` and use it in all these pases. Another one maybe is that when we insert code we use an `instr_info` derived from the location of the one from the call site, but without the annotations. But Benjamin said that the inline annotation makes no sense for instructions other than `Ccall` or `Cif`, so changing checkAnnot means less changes.
If you think this check is important, I'll implement one of the alternatives. Or something else if there are better ways of doing this.

compiler/src/prog.ml
sarranz: Oh, sorry. What does this mean without the `; _` then? I can't find it in the documentation

vbgl: With an explicit trailing underscore, it means: “other fields are to be ignored, even if more fields are added to this record type in the future”. It is analogous to a final `| _ -> …` catch-all branch in a `match` construct.

compiler/src/checkAnnot.ml
vbgl: Oh, I see. Consuming the annotation sounds better, but might deserve an other PR (up to you).

sarranz: To me it seems reasonable, but consuming exactly the inline annotation in Array Initialization sounds too ad hoc. I asked about this on Zulip.

compiler/src/prog.ml
sarranz: Done

compiler/src/checkAnnot.ml
sarranz: I created #592 and rebased on top of that.

sarranz:
Now we keep the same check in `checkAnot`.

vbgl:
Just rebased on main to address merge conflicts.

vbgl (approved):
I’ll squash all commits once CI completes.

PR #572, opened 2023-09-15
Coq: remove uses of :> to declare instances

vbgl:
This is to prepare for future versions of Coq

### DIFF
--- a/proofs/lang/sem_params.v
+++ b/proofs/lang/sem_params.v
@@ -22,17 +22,23 @@ Require Import
    [syscall_state] as a parameter instead of a record field. *)
 Class EstateParams (syscall_state : Type) := mk_ep
   {
-    _pd :> PointerData | 1000;
-    _msf_size :> MSFsize | 1000;
+    _pd : PointerData;
+    _msf_size : MSFsize;
   }.
+
+#[global]
+Existing Instances _pd _msf_size | 1000.
 
 Arguments mk_ep {_ _}.
 
 (* Parameters needed to evaluate expressions. *)
 Class SemPexprParams := mk_spp
   {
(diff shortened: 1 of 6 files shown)


bgregoir (commented):
:> will be deprecated ?

vbgl:
No, but it will change meaning (to define a coercion instead of an instance).

vbgl:
> Warning: A coercion will be introduced instead of an instance in future
versions when using ':>' in 'Class' declarations. Replace ':>' with '::' (or
use '#[global] Existing Instance field.' for compatibility with Coq < 8.17).
Beware that the default locality for '::' is #[export], as opposed to
#[global] for ':>' currently. Add an explicit #[global] attribute to the
field if you need to keep the current behavior. For example: "Class foo := {
#[global] field :: bar }." [future-coercion-class-field,records]

proofs/lang/sem_params.v
```diff
@@ -22,18 +22,24 @@ Require Import
...
   {
-    _pd :> PointerData | 1000;
-    _msf_size :> MSFsize | 1000;
+    _pd : PointerData;
+    _msf_size : MSFsize;
   }.
 
+#[global]
+Existing Instances _pd _msf_size.
+
```
eponier, line 31: Did you drop the priority on purpose?

vbgl: No. My bad.

bgregoir:
I wait for the fix proposed by JC, and then I think we can merge.

proofs/lang/sem_params.v
vbgl: Fixed.

bgregoir (approved):

PR #574, opened 2023-09-15
Safety: do not crash on (I)DIV

### LINKED ISSUE #561: Safety checker crashes on DIV and IDIV
The safety checker thinks that these instructions take 2 arguments, while they take 3. I do not know how to fix that properly.

vbgl:
Fixes #561.

### DIFF
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -1011,8 +1011,11 @@ module Leak = struct
     let id = Sopn.get_instr_desc pd asmOp opn in
     List.pmap (fun c ->
         match c with
-        | Wsize.NotZero(sz, i) ->
-          Some (NotZero(sz, List.nth es (Conv.int_of_nat i)))
+        | Wsize.X86Division(sz, sg) ->
+          Some (NotZero(sz, List.nth es 2))
+        (* FIXME: there are more properties to check *)
+        | Wsize.InRange _ -> None
+        (* FIXME: there are properties to check *)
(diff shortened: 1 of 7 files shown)


eponier:
I don't understand the changes to toEC.ml (but I don't know this file). The second commit interprets divisions as "top", correct?

compiler/src/toEC.ml
```diff
@@ -988,14 +989,14 @@ module Leak = struct
...
-    | Pget (aa, ws, x, e) -> 
+    | Pget (aa, ws, x, e) ->
       assert (aa = Warray_.AAscale); (* NOT IMPLEMENTED *)
       let x = x.gv in
       let safe = 
         let (_s,option) = Mv.find (L.unloc x) env.vars in
         if option then Initai(ws, L.unloc x, e) :: safe 
         else safe in
-      in_bound ws x e :: safe 
+      in_bound x e :: safe
```
bgregoir, line 999: I don't understand why the ws argument of in bound is removed.
stack u64[1] t;
x = t[u16 2];
I have the impression that this code will be rejected but it should not.

vbgl: Let me drop the first commit that is unrelated and clearly bad.

eponier (approved):
We are imprecise about divisions, but this is clearly better than to crash.

vbgl:
Now it is also precise. And it checks what should be checked.

bgregoir (approved):

PR #575, opened 2023-09-15
Add implicit xreg argument

sarranz:


proofs/arch/asm_gen_proof.v
```diff
@@ -1664,7 +1715,9 @@ Proof.
       move: (hinK); rewrite /syscall_kill => hnin.
       have : Sv.In (to_var r) one_varmap.callee_saved by have := regx_in_all r; SvD.fsetdec.
       rewrite /one_varmap.callee_saved /= => /sv_of_listP /mapP [x] /hpr.
-      by move=> h /to_var_typed_regx ?; subst x; rewrite -h heqx // /X; SvD.fsetdec.
```
sarranz, line 1667: I don't know why `fsetdec` stopped working here. The only way I can get it to work again is by generalizing the variable (`move: (to_var r) => ?`), so I guess it has to do with the implicit arguments of `to_var`?

bgregoir (approved):

PR #576, opened 2023-09-19
Remove propagate inline parameters and remove CBC_O

sarranz:
There are two parts to this PR, I can split it if it's worth it.

The propagate inline pass relies on the fact that interpreting a flag combination as an expression and evaluating it is the same as interpreting the values as a booleans (`sem_pexpr` composed with `fc_sem enot eand eor eeq` is `fc_sem negb andb orb eqb`). This uses an architecture parameter but is true regardless of what it does.

I removed the `CFC_O` constructor because `combine_flags_core` is how we represent conditions (`==`, `!=`, `<=`, `<`, etc.) and there is no condition for overflow (i.e. it's inaccessible from the source). The overflow flag is needed to compute some conditions, but we don't need the condition by itself. You can see this in the definition of `cf_tbl`: `CFC_O` is not the image of any `combine_flags`.

### DIFF
--- a/proofs/compiler/arch_params_proof.v
+++ b/proofs/compiler/arch_params_proof.v
@@ -15,7 +15,6 @@ Require
   linearization
   linearization_proof
   lowering
-  propagate_inline_proof
   stack_alloc
   stack_alloc_proof
   slh_lowering_proof.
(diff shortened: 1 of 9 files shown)


vbgl (approved):
A welcome simplification.

vbgl:
It seems that `CFC_S` is also not used, right?

sarranz:
Yes, it looks like that. I'll remove it.

PR #579, opened 2023-09-21
Add monadic let for options

sarranz:
While working on lowering for OTBN and some other lowering changes related to combine flags, I felt like we could simplify the code by adding an equivalent to `Let` but for options.

I think it's clearer because we indent less, avoid intermediate definitions (to later match against `Some`), and use `subst` instead of `case` in proofs (so we don't need to care about dependencies in hypotheses).

For the definition of `let%opt` I tried using `obind` instead of matching, but it blocked reduction in many places and I felt using `match` made things simpler. I couldn't get `Option.obindP` to work, that's why I defined one outside the module. We don't seem to use `Option.obindP`, but I would like to know how to use it instead of the one I defined.

Also, we defined `VarI _ dummy_var_info`  in several places, so I moved it to `expr.v`.

### DIFF
--- a/proofs/compiler/arm_params_proof.v
+++ b/proofs/compiler/arm_params_proof.v
@@ -515,9 +515,7 @@ Proof.
     hbody hset_up ? hneq_tmp_rsp hnot_saved_stack hneq_r_rsp hgetrsp;
     subst rtype.
 
-  move: hset_up.
-  rewrite /arm_set_up_sp_register.
-  case: ifP => // hset_up _.
+  move: hset_up => /oassertP_isSome [hset_up _].
 
   have hneq_r_tmp :
     v_var r <> vtmp.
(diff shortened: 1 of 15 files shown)


proofs/compiler/arm_params_proof.v
```diff
@@ -346,8 +346,7 @@ Proof.
     + rewrite -(addn0 (size P)).
       rewrite (find_instr_skip hbody) /=.
       rewrite /eval_instr /= /with_vm /= /of_estate /=.
-      rewrite /exec_sopn /= truncate_word_u /= addn0.
-      reflexivity.
```
vbgl, line 350: The `reflexivity` is here on purpose. Please leave it.

Unless there is a guarantee somewhere that `by …` always closes a goal of the form “X = ?evar” by `eq_refl`.

sarranz: Oh I see. Done.

vbgl (approved):
Nice. Thanks.

PR #585, opened 2023-09-26
use RSB for x = (y << n) - z;

bgregoir:
Should I add something in the change log?

### DIFF
--- a/proofs/compiler/arm_lowering.v
+++ b/proofs/compiler/arm_lowering.v
@@ -130,6 +130,9 @@ Definition get_arg_shift
   else
     None.
 
+Definition is_shift (ws : wsize) (e : pexpr) := 
+  if get_arg_shift ws [::e] is None then true else false. 
+           
 Definition arg_shift
   (mn : arm_mnemonic) (ws : wsize) (e : pexprs) : arm_op * seq pexpr :=
   let '(osh, es) :=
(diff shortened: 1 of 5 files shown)


proofs/compiler/arm_lowering.v
```diff
@@ -236,11 +239,14 @@ Definition lower_Papp2_op
       Some (ADD, e0, [:: e1 ])
   | Omul (Op_w _) =>
       Some (MUL, e0, [:: e1 ])
-  | Osub (Op_w _) =>
+  | Osub (Op_w ws) =>
```
vbgl, line 242: This is not needed. There is a check above that `ws` is `reg_size`.

vbgl (commented):
Is it necessary (useful?) to have such a complex test case?

proofs/compiler/arm_lowering.v
bgregoir: I will fix that

vbgl:
And indeed, a changelog entry would be nice.

bgregoir:
I think it is as useful x = y + z << n   or  x = y - z << n;
It is not needed since we can write x = #ADD(y, z << n) or x = #SUB(y, z << n); or x = RSB(z, y << n).
But it is maybe less readable.

vbgl:
I’m still not convinced by the large test case.

bgregoir:
Sorry Vincent, I miss understood your message: 
" to have such a complex test case?"
I will fix that.

bgregoir:
Done, I think it is now very simple.
Again sorry for the miss understanding

vbgl:
No problem. Thanks for your contribution.

vbgl (approved):

PR #597, opened 2023-10-02
Optimize arm_cmd_large_arith_imm

### LINKED ISSUE #379: Optimize `arm_cmd_large_arith_imm`
The function `arm_cmd_large_arith_imm` is used to perform an arithmetic instruction with a 32 bit immediate in ARM. To do so, we split the immediate in its lower and higher 16 bits, load each half with `MOV` and `MOVT`, and then perform the operation. If the immediate is less than 2^16 then we can omit the `MOVT`, and if it's less than 4096 we can omit the `MOV` as well .

sarranz:
Now `$ ./jasminc -arch arm-m4 tests/success/arm-m4/memcmp.jazz -pasm` gives
```
memcmp:
	push	{lr}
	SUB 	r12, sp, #8
```
instead of
```
memcmp:
	push	{lr}
	MOV 	r12, #8
	MOVT	r12, #0
	SUB 	r12, sp, r12
```

Closes #379.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@
   other parameters. See `tests/success/common/test_globals.jazz`.
   ([PR #595](https://github.com/jasmin-lang/jasmin/pull/595)).
 
+- The generated code for allocating and freeing stack frames in ARM has been
+  slightly optimized: small constants are used as immediates, 16-bit or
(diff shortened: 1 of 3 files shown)


vbgl (approved):
Does this deserve a few words in the changelog?

sarranz:
> Does this deserve a few words in the changelog?

Sure, give me a minute

vbgl:
Thanks.

PR #602, opened 2023-10-04
Cleanup linearization proof

sarranz:


bgregoir (commented):
I have not read everything but I think get_varis should be rename in get_var_is or get_vars_i

sarranz:
> I have not read everything but I think get_varis should be rename in get_var_is or get_vars_i

You are right, done.

PR #603, opened 2023-10-04
add some  bewteen CT and SCT annotation

bgregoir:
This is a partial patch. But it allows to have usual annotation for CT and SCT that at used for export function compatible 
with both type system.


bgregoir:
At some point, more refactoring between the two type system should be done.

Issue #604, opened 2023-10-05
inconsistant/non uniform traitement  of global and param


bgregoir:
```
u32 x = 2;
param int x = 1;

export fn f_param() -> reg u32 {
  reg u32 r;
  r = x;
  return r;
}

param int y = 1;
u32 y = 2;

export fn f_global() -> reg u32 {
  reg u32 r;
  r = y;
  return r;
}
```
both functions return 1; I think f_global should return 2.

bgregoir:
/* -------------------------------------------------------------------- */
/* After param expansion */

global u32 x.177 = 0X2;

global u32 y.176 = 0X2;
export
fn f_param () -> (reg u32) {
  reg u32 r.179;
  
  r.179 = ((32u) 1); /* u32 */
  return (r.179);
}

export
fn f_global () -> (reg u32) {
  reg u32 r.178;
  
  r.178 = ((32u) 1); /* u32 */
  return (r.178);
}

vbgl:
Compiling with version 2023.06.3 produces the following warnings:
> "bug_604.jazz", line 2 (10-11)
warning: the variable x is already declared at "bug_604.jazz", line 1 (4-5)
"bug_604.jazz", line 11 (4-5)
warning: the variable y is already declared at "bug_604.jazz", line 10 (10-11)

Moreover, the generated assembly is as follows:

~~~
f_global:
        movl    glob_data + 0(%rip), %eax
        ret
f_param:
        movl    $1, %eax
        ret
        .data
        .p2align        5
glob_data:
      .byte 2
      .byte 0
      .byte 0
      .byte 0
~~~

So function `f_global` reads from the global `y` and returns 2, whereas `f_param` returns 1 (i.e., the value of the parameter ’x`).

PR #605, opened 2023-10-05
clearer resolution of variables + add warning

bgregoir:
this PR solve #604.

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -405,7 +405,7 @@ end  = struct
       env.e_declared := P.Spv.add v !(env.e_declared);
       push_core env v Slocal
 
-    let push_implicit (env : 'asm env) (v : P.pvar) = 
+    let push_implicit (env : 'asm env) (v : P.pvar) =
       assert (not (Map.mem v.P.v_name env.e_vars));
       push_core env v Slocal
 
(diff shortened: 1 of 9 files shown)


bgregoir:
Any comment ?

vbgl:
The refactoring looks OK. I don’t understand the new warnings though. IMHO they do not belong to pre-typing but to your IDE or the so-called linter.

bgregoir:
This is true, but we spend two hours (with Tiago) to debug a program due to this absence of warning.
Another problem, for used declaration, is that we do not keep track of the local variables declaration, this mean that it is hard to write a linter that will warn stuff after pretyping (since the info does not exists anymore).

bgregoir:
Maybe I can remove the warning from that file and store the needed info somewhere else. 
Then I can add a linter.ml that will for the moment only print that warning.
But we can extent it to print more.
Should I do that ?

compiler/src/pretyping.ml
```diff
@@ -369,30 +371,51 @@ end = struct
...
+      { env with e_decls = P.MIparam d :: env.e_decls } 
+    
+    let push_local (env : 'asm env) (v : P.pvar) =
+      env.e_declared := P.Spv.add v !(env.e_declared);
+      push_core env v Slocal
 
-    let find (x : A.symbol) (env : 'asm env) =
-      Map.Exceptionless.find x env.e_vars
-  end
+    let push_implicit (env : 'asm env) (v : P.pvar) =
```
vbgl, line 408:
```suggestion
    let push_implicit (env : 'asm env) (v : P.pvar) =
```

vbgl (approved):
Please add a CHANGELOG entry.

PR #606, opened 2023-10-05
Allow spilling MSFs to MMX

sarranz:
I didn't add anything to the CHANGELOG because we already have the SLH item, let me know if I should point this out.


bgregoir (approved):
Thank

PR #611, opened 2023-10-10
Remove dead functions after loop unrolling

### LINKED ISSUE #607: Variable allocation internal error
I have no clue what's going on here, so I don't know whether this is already covered by a different issue. The `if (false)` seems to be necessary.
```
fn f(reg ptr u64[1] a, reg u64 x) -> stack u64[1], reg u64 {
    stack ptr u64[1] s;
    s = a;
    a = s;
    return a, x;
}

export
fn main() {
    reg u64 x;
    x = 0;

    stack u64[1] a;
    a[0] = 0;

    if (false) {
        _, _ = f(a, x);
    }
}
```
gives
```
"ble.jazz", line 3 (4-10):
internal compilation error in function f:
  Variable allocation: a (defined at "ble.jazz", line 1 (20-21)) and RAX : variable already set
```

vbgl:
Fixes #607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,10 @@
   ([PR 600]((https://github.com/jasmin-lang/jasmin/pull/600);
   fixes [#599](https://github.com/jasmin-lang/jasmin/issues/599)).
 
+- Remove dead functions after loop unrolling
+  ([PR 611](https://github.com/jasmin-lang/jasmin/pull/611);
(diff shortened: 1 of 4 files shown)


sarranz:
If you think it a good idea, I can add
```
    if (false) {
        _, _ = f(a, x);
    } else {
        x += 1;
    }

    x += a[0];
    return x;
```
to the test case, similar to what the libjade code does, so that it's not completely dead code.

bgregoir (approved):

PR #614, opened 2023-10-10
Avoid explicitly using lstate records

sarranz:
Using explicit records for `lstate`s in lemmas makes changing that datatype very involved. I made some cosmetic changes to `eval_instr` in `linear_sem` to use `Let` instead of if-then-else, let me know if you think it's a bad idea.
Sorry for the string of boring refactoring PRs, hopefully I'm done.

### DIFF
--- a/proofs/lang/linear_facts.v
+++ b/proofs/lang/linear_facts.v
@@ -1,16 +1,18 @@
-From mathcomp Require Import all_ssreflect all_algebra.
-Require Import Lia Relations.
+From Coq Require Import Relations.
+From mathcomp Require Import
+  all_ssreflect
+  all_algebra.
+
 Require Import
-  utils
-  psem
-  psem_facts
-  one_varmap
-  fexpr
+  fexpr_facts
   label
   linear
-  fexpr_sem
-  fexpr_facts
-  linear_sem.
+  linear_sem
+  one_varmap
+  psem
+  psem_facts
+  sem_one_varmap
+.
 
 Set Implicit Arguments.
 Unset Strict Implicit.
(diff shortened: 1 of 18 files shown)


proofs/lang/linear_facts.v
```diff
@@ -0,0 +1,165 @@
...
+
+Lemma lsem_trans4 lp ls0 ls1 ls2 ls3 ls4 :
+  lsem lp ls0 ls1 ->
+  lsem lp ls1 ls2 ->
+  lsem lp ls2 ls3 ->
+  lsem lp ls3 ls4 ->
+  lsem lp ls0 ls4.
+Proof. by repeat (move=> /lsem_trans /[apply]). Qed.
+
+Lemma lsem_trans5 lp ls0 ls1 ls2 ls3 ls4 ls5 :
```
bgregoir, line 62: I think we can define something in a more generic way.

From mathcomp Require Import all_ssreflect.

Section Trans.

Context (A : Type) (R : A -> A -> Prop).
Context (Rtrans : forall x y z, R x y -> R y z -> R x z).

Fixpoint trans_spec_aux (a0 an:A) (l:list A) : Prop :=
  match l with
  | [::]     => R a0 an
  | an1::l => R an an1 -> trans_spec_aux a0 an1 l
  end.

Definition trans_spec (l:list A) := 
   match l with
   | [::] => True
   | [::a0] => R a0 a0 -> R a0 a0
   | a0::a1::l => R a0 a1 -> trans_spec_aux a0 a1 l
   end.

(* Another definition but need reflexivity *)
Definition trans_spec' (l:list A) := 
   match l with
   | [::] => True
   | a0::l => trans_spec_aux a0 a0 l
   end.

Lemma trans_spec_auxP a0 an l : 
  R a0 an ->
  trans_spec_aux a0 an l.
Proof. 
  elim: l an => //= an1 l hrec an h0n hnn1.
  apply/hrec;apply:Rtrans h0n hnn1.
Qed.

Lemma trans_specP l : trans_spec l.
Proof.
  case: l => //= a0 [ // | a1 l].
  apply trans_spec_auxP.
Qed.

Context (Rrefl : forall x, R x x). 

Lemma trans_spec'P l : trans_spec' l.
Proof.
  case: l => //= a0 l; apply/trans_spec_auxP/Rrefl.
Qed.

sarranz: Should I still define
```
Lemma lsem_trans3 lp ls0 ls1 ls2 ls3 :
  lsem lp ls0 ls1 ->
  lsem lp ls1 ls2 ->
  lsem lp ls2 ls3 ->
  lsem lp ls0 ls3.
Proof. exact: (trans_spec'P [:: _; _; _; _ ]). Qed.
```
or use `trans_spec'P` directly? Maybe it particularizing makes it more convenient for `eapply`, `eauto` and `Search`.

bgregoir: I have no strong opinion on that...

proofs/lang/linear_facts.v
```diff
@@ -0,0 +1,217 @@
...
+  linear_sem
+  one_varmap
+  psem
+  sem_one_varmap.
+
+Set Implicit Arguments.
+Unset Strict Implicit.
+Unset Printing Implicit Defensive.
+
+Section TRANSN.
```
bgregoir, line 18: Can you move this section to utils

bgregoir (commented):
After that, I think we can merge

proofs/lang/linear_facts.v
bgregoir: Ping!

sarranz: Done, sorry for the delay.

bgregoir:
Can you fix the conflict, I think then we can merge.

sarranz:
I fixed the conflicts. I did not propagate the changes to the stack zeroization proofs, we still use the lstate record explicitly there. That's the reason we still need the lemma I left with FIXME.

bgregoir (approved):

PR #617, opened 2023-10-16
Fix build with OCaml 5

vbgl:


bgregoir (commented):
Is it the only change that is needed ?

vbgl:
Yes it is.

bgregoir (approved):

PR #621, opened 2023-10-18
OTBN: Make AK_mem more precise

sarranz:
In OTBN instructions can only use absolute memory addresses, except for the one instruction whose only purpose is to compute PC relative addresses. I think this way of making the distinction would lead to clearer errors, let me know if you think just checking at the end is better.


sarranz:
I will see if this can be solved together with some other issue we discussed with Benjamin.

PR #623, opened 2023-10-19
Fix #622

### LINKED ISSUE #622: ARM - compilation error when copying reg ptr into the stack
When I compile the following program: 

```
u32[1] C = {1};

export fn main() -> reg u32
{
  reg ptr u32[1] Cp;
  stack ptr u32[1] Cps;
  reg u32 c;

  Cp = C;
  Cps = Cp;
  Cp = Cps;
  c = Cp[0];
  return c;
}
```

I get:
```
, line 10 (2-11):
asmgen: instruction ADD is given incompatible args.
Allowed args are:
  [[reg]; [reg]; [reg]]
  [[reg]; [reg]; [imm 32]]
```

Line 10, corresponds to ` Cps = Cp;`.

Is seems that it is not possible to move a reg ptr into the stack. Is this a bug?

bgregoir:
This patch as the advantage to fixes the current problem, but I think it is not the write way to do it. 
I think, the best will be to introduce speudo instruction in arm:
   x = #lea e 
   this one can be decomposed in a sequence of addition with shift.
   ie x = #lea (y + z << i + c) 
   ---->
      x = #ADD(y, z << i);
      x = #ADD(x, c);

   x = #lea (y + c) 
   ---->
   x = #ADD(y, c);   // if c is to large we can use x to compute it.

and 
   x = #GMOV e 
this 
   can become either a LDR or a STR or a MOV  depending of the source e and destination x.

The current patch poorly do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,10 @@
 - Fix extraction to easycrypt on "for loop" that modifies the loop counter
   ([PR 616](https://github.com/jasmin-lang/jasmin/pull/616).
 
+- Fix instruction selection for stack-allocation on ARM
+  ([PR 623](https://github.com/jasmin-lang/jasmin/pull/623);
+  fixes [#622](https://github.com/jasmin-lang/jasmin/issues/622)).
+
 ## Other changes
 
 - Pretty-printing of Jasmin programs is more precise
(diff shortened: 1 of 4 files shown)


bgregoir:
It seem that I did not start from the good branch...

PR #627, opened 2023-10-25
Check branches of inline conditionals in the SCT checker

sarranz:
I think the example I added should fail, is it so?


vbgl (approved):
Well spotted. Generally speaking, we have to write simple test cases for the SCT checker (both positive and negative tests).

sarranz:
> Well spotted. Generally speaking, we have to write simple test cases for the SCT checker (both positive and negative tests).

Do you mean I should simplify the example I added?

PR #629, opened 2023-10-30
SCT checker: check the condition of a while after checking the first branch

sarranz:


compiler/tests/sct-checker/success/while.jazz
```diff
@@ -0,0 +1,15 @@
+fn while_first_branch(
+    #public reg u64 pub,
+    #secret reg u64 sec
+) ->
+    #public reg u64
+{
+    reg bool b;
+    b = 0 <= sec;
```
vbgl, line 8: Please use a non-trivial condition here.

sarranz: Thanks, done.

vbgl (approved):

PR #631, opened 2023-10-31
Stack zeroization on main with proofs

eponier:
This is (finally...) stack zeroization on latest main with proofs. The code is the last version (the one that makes Valgrind happy). The separation in commits should be meaningful.

I tried to consistently rename every "clear stack" to "stack zero".

The arm part is missing for the moment, so I put the PR as a draft for now.


eponier:
ARM part done. The PR is ready!

bgregoir:
Great, I will try to read it soon.

proofs/compiler/compiler_proof.v
```diff
@@ -712,38 +857,63 @@ Lemma compile_prog_to_asmP
             -> exists xm',
                 [/\ asmsem_exportcall xp fn xm xm'
                   , mem_agreement m' (asm_mem xm') (asm_rip xm') (asm_globs xp), asm_scs xm' = scs'
+                  , (cparams.(stack_zero_info) fn <> None ->
```
bgregoir, line 860: I think this is not very nice to use cparams.(stack_zero_info) here.
Because there is no guaranties on its value.
I think we should add a boolean flag in the function (source level) and refers to it. So we will have the guaranty that if the source function is annotated then the stack is cleared.

bgregoir (commented):
I will not block the PR for my remark but it should be added in the TODO list.
I think it would be nice if some other reviewer are agree that the final lemma is what we want.

sarranz:
I reviewed the statements (not proofs) of the lemmas in this PR and they look good to me.

bgregoir (approved):

PR #635, opened 2023-11-03
Change Cassgn error message in linearization

sarranz:
In response to the last few questions on Zulip. I know that Lowering is not necessarily the culprit here, so the error message might be misleading (although I believe that the passes after Lowering no longer introduce assignments). But I think the worst case scenario is what currently happens: get confused, have to ask on Zulip.

### DIFF
--- a/compiler/tests/fail/x86-64/shift.jazz
+++ b/compiler/tests/fail/x86-64/shift.jazz
@@ -1,0 +1,7 @@
+export
+fn main() -> reg u64 {
+    reg u64 x;
+    x = 1;
+    x <<= x;
+    return x;
+}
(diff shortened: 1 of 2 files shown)


vbgl:
It is true that the error message could be improved. An other thing to be done is to add some explanation in the [FAQ](https://github.com/jasmin-lang/jasmin/wiki/FAQ#how-to-debug-compilation-errors).

sarranz:
Good idea, I will add a section in the FAQ in this PR shortly.

eponier:
A better naming is good, but probably the good fix (in most of the cases) is to fail at lowering explaining why the compiler cannot lower a given instruction.

bgregoir (approved):

vbgl:
The hint “try using an intrinsic” is certainly misleading. I prefer the message that was there before.

bgregoir:
You means "assign remains" ?

vbgl:
Yes.

PR #639, opened 2023-11-07
Register-allocation: fix handling of condition registers

### LINKED ISSUE #615: Register allocation and ?{} syntax
This program
```
export
fn main(reg u64 x) -> reg u64 {
    reg bool lt;

    ?{ "<u" = lt } = #CMP(x, 10);

    while (lt) {
        x += 2;
        ?{ "<u" = lt } = #CMP(x, 10);
    }

    x = x;
    return x;
}
```
gives `compilation error: register allocation: variables { lt.238; lt.226 } remain unallocated`, but if we change one (or both) of the `?{ }` operators by `_, lt, _, _, _` then it compiles correctly.

vbgl:
Fixes #615

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,10 @@
   ([PR 623](https://github.com/jasmin-lang/jasmin/pull/623);
   fixes [#622](https://github.com/jasmin-lang/jasmin/issues/622)).
 
+- Fix handling of “condition” registers in register-allocation
+  ([PR 639](https://github.com/jasmin-lang/jasmin/pull/639);
+  fixes [#615](https://github.com/jasmin-lang/jasmin/issues/615)).
+
 ## Other changes
 
 - Pretty-printing of Jasmin programs is more precise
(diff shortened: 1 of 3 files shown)


bgregoir (commented):
Hi Vincent, 
   We can merge this because it is very simple. 
   Remark that it does not really solve the problem because if you replace <u by <=u it does not work anymore. 
   My impression is that the real problem come from the propagation of reg bool variable.
   Maybe what we should do is first do a pass that will "allocate the flag" first. Then 
   progagation of inline bool variable may work better.

bgregoir (commented):
To be clearer if you take this program:
fn main(reg u64 x) -> reg u64 {
    reg bool lt;

    

    while 
    {
       ?{ "<=u" = lt } = #CMP(x, 10); 
    } (lt) {
       x += 2;
    }

    x = x;
    return x;
}
Then it compiles perfectly.

vbgl:
I agree that this patch does not fix the issue.

PR #643, opened 2023-11-12
ARM: Add BFI and BFC

sarranz:
Please double check the EasyCrypt part, I'm really not familiar with it.


bgregoir (commented):
I think the semantic is correct, do you think there is an easy way to write it using a style closer to the arm semantic

bgregoir (approved):

PR #644, opened 2023-11-12
ARM: Add signed multiply halfwords instructions

### LINKED ISSUE #358: Add ARM Cortex M4 instructions for subword multiplication
- [x] SMULBB, SMULBT, SMULTB, SMULTT
- [x] SMLABB, SMLABT, SMLATB, SMLATT
- [x] SMULWB, SMULWT

sarranz:
Closes #358.

### DIFF
--- a/eclib/JModel_m4.ec
+++ b/eclib/JModel_m4.ec
@@ -250,6 +250,58 @@ op SMMULR (x y: W32.t) : W32.t =
   W32.of_int (IntDiv.(%/) (to_sint x * to_sint y + 2 ^ 31) (2 ^ 32)).
 op SMMULRcc x y g o = if g then SMMULR x y else o.
 
+op get_hw (is_hi: bool) (x: W32.t) : W16.t =
+  W2u16.\bits16 x (if is_hi then 1 else 0).
+
+op smul_hw (hwx hwy: bool) (x y: W32.t) : W32.t =
+  let x = to_sint (get_hw hwx x) in
+  let y = to_sint (get_hw hwy y) in
+  W32.of_int (x * y).
+op smul_hwcc hwx hwy x y g o = if g then smul_hw hwx hwy x y else o.
+
+abbrev SMULBB = smul_hw false false.
+abbrev SMULBBcc = smul_hwcc false false.
+
+abbrev SMULBT = smul_hw false true.
+abbrev SMULBTcc = smul_hwcc false true.
+
+abbrev SMULTB = smul_hw true false.
+abbrev SMULTBcc = smul_hwcc true false.
+
+abbrev SMULTT = smul_hw true true.
+abbrev SMULTTcc = smul_hwcc true true.
+
+op smla_hw (hwx hwy: bool) (x y acc: W32.t) : W32.t =
+  let x = to_sint (get_hw hwx x) in
+  let y = to_sint (get_hw hwy y) in
+  W32.of_int (x * y + to_sint acc).
+op smla_hwcc hwx hwy x y acc g o = if g then smla_hw hwx hwy x y acc else o.
+
+abbrev SMLABB = smla_hw false false.
+abbrev SMLABBcc = smla_hwcc false false.
+
+abbrev SMLABT = smla_hw false true.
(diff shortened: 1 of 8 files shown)


proofs/compiler/arm_instr_decl.v
```diff
@@ -1040,6 +1056,89 @@ Definition arm_SMMULR_instr : instr_desc_t :=
...
+    id_eq_size := refl_equal;
+    id_tin_narr := refl_equal;
+    id_tout_narr := refl_equal;
+    id_check_dest := refl_equal;
+    id_str_jas := pp_s (string_of_arm_mnemonic mn);
+    id_safe := [::];
+    id_pp_asm := pp_arm_op mn opts;
+  |}.
+
+Definition arm_smla_hw_semi (hwn hwm : bool) (wn wm acc : wreg) : exec wreg :=
```
bgregoir, line 1090: I think smla also set the sign overflow flag.

bgregoir (commented):
I think the current semantic is wrong. Also, I do not like the fact that we add 4 versions of the "same" operation. Is it possible to add only a single parametric constructor ?

proofs/compiler/arm_instr_decl.v
sarranz: It sets the saturate flag, `Q`. Back in 2021 we decided not to model it because we weren't interested in having any of the instructions that read from it. I can add it, but we still won't be able to do anything with it I think.

bgregoir: Ok, I forgot that.

sarranz:
I introduced 3 parametric mnemonics, and redefined the semantics so that it coincides more directly with the ARM documentation.
I don't know if my definition of `split_32_2x16` is ideal, same with the use of `\bits16` in EasyCrypt.
Also, unrelated, I added conditional to the test of CLZ.

eclib/JModel_m4.ec
```diff
@@ -250,6 +250,50 @@ op SMMULR (x y: W32.t) : W32.t =
   W32.of_int (IntDiv.(%/) (to_sint x * to_sint y + 2 ^ 31) (2 ^ 32)).
 op SMMULRcc x y g o = if g then SMMULR x y else o.
 
+abbrev get_hw (is_hi: bool) (x: W32.t) : W16.t =
```
bgregoir, line 253: I know we have discussed this, and maybe I confuse you.
I think smul_hw should not be an abbrev because it is the generic definition. 
I think also that you can define the generic definition
op smul_hwcc hwx hwy x y g o = 
   if g then smul_hw hwx hwy x y else o.
op smula_hw hwx hwy x y acc = smul_hw x y + acc.
op smula_hwcc hwx hwy x y acc g o = if g then smula_hw x y else o.
then all definitions become simpler.

bgregoir (commented):
Excepted that, I think it is ok.

proofs/lang/word.v
```diff
@@ -1290,6 +1290,9 @@ Definition split_vec {sz} ve (w : word sz) :=
   let wsz := (sz %/ ve + sz %% ve)%nat in
   [seq subword (i * ve)%nat ve w | i <- iota 0 wsz].
 
+Definition split_32_2x16 (w : u32) : u16 * u16 :=
+  if split_vec 16 w is [:: w0; w1 ] then (w0, w1) else (0, 0)%R.
+
```
vbgl, line 1295: I think that this function is not needed.

sarranz: What could I use instead?

vbgl: Nothing. Just inline the definition at the single call-site.

sarranz: Oh ok, good

eclib/JModel_m4.ec
sarranz: I've just changed this, I did not use `smul_hw` to defined `smla_hw` to be consistent with the reference and Coq code (addition for Z instead of words).

bgregoir (approved):

PR #646, opened 2023-11-15
Simplify reg_in_vars proofs

sarranz:


proofs/lang/var.v
```diff
@@ -390,6 +390,16 @@ Lemma Sv_diff_empty s :
   Sv.Equal (Sv.diff s Sv.empty) s.
 Proof. SvD.fsetdec. Qed.
 
+Lemma enum_in_Sv X (_ : finTypeC X) to_var s x :
+  Sv.Subset (sv_of_list to_var cenum) s ->
+  Sv.In (to_var x) s.
+Proof.
+  have : List.In x cenum.
```
sarranz, line 397: I use List.In instead of (_ \in _) because I can't convince Coq that X is an eqType without extra commands before. Is there a way of doing this?

vbgl (approved):

proofs/lang/var.v
vbgl: Do you want to solve this before merging?

vbgl: I’ve just pushed a new proof. The explicit argument to `map_f` is still a bit annoying.

What do you think?

sarranz: Looks better, thanks! I think we can merge

PR #647, opened 2023-11-15
Move expected SCT errors to individual files, add an annotation to exclude functions from test cases

sarranz:
I propose to move the test cases for SCT error messages to individual files.
With the current single file, every time I edited a test case all line numbers changed
Besides, we need to keep the list in sct_errors.ml updated

I added an annotation to exclude auxiliary functions from the tests. Before we were using the inline annotation for this, but for some cases the checker acts differently on inline functions and subroutines, so it would be good to be able to test both.


vbgl (commented):
Please amend the README to reflect the `inline` → `#[test = nosct]` change.

If the motivation for scattering the test case into many files is line numbers, there might be a better fix (e.g., that just deals with line numbers / formatting of error messages).

sarranz:
I'm open to suggestions on how to deal with just line numbers instead

vbgl:
See #648.

sarranz:
I think that's better, thanks! I'll do a separate PR for the annotation part

PR #651, opened 2023-11-21
SCT checker: Fix inference of for loops

sarranz:
This error made the checker reject the test cases I added. The fix triggered the error message I changed which was rather confusing in the original phrasing.


vbgl (approved):
Nice catch. Thanks.

PR #653, opened 2023-11-22
Exit alt ergo

vbgl:


bgregoir (commented):
What is the motivation of removing alt-ergo?

vbgl:
It is not needed.

It’s latest version is not supported by any known version of why3, so it will bring trouble when we want to update.

vbgl:
Alright. Let’s keep our CI environment bloated.

I’ve removed the second commit and only left the first one, which should not be controversial.

PR #656, opened 2023-11-29
Stack alloc: option to print in JSON format

eponier:


compiler/src/dune
```diff
@@ -12,7 +12,7 @@
  (public_name jasmin.jasmin)
  (flags -rectypes)
  (modules :standard \ arch main_compiler x86_safety uint63_63 uint63_31)
- (libraries batteries.unthreaded zarith menhirLib))
+ (libraries batteries.unthreaded zarith menhirLib yojson))
```
vbgl, line 15: Please mention this change in the changelog.

vbgl (commented):
I find it weird to output this JSON data to the error output, mixed with other debug information not meant for machine consumption.

Wouldn’t it be more convenient for the user to specify a file name?

Also, there is no test of this functionality (therefore it will break).

eponier:
Ok, I will print to a dedicated file.

What do you have in mind for the test? Testing this is valid JSON, that it prints the same information that the option we already have, something else?

vbgl:
I don’t have anything in mind beyond the following invariant of software engineering: untested features eventually break.

compiler/src/dune
eponier: Done.

eponier:
I implemented what you suggested.

And I added a test that checks that `-print-stack-alloc` and `-json-stack-alloc` stay in sync. I find the implementation dirty (the parsing of JSON is written in OCaml and I call `ocaml` to interpret it), but I don't know how to do it in a cleaner way.

bgregoir:
Are we happy with this one ?

vbgl:
I now regret to have suggested to test that feature…

eponier:
If you find the tests too ugly, I can just drop them. Another option, suggested by Benjamin last week, is to test the output like we do for the SCT checker.

eponier:
I'd like to close this one, I think the consensus is rejection. @bgregoir, you mentioned that you wanted to discuss with Tiago about this PR, did you do it?

bgregoir:
I forgot, I should do it.

PR #658, opened 2023-12-01
SCT checker: accept compiler introduced assignments to MSF

sarranz:
This example
```
inline
fn f() -> #msf reg u64 {
    reg u64 msf;
    msf = #init_msf();
    return msf;
}

export
fn protect_leak(#transient reg u64 x) -> #public reg u64 {
    reg u64 msf;
    msf = f();
    x = #protect(x, msf);
    x = x;
    return x;
}
```
doesn't work with `-checkSCTafter inline`

### DIFF
--- a/compiler/src/sct_checker_forward.ml
+++ b/compiler/src/sct_checker_forward.ml
@@ -194,8 +194,7 @@ let warn ~loc = warning SCTchecker loc
 (* used as an oracle                                         *)
 
 let is_register ~direct x =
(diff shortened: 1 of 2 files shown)


vbgl:
May you add the example to the test-suite?

sarranz:
Do you have suggestions for how to tell the test suite that this example goes with `-checkSCTafter inline`?

vbgl:
Ah my bad. This should be addressed by #608. Let’s put a note somewhere not to forget to add this test when the checker is ready.

sarranz:
I created #660

PR #663, opened 2023-12-07
CI: update nixpkgs & nix

vbgl:
This gives access to latest versions of Coq (8.18) & math-comp (2.1)


eponier:
In the end, there is no perf problem with mathcomp 2? This was mentioned at some point.

vbgl:
This is still to be investigated, but unrelated to this PR.

PR #666, opened 2023-12-13
Address a few kinds of warnings

vbgl:
- Get rid of notations deprecated in math-comp 1.17
  - Use local lemmas instead of Let that trigger warnings with Coq 8.18
  - Ensure function arguments are used in the tests for the SCT checker.


bgregoir (approved):
Thank

PR #673, opened 2023-12-20
672 it should be possible to access the element of an array at a given location i i being an int but also a word

clebreto:


bgregoir:
I don't understand why the check is not restarted after the last commit.

bgregoir:
No it is ok.

vbgl (approved):

PR #676, opened 2024-01-03
Move arm utils to a module

sarranz:
I parameterized the common operations defined in arm params so that we can use them for `Copn` and `Lopn`. I don't think we can easily abstract the proofs, but if you let me know how I would like to.


bgregoir (commented):
Nothing special to they on this PR. Maybe you can use a different name for Opn, Copn, Fopn. I mean something more specific to arm.

sarranz:
Nothing special to whom, sorry? I will change the names. I was thinking that we never `Import` both x86 and arm, so the way I was prefixing everything was cumbersome. But with a module this is probably less annoying.

sarranz:
I solved the conflicts and renamed the modules

bgregoir (approved):

PR #677, opened 2024-01-08
Arm large stack

bgregoir:
Fix the generate code for arm when stack is large.

### DIFF
--- a/compiler/src/arm_arch_full.ml
+++ b/compiler/src/arm_arch_full.ml
@@ -26,6 +26,9 @@ module Arm_core = struct
 
   let known_implicits = ["NF", "_nf_"; "ZF", "_zf_"; "CF", "_cf_"; "VF", "_vf_"]
 
+  let alloc_stack_need_extra sz =
+    not (Arm_params_common.ARMFopn.is_arith_small (Conv.cz_of_z sz))
+
 end
 
 module Arm (Lowering_params : Arm_input) : Arch_full.Core_arch = struct
--- a/proofs/compiler/arm_params.v
+++ b/proofs/compiler/arm_params.v
@@ -78,13 +78,17 @@ Section LINEARIZATION.
 
 Notation vtmpi := (mk_var_i (to_var R12)).
 
-(* TODO_ARM: This assumes 0 <= sz < 4096. *)
-Definition arm_allocate_stack_frame (rspi : var_i) (sz : Z) :=
-  ARMFopn.subi rspi rspi sz.
+Definition arm_allocate_stack_frame (rspi : var_i) (tmp: option var_i) (sz : Z) :=
+  if tmp is Some aux then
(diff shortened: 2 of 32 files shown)


compiler/src/arm_arch_full.ml
```diff
@@ -26,6 +26,9 @@ module Arm_core = struct
 
   let known_implicits = ["NF", "_nf_"; "ZF", "_zf_"; "CF", "_cf_"; "VF", "_vf_"]
 
+  let alloc_stack_need_extra sz = 
+    not (Arm_decl.is_expandable (Conv.cz_of_z sz))
```
sarranz, line 30: If this check corresponds to whether we can allocate/free stack with `SUB sp, sp, #imm` and `ADD sp, sp, #imm`, we can also allow 12-bit immediates. We should have `is_expandable || is_w12_encoding` (this corresponds to encodings T3 and T4 from the manual, e.g. A7.7.5 for ADD).

proofs/compiler/arm_params.v
```diff
@@ -79,12 +79,18 @@ Section LINEARIZATION.
 Notation vtmpi := (mk_var_i (to_var R12)).
 
 (* TODO_ARM: This assumes 0 <= sz < 4096. *)
```
sarranz, line 81: This is no longer true.

proofs/compiler/arm_params.v
```diff
@@ -79,12 +79,18 @@ Section LINEARIZATION.
...
-Definition arm_allocate_stack_frame (rspi : var_i) (sz : Z) :=
-  arm_op_subi rspi rspi sz.
+Definition arm_allocate_stack_frame (rspi : var_i) (tmp: option var_i) (sz : Z) :=
+  if tmp is Some aux then
+    arm_cmd_large_subi_tmp rspi aux sz 
+  else
+    [:: arm_op_subi rspi rspi sz].
 
 (* TODO_ARM: This assumes 0 <= sz < 4096. *)
```
sarranz, line 88: This is no longer true.

bgregoir: Sorry, what do you means here ? What is not longer true ?

compiler/src/arm_arch_full.ml
bgregoir: I will do that.

bgregoir: I now use is_arith_small

proofs/compiler/arm_params.v
bgregoir: I have remove the comment.

proofs/compiler/arm_params.v
bgregoir: Same here.

vbgl (approved):
Well done! Thanks.

PR #678, opened 2024-01-10
add x86_64 xchg instruction.

tfaoliveira:
With some help :-), I was able to create this PR.

I would like to be possible to exchange two register pointers as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## New features
 
+- Add the x86_64 instruction `XCHG`,
+    `a, b = #XCHG(a, b);` to allow swapping the contents of two operands.
+  ([PR #678](https://github.com/jasmin-lang/jasmin/pull/678)).
+
 - Support Selective Speculative Load Hardening.
   We now support operators SLH operators as in [Typing High-Speed Cryptography
   against Spectre v1](https://ia.cr/2022/1270).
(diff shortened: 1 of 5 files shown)


vbgl:
Thanks.

As hinted by the failing CI jobs, you may need to define the EasyCrypt operators matching these instructions.

The test case seems unnecessarily complex; also, it omits all the uses of the instruction involving memory operands.

Out of curiosity, what are the use cases for such an instruction?

What could be a good high-level syntax for this instruction that could work for arrays too? What about: `a, b = b, a;`?

bgregoir:
For the extraction, it is my bad, I forgot to tell this to Tiago.
For a,b = b, a;    I think this will require a bigger change in the passer. 
I was thinking to a, b = #swap(a, b);

bgregoir:
For the use case, I think the idea is the following:
you want to implement:
r = f(a,b,r);
r = f(b, a, r); 
without inline f. 
the register allocation will complain.
So what you can do is:
r = f(a,b,r);
a, b = #swap(a, b) 
r = f(a, b, r); 
The point is that swap allow you to perform the change without using extra register (good for register allocation) or memory load/store (good for protection against spectre attacks).
I was thinking on how to implement swap for arm without using extra register. 
A possibility is to use xor 
a = a ^ b;
b = a ^ b; // a^b^b = a;
a = a ^ b; // a^b^a = a;
I don't know if there is a other way of doing that.

bgregoir:
Why we do not see the test for this PR?

bgregoir:
Remark, is one of the operand is in memory my trick does not work anymore for arm.

vbgl:
I’ve just rebased & addressed the FIXME left by Benjamin in the test case. There are three more open issues now ;-)

bgregoir:
I am working on extending this with a swap intruction allowing to swap reg ptr/ stack ptr. Should we do it in that PR or in a separated one ?

vbgl (approved):

vbgl:
Separate, of course.

tfaoliveira:
thanks!

PR #683, opened 2024-01-12
Share some definitions in arm_params_common

sarranz:
Unify the smart immediate operations, and use `sem_fopn_args` more.

### DIFF
--- a/proofs/compiler/arm_params_common.v
+++ b/proofs/compiler/arm_params_common.v
@@ -67,6 +67,9 @@ Module ARMOpn (Args : OpnArgs).
       let '(hbs, lbs) := Z.div_eucl imm (wbase U16) in
       [:: movi x lbs; movt x hbs ].
 
+  Definition smart_mov x y :=
+    if v_var x == v_var y then [::] else [:: mov x y ].
+
   (* Return a command that performs an operation with an immediate argument,
      loading it into a register if needed.
      In symbols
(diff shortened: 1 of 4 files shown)


proofs/compiler/arm_params_common.v
```diff
@@ -99,36 +102,17 @@ Module ARMOpn (Args : OpnArgs).
...
-      then [:: on_imm x x imm ]
-      else li y imm ++ [:: on_reg x x y ].
-
-  (* Precondition: if [imm] is large, [x <> y]. *)
-  Definition smart_subi_tmp :=
-    gen_smart_opi_tmp sub subi is_arith_small (Some 0%Z).
+  (* Compute [R[x] := R[x] <+> imm % 2^32].
+     Precondition: if [imm] is large, [x <> tmp]. *)
+  Let gen_smart_opi_tmp on_reg on_imm is_small neutral x tmp imm :=
+    gen_smart_opi on_reg on_imm is_small neutral tmp x x (imm mod (wbase U32)).
```
bgregoir, line 108: Normally we can remove the mod operation if we remove the unnecessary 0 <= n < wbase u32 hypothesis in lemma mov_movt.
But I was lazy in doing it.

bgregoir (approved):

proofs/compiler/arm_params_common.v
sarranz: I'm on it.

vbgl: So it will be in a future PR, I guess…

PR #684, opened 2024-01-16
Compiler: emit DWARF2 line number information

vbgl:
Demo: when compiling the program below (`jasminc -o crash.s crash.jazz && cc -o crash crash.s`) and running it within `gdb`, here is the resulting output:

~~~
Program received signal SIGSEGV, Segmentation fault.
main () at crash.jazz:6
6         r = (u8)[end];
~~~

Here are the full contents of `crash.jazz`:
~~~
export
fn main(reg u64 argc argv) -> reg u8 {
  reg u64 end;
  reg u8 r;
  end = [argv + 8 * argc];
  r = (u8)[end];
  return r;
}
~~~

and the assembly listing produced by `jasminc`:
~~~
        .att_syntax
        .text
        .p2align        5
        .globl  _main
        .globl  main
_main:
main:
        .file 1 "crash.jazz"
        .loc 1 5 2
        movq    (%rsi,%rdi,8), %rax
        .loc 1 6 2
        movb    (%rax), %al
        ret
~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@
   against Spectre).
   ([PR #631](https://github.com/jasmin-lang/jasmin/pull/631)).
 
+- Interleave references to source-code positions within the assembly listings
+  when the `-g` command-line flag is given (off by default).
+  ([PR #684](https://github.com/jasmin-lang/jasmin/pull/684)).
+
 ## Bug fixes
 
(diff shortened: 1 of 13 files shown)


bgregoir (approved):

vbgl:
I’ve just added a command line flag `-g` to turn ON this new feature. It is OFF by default.

bgregoir (approved):

PR #686, opened 2024-01-16
Fix 675

### LINKED ISSUE #675: Large `param int` throws `Fatal error: exception Z.Overflow`
```
param int N = 10000000000000000000;

export
fn main() -> reg u64 {
    stack u64[N] a;
    a[0] = 0;

    reg u64 r;
    r = a[0];

    return r;
}
```
Throws an exception. It should at least be an error, but I don't see why this should fail.

eponier:
I'm not sure the fix is super clean. We don't have the precise locations at hand, so I did my best.

Fix #675


bgregoir:
Hi, I am not sure it is the good patch. 
Another solution will by to use the type Z.t instead of int in prog.ml:
i.e type ty = int gty become type ty = Z.t gty
But maybe your path is simpler, and I think for the user it change nothing

vbgl:
I’d like to try the other patch: keep the Z.t.

eponier:
Ok, but note that the two first commits are still valid.

eponier:
> I’d like to try the other patch: keep the Z.t.

Are you working on it, or should I try?

vbgl:
> Ok, but note that the two first commits are still valid.

Right. Please anyone approve & merge #688.

> Are you working on it, or should I try?

I’m not working on it. Sorry for the confusing phrasing of my earlier comment.

eponier:
I tried the other approach in https://github.com/jasmin-lang/jasmin/pull/703.

bgregoir:
The alternative approach seems more complex that expected. 
Maybe we should use this one.
Comment welcome!

eponier:
I think this one is definitely simpler. But if we aim at having only one representation ultimately, maybe the other one is better. Anyway, if we add some form of symbolic expressions in the array sizes, the type will have to change.

bgregoir:
So you think we should use the other one ?

eponier:
Actually, I think we can select this small fix for now. There is no point in making a big change if we have to do another big change soon. And I don't think the first change really helps for the second one.

bgregoir:
@vbgl do you have opinion on that ?

vbgl (approved):

PR #687, opened 2024-01-17
Add spill/unspill primitives

bgregoir:
This is mostly to simplify the life of users.
Maybe it can be useful if we want to add spillable variables.


vbgl (approved):
Future work: guess at compile-time what spilling is not necessary.

PR #690, opened 2024-01-18
Update to EasyCrypt 2024.01

vbgl:


compiler/src/toEC.ml
```diff
@@ -126,105 +126,86 @@ let leaks_lval pd = function
...
- ; "timeout"
- ; "why3"
- ; "dump"
- ; "Top"
- ; "Self"
- ; "time"
- ; "undo"
- ; "debug"
- ; "pragma"
+ [ "admit"
```
bgregoir, line 131: I think there is a way to ask easycrypt to provide the list of KW.
This does not need to be done here but we should remember it.

bgregoir (approved):

PR #691, opened 2024-01-18
Swap reg /reg ptr work on x86 and arm

bgregoir:


proofs/lang/sopn.v
```diff
@@ -24,23 +24,34 @@ Variant arg_desc :=
 | ADExplicit  of nat & option var.
 
 Record instruction_desc := mkInstruction {
+  (* Printing function to jasmin *)
   str      : unit -> string;
+  (* Printing function to EC, by default the jasmin printing is used *)
+  ec_str   : option string;
```
vbgl, line 30: This is wrong. Can’t you simply reuse the list of EC keywords in a generic way?

bgregoir: you means, something like : if it is key word add "_" at the end ?

vbgl: That would work.

bgregoir: I was liking the idea of being able to use different name for EC but this can be done latter if really needed. 
I will do that.

vbgl: I do not like said idea. What is special about EasyCrypt that should leak into the definition of the Jasmin language. When someone will want to translate Jasmin into Coq, will you add a “coq_str”? and then when you’ll need to translate to Dafny, will you add a “dafny_str”? and then when you’ll want to translate to cryptoline, will you add a “cryptoline_str”?

bgregoir: I understand that and I rather agree. Remark, that we have anyway to add the information somewhere.
For the moment, I have added a wrapper to pp_opn in to_EC. I don't like the patch, but the way printCommon.pp_opn is written make it hard to do a case on its result.

bgregoir: I have done a last commit, finally I was just stupid...

compiler/src/toEC.ml
```diff
@@ -882,6 +885,11 @@ let rec remove_for_i i =
   { i with i_desc }   
 and remove_for c = List.map remove_for_i c
 
+let pp_opn pd asmOp fmt o = 
+  let s = Conv.string_of_cstring (Sopn.string_of_sopn pd asmOp o) in
+  let s = if Ss.mem s keywords then s^"_" else s in
+  Format.fprintf fmt "%s" s
```
eponier, line 891:
```suggestion
let pp_opn pd asmOp fmt o = 
  let s = Format.asprintf "%a" (pp_opn pd asmOp) o in
  let s = if Ss.mem s keywords then s^"_" else s in
  Format.fprintf fmt "%s" s
```
to reuse `pp_opn` from PrintCommon.ml?

vbgl (approved):

vbgl:
There might be a missing check, maybe in pre-typing:

~~~
export
fn main(reg u128 x y) -> reg u128, reg u128 {
  y, x = #swap(x, y);
  return y, x;
}
~~~

> "swap128.jazz", line 3 (2-21):
  asmgen: assemble_sopn: unexpected operator swap

vbgl:
Said missing check is added in #816.

PR #692, opened 2024-01-19
allow to spill to mmx

bgregoir:
The pull request may look more complex that what it should be but it is mostly to remove stupid cycle dependency.


eponier:
Maybe you could have split the two changes to make this fact clear, but now that's done.

vbgl:
What happens when you target an architecture that does not have an “extra” register bank? Do you want to be able to write portable code (spill to mmx if the mmx set is not void, to usual registers otherwise)? or just crash (too late)?

bgregoir:
It crash, as if you declare a #mmx in arm, certainly too late, sometime generating an uncaught exception.
Do you think the fix should be done in the commit?

bgregoir:
I have added a last commit, that fix the problem of error with mmx for arm.
One other thing to discus is, there is 2 possible ways to indicate where the variable should be spill:
    1/ At the declaration of the variable (the current implemenantion)
    2/ When we spill, by having a spill_operator to mmx and on to stack.
   The second possibility is more flexible but also more cumbersome to use for the user.
   It is also a little bit more complex for the implementation.

vbgl (approved):

PR #694, opened 2024-01-22
Register allocation: extra registers have their own “bank”

vbgl:
Currently (branch `main`), compiling the new test case crashes with:

> compilation error:
register allocation: no more free register to allocate variable:
    m#6.312 (defined at "compiler/tests/fail/x86-64/mmx_conflicts.jazz", line 12 (26-27))
Conflicts with:
Fatal error: exception Not_found


compiler/src/regalloc.ml
```diff
@@ -100,10 +102,11 @@ let asm_equality_constraints ~loc pd reg_size asmOp is_move_op (int_of_var: var_
     (k': int -> int -> unit)
     (lvs: 'ty glvals) (op: 'asm sopn) (es: 'ty gexprs) : unit =
   let assert_compatible_types x y =
-    if types_cannot_conflict reg_size (L.unloc x).v_ty (L.unloc y).v_ty then
+    let x = L.unloc x and y = L.unloc y in
+    if types_cannot_conflict reg_size x.v_kind x.v_ty y.v_kind y.v_ty then
```
bgregoir, line 106: Maybe the error message should be change, since the type of reg and mmx
are the same. No idea of what we should use.

bgregoir (approved):

compiler/src/regalloc.ml
vbgl: I gave it a try in the second commit.

bgregoir (approved):

PR #701, opened 2024-01-25
B.G. and J.C.L. eval_cond also take registers as argument.

clebreto:
Adapt x86 and arm proofs to handle the additional argument. Renames specific arm Definition of eval_cond to arm_eval_cond. Use generic eval_assemble_cond definition for arm_eval_assemble_cond.


bgregoir (approved):

bgregoir:
I will be good to push this in main too.

PR #707, opened 2024-01-26
Reg ptr export

eponier:
This adds the support of reg ptr in export function arguments.

This is not polished at all for the moment. I would like to have feedback about the correctness theorem statement. I had to change it to adapt to the new feature, but the most painful task was to make the part about stack zero still pass (I had to change `match_mem` to do that).

Btw, I am not sure the current statement about stack_zero is the one we want.
For the moment, we have
```
cparams.(stack_zero_info) fn <> None -> (* if we ask to clear the stack *)
forall p, (* for every address *)
  ~ validw m p U8 -> (* not valid in the initial source memory *)
  disjoint_from_writable_params' p fn p va (get_typed_reg_values xm (asm_fd_arg xd)) -> (* not pointing to a writable param *)
  read xm'.(asm_mem) p U8 = read xm.(asm_mem) p U8 (* either we read the same thing in the final target memory and the initial target memory *)
  \/ read xm'.(asm_mem) p U8 = ok 0%R) (* or we read 0 *)
```

The correctness theorem also shows
```
forall p w, read m' p U8 = ok w -> read xm' p U8 = ok w (* if we can read in the final source memory, we can read the same data in the final target memory *)
```

But there is a discrepancy here: we do not say anything about the addresses that are valid in the source and not writable. Thus I wonder if we should not express something only relying on `read`. Something of the form:

```
cparams.(stack_zero_info) fn <> None -> (* if we ask to clear the stack *)
forall p, (* for every address *)
  disjoint_from_writable_params' p fn p va (get_typed_reg_values xm (asm_fd_arg xd)) -> (* not pointing to a writable param *)
  read xm'.(asm_mem) p U8 = read m' p U8 (* either we read the same thing in the final target memory and the final source memory (this corresponds to p being a valid address in the source *)
  \/ read xm'.(asm_mem) p U8 = read xm.(asm_mem) p U8 (* or we read the same thing in the final target memory and the initial target memory *)
  \/ read xm'.(asm_mem) p U8 = ok 0%R) (* or we read 0 *)

### DIFF
--- a/compiler/src/fInfo.ml
+++ b/compiler/src/fInfo.ml
@@ -24,16 +24,23 @@ let f_annot_empty = {
     f_user_annot          = [];
   }
 
-type subroutine_info = { returned_params : int option list }
+type arg_ret_info = { returned_params : int option list }
+(** When a non-inlined function returns a reg ptr, it has to be one of its
+    arguments. [returned_params] associates to each return value the index of
+    the corresponding argument if it is a reg ptr, and [None] otherwise. *)
 
 type call_conv =
-  | Export  (** The function should be exported to the outside word *)
-  | Subroutine of subroutine_info (** internal function that should not be inlined *)
+  | Export of arg_ret_info  (** The function should be exported to the outside word *)
+  | Subroutine of arg_ret_info (** internal function that should not be inlined *)
   | Internal  (** internal function that should be inlined *)
 
 let is_subroutine = function
   | Subroutine _ -> true
   | _            -> false
+
+let is_export = function
+  | Export _ -> true
+  | _ -> false
 
 type t =
   Location.t * f_annot * call_conv * Annotations.annotations list
--- a/compiler/src/stackAlloc.ml
+++ b/compiler/src/stackAlloc.ml
@@ -217,7 +217,7 @@ let memory_analysis pp_err ~debug up =
     let res =
       match returned_params fn with
       | Some l ->
-          Format.printf "returned_params %s@." fn.fn_name;
+          Format.eprintf "returned_params %s@." fn.fn_name;
           let l' = List.map ((=) None) l in
           if List.for_all (fun x -> x) l' then None else Some l'
       | None -> tokeep fn
@@ -470,7 +470,7 @@ let memory_analysis pp_err ~debug up =
       ao_stack_alloc  =
         fun fn ->
         try Hf.find atbl fn
-        with Not_found -> assert false
+        with Not_found -> (Format.printf "function %s is missing@." fn.fn_name; assert false)
     })
   in
 
--- a/compiler/src/stackAlloc.ml
+++ b/compiler/src/stackAlloc.ml
@@ -208,23 +208,34 @@ let memory_analysis pp_err ~debug up =
   (* FIXME: the code is duplicated between here and compiler.v, this is horrible *)
   let returned_params fn =
     let sao = get_sao fn in
-    match sao.sao_return_address with
-    | RAnone -> Some sao.sao_return
+    let _, fd = List.find (fun (_, fd) -> fd.f_name = fn) fds in
+    match fd.f_cc with
+    | Export _ -> Some sao.sao_return
     | _ -> None
   in
   let tokeep fn =
-    match returned_params fn with
-    | Some l ->
-        let l' = List.map ((=) None) l in
-        if List.for_all (fun x -> x) l' then None else Some l'
-    | None -> tokeep fn
+    let res =
+      match returned_params fn with
+      | Some l ->
+          Format.printf "returned_params %s@." fn.fn_name;
+          let l' = List.map ((=) None) l in
+          if List.for_all (fun x -> x) l' then None else Some l'
+      | None -> tokeep fn
+    in
+    Format.eprintf "function %s: @[<h>%a@]@."
+      fn.fn_name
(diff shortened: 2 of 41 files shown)

### CODE REVIEW

**Comment by eponier** [comment]
Interestingly, your example does not compile on ARM. We need an auxiliary register, I guess.

**Comment by vbgl** [comment]
Right, expansion of array copy or lowering could be enhanced.

**Comment by vbgl** [comment]
PR #735 should fix the issue with `#copy`.

File: CHANGELOG.md [hunk 1]
```diff
@@ -3,6 +3,14 @@
 
 ## New features
 
+- Export functions can have `reg ptr` as arguments and results.
+  The compiler assumes that writable `reg ptr` are disjoint from the other
+  `reg ptr` arguments and from the global data. This is the responsibility of
+  the caller to ensure that this holds.
+  For now, writable `reg ptr` must come first in the list of arguments and be
+  returned first and in the same order in the list of results.
+  ([PR #707](https://github.com/jasmin-lang/jasmin/pull/707)).
```
vbgl, line 12:
```suggestion
- Export functions can have arrays as arguments and results.
  The compiler assumes that writable `ptr` are disjoint from the other
  `ptr` arguments and from the global data. This is the responsibility of
  the caller to ensure that this holds.
  For now, writable `ptr` must come first in the list of arguments and be
  returned first and in the same order in the list of results.
  ([PR #707](https://github.com/jasmin-lang/jasmin/pull/707)).
```

I think that the wording “reg ptr” is wrong and misleading.

compiler/src/fInfo.ml
```diff
@@ -24,19 +24,22 @@ let f_annot_empty = {
     f_user_annot          = [];
   }
 
-type subroutine_info = { returned_params : int option list }
+type arg_ret_info = { returned_params : int option list }
+(** When a non-inlined function returns a reg ptr, it has to be one of its
+    arguments. [returned_params] associates to each return value the index of
+    the corresponding argument if it is a reg ptr, and [None] otherwise. *)
```
vbgl, line 30:
```suggestion
(** When a non-inlined function returns a `ptr` array, it has to be one of its
    arguments. [returned_params] associates to each return value the index of
    the corresponding argument if it is a `ptr` array, and [None] otherwise. *)
```

compiler/src/stackAlloc.ml
```diff
@@ -205,7 +205,21 @@ let memory_analysis pp_err ~debug up =
   
   (* remove unused result *)
   let tokeep = RemoveUnusedResults.analyse fds in
-  let tokeep fn = tokeep fn in
+  (* FIXME: the code is duplicated between here and compiler.v, this is horrible *)
```
vbgl, line 208: FIXME remains. This might be turned into a TODO, with a less scary comment.

compiler/src/varalloc.ml
```diff
@@ -262,26 +262,26 @@ let init_slots pd stack_pointers alias coloring fv =
...
-    | Reg (_, Pointer _) ->
+    | Reg (_, Pointer writable) ->
       let c = Alias.normalize_var alias x in
       assert (V.equal x c.in_var && c.scope = E.Slocal);
       let pi_ptr = 
         match Hv.find lalloc x with
         | RegPtr p -> p
         | _ | exception Not_found -> assert false in
-      let pi_writable = List.mem (Some i) ret in
+      let pi_writable = writable = Writable in
```
vbgl, line 280: May you chose a different variable name? The strings `writable` and `Writable` look so similar.

vbgl (commented):
Why do mutated arrays need to come first in the argument list?

CHANGELOG.md
eponier: The answer was hidden in compiler_proof.v
```
(** Export functions (entry points) have restrictions on ptr arguments and returns.
    Namely, the reg mut ptr should be the first arguments and returned at the same
    positions in the return values.
    This is not a constraint coming from the implementation, it is just meant to give
    a readable correctness theorem. *)
```

eponier:
> Why do mutated arrays need to come first in the argument list?

Sorry, Benjamin told me I answered the wrong message.

The answer was hidden in compiler_proof.v
```
(** Export functions (entry points) have restrictions on ptr arguments and returns.
    Namely, the reg mut ptr should be the first arguments and returned at the same
    positions in the return values.
    This is not a constraint coming from the implementation, it is just meant to give
    a readable correctness theorem. *)
```
I don't say it is a convincing argument, but in the current state, this would be painful to match each result to the corresponding argument. If, in the long term, we have a nice mechanism to express the matching, we can certainly remove this artificial constraint.

compiler/src/fInfo.ml
eponier: Idem.

compiler/src/varalloc.ml
eponier: Done.

compiler/src/stackAlloc.ml
eponier: Done.

vbgl:
So the motivation for this restriction is to be able to simply relate the source and assembly arguments & returned values (using `take` and `drop`). It might feel restrictive when writing Jasmin programs, but it surely makes sense when trying to use these programs after compilation from an other language.

vbgl (approved):
Yet an other long-awaited feature!

vbgl:
Future work: safety checker.

PR #710, opened 2024-01-30
fix #709 : push-pop on export function for arm

### LINKED ISSUE #709: compiler generate invalid arm code for function with large stack and callee_saved resgiter
param int N = 1024;
param int T = 10;

export fn main() -> reg u32 {
  stack u32[N] s;
  reg ptr u32[N] ps;
  reg u32 i n d;
  inline int j;

  i = 0;
  n = N;
  ps = s;
  while(i < n) {
    ps[i] = i;
    i += 1;
  }

  reg u32[T] t;
  for j = 0 to T {
    t[j] = ps[j];
  }
  i = T;
  n = (N / T) * T;
  while (i < n) {
    for j = 0 to T {
      d = ps[i];
      t[j] += d;
      i += 1;
    }
  }
  n = N;
  while (i < n) {
    d = ps[i];
    t[0] += d;
    i += 1;
  }

  for j = 1 to T {
    t[0] += t[j];
  }

  i = t[0];
  return i;

}

bgregoir:
Fix #709

### DIFF
--- a/proofs/compiler/arm_params_proof.v
+++ b/proofs/compiler/arm_params_proof.v
@@ -167,10 +167,9 @@ Proof.
   move=> sp_rsp tmp s ts sz htmp hget /=.
   rewrite /arm_allocate_stack_frame.
   case: tmp htmp => [tmp [h1 h2]| _].
-  + have [? [-> ? hget']] := [elaborate
-      ARMFopnP.smart_subi_tmp_sem_fopn_args dummy_var_info sz h1 h2 hget
+  + have [? [-> ? /get_varP [-> _ _]]] := [elaborate
+      ARMFopnP.smart_subi_tmp_sem_fopn_args dummy_var_info sz h1 h2 (to_word_get_var hget)
     ].
-    move: hget' => /get_varP [-> _ _].
     by eexists.
   rewrite /= hget /=; t_arm_op.
   eexists; split; first reflexivity.
@@ -184,10 +183,9 @@ Proof.
   move=> sp_rsp tmp s ts sz htmp hget /=.
   rewrite /arm_free_stack_frame.
   case: tmp htmp => [tmp [h1 h2]| _].
-  + have [? [-> ? hget']] := [elaborate
-      ARMFopnP.smart_addi_tmp_sem_fopn_args dummy_var_info sz h1 h2 hget
+  + have [? [-> ? /get_varP [-> _ _]]] := [elaborate
+      ARMFopnP.smart_addi_tmp_sem_fopn_args dummy_var_info sz h1 h2 (to_word_get_var hget)
     ].
-    move: hget' => /get_varP [-> _ _].
     by eexists.
   rewrite /= hget /=; t_arm_op.
   eexists; split; first reflexivity.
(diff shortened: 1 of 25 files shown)


eponier:
I'm looking at it.

proofs/compiler/arm_params_proof.v
```diff
@@ -436,6 +292,15 @@ Proof.
   exact: to_identK.
 Qed.
 
+Lemma arm_ok_lip_tmp2 :
+  exists r : reg_t, of_ident (lip_tmp2 (ap_lip arm_params)) = Some r.
+Proof.
+  exists LR.
+  rewrite /=.
+  change arm_tmp with (to_ident LR).
```
eponier, line 300:
```suggestion
  change arm_tmp2 with (to_ident LR).
```
Or better, just remove the change (and the other a few line above).

proofs/compiler/linearization.v
```diff
@@ -110,84 +113,132 @@ Record linearization_params {asm_op : Type} {asmop : asmOp asm_op} :=
...
        So ra need to be known at call cite and at the entry of the function.
        The stack frame is incremented by the caller.
 
-       The return is implemented by 
+       The return is implemented by
        Lret  (i.e POP PC in arm v7)
 
 *)
 
+Section DEFAULT.
```
eponier, line 204:
```suggestion
(* The following functions are defined here, so that they can be shared between the architectures. The proofs are shared too (see linearization_proof.v). An architecture can define its own functions when there is something more efficient to do, and rely on one of these implementations in the default case. *)
Section DEFAULT.
```

eponier (commented):
I did not have the time to read everything. I will continue tomorrow with a clear mind.

proofs/compiler/linearization.v
```diff
@@ -110,84 +113,132 @@ Record linearization_params {asm_op : Type} {asmop : asmOp asm_op} :=
...
-      var_i    (* Variable with stack pointer register. *)
-      -> Z     (* Size of the stack frame to allocate. *)
-      -> wsize (* Alignment. *)
-      -> Z     (* Offset to save old stack pointer. *)
-      -> option (seq fopn_args);
+               d := (Uptr)s
+     *)
+    lip_lmove :
+      var_i       (* Value to overwrite. *)
+      -> var_i    (* Value to assign. *)
```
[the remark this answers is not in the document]
bgregoir: Done

proofs/compiler/arm_params_proof.v
bgregoir: done, line removed.

bgregoir:
I have done the rebase on main. I hope I have not forgot to apply some correction done by JC.

eponier (approved):
Too many changes to make a precise review, but it seems ok. Just two questions:
- what is `check_ws`? can you add a comment?
- what is the difference between `killed_on_exit` and `killed_by_exit`

Just add comments about those, and I'll merge quickly.

bgregoir:
check_ws is a architecture dependent function, indicating if it is possible to load/store date of size ws.
I will add a comment.

bgregoir:
killed_on_exit : variables wrote by the function minus the saved variable (since they are restored). 
killed by_exit : variables wrote during the exit phase, the saved one that are restored and maybe some extra tmp variables.

eponier:
> check_ws is a architecture dependent function, indicating if it is possible to load/store date of size ws. I will add a comment.

Is this really needed? I have the impression that we use this function with slots that are systematically of size `Uptr`. But never mind, we can merge as is and simplify later if needed.

bgregoir:
My long term goal is to be able to reuse this code for spilling/unspilling. In that case your invariant will not be true anymore
(i.e if you spill large or small words).

proofs/compiler/linearization_proof.v
```diff
@@ -1722,6 +1917,15 @@ Section PROOF.
     | RAstack _ _ _ => Sv.add vrsp killed
     end.
 
+  (* The set of variable written by the execution of the exit code of function *)
```
bgregoir, line 1920: added comment here

proofs/compiler/linearization_proof.v
```diff
@@ -1709,11 +1902,13 @@ Section PROOF.
 
   Definition killed_on_entry (ra : return_address_location) : Sv.t :=
     match ra with
-    | RAnone => Sv.singleton var_tmp
+    | RAnone => var_tmps
     | RAreg x _ => Sv.singleton x
     | RAstack or _ _ => sv_of_option or
     end.
 
+  (* The set of variable killed/written by the execution of the function,
```
bgregoir, line 1910: added comment here

eponier:
My bad, I misread the PR. But `lip_check_ws` is still missing a comment, if I'm not mistaken.

bgregoir:
Yes, my bad for this one. I will do it.

bgregoir:
Done

eponier (approved):

PR #711, opened 2024-01-31
Small fix to a notation

eponier:
Before the fix, there was an error with mathcomp 1.19.


eponier:
Oh indeed, I did not pay attention to the fact I was not playing with a version compatible with the opam file.

vbgl:
The CI `opam` job would have caught the issue otherwise ;-)

eponier:
Indeed.

vbgl (approved):

PR #712, opened 2024-02-01
Fix arm large regptr

bgregoir:


proofs/compiler/stack_alloc_proof.v
```diff
@@ -2361,7 +2361,7 @@ Record h_stack_alloc_params (saparams : stack_alloc_params) :=
         -> (Let i' := sem_pexpr true [::] s1 e in to_pointer i') = ok i
         -> sap_mov_ofs saparams x tag vpk e ofs = Some ins
         -> write_lval true [::] x (Vword (i + wrepr Uptr ofs)) s1 = ok s2
-        -> sem_i P' w s1 ins s2;
+        -> exists2 vm2, sem_i P' w s1 ins (with_vm s2 vm2) & evm s2 =1 vm2;
```
vbgl, line 2364: The notation `… =1 …` is likely to cause trouble with recent versions of mathcomp.

bgregoir: We do not use the same version for opam and nix ?

vbgl: No. `opam` configuration is taken at random from a huge configuration space. It is highly unreliable. You should not care about the CI failure. But you should know that opam users among your co-workers might be dissatisfied.

vbgl: Also I’m not sure what the issue is. Is your code wrong? Is it a Coq bug? Did mathcomp developers messed something up?

eponier: It's not taken at random, it takes the newest version of each package allowed. And since yesterday, I pushed a commit allowing mathcomp 1.19, it now tests with this version.

eponier: Cf https://github.com/jasmin-lang/jasmin/pull/664

vbgl: > it takes the newest version of each package allowed

This is a lie.

eponier: I'm a bit optimistic since there can be conflicts between packages, and the solver can probably go crazy in some cases, but "at random" was a bit pessimistic.

vbgl (approved):

PR #715, opened 2024-02-05
Adds SUB AND MOV.

clebreto:
WIP for LOAD and STORE.

Co-authored-by: Jean-Christophe Léchenet
<jean-christophe.lechenet@inria.fr>

### DIFF
--- a/proofs/lang/sopn.v
+++ b/proofs/lang/sopn.v
@@ -54,6 +54,7 @@ Notation mk_instr_desc str tin i_in tout i_out semi safe :=
 
 Variant prim_x86_suffix :=
   | PVp of wsize
+  | PVs of signedness & wsize
   | PVv of velem & wsize
   | PVsv of signedness & velem & wsize
   | PVx of wsize & wsize
@@ -65,8 +66,7 @@ Variant prim_constructor (asm_op:Type) :=
   | PrimARM of
     (bool                 (* set_flags *)
      -> bool              (* is_conditional *)
-     -> asm_op)
-   | PrimRISCV of asm_op. 
+     -> asm_op).
 
 Class asmOp (asm_op : Type) := {
   _eqT           : eqTypeC asm_op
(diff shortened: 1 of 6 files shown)


eponier:
The last commit contains a dirty implementation of everything needed to have a running compiler for risc-v. Of course, it cannot compile anything, but at least it parses.

clebreto:
> The last commit contains a dirty implementation of everything needed to have a running compiler for risc-v. Of course, it cannot compile anything, but at least it parses.

While it is great (thanks !) to have something to get the parsing started with, this could be pushed to a different branch (maybe) ?

proofs/lang/sopn.v
```diff
@@ -54,22 +54,19 @@ Notation mk_instr_desc str tin i_in tout i_out semi safe :=
...
   | PVv of velem & wsize
   | PVsv of signedness & velem & wsize
   | PVx of wsize & wsize
   | PVvv of velem & wsize & velem & wsize
 .
 
-Variant prim_riscv_suffix :=
-  | PRsw of signedness & wsize.
-
 Variant prim_constructor (asm_op:Type) :=
```
bgregoir: Maybe this can be flatten/merged with prim_x86_suffix (with a renaming)

bgregoir (approved):

proofs/lang/sopn.v
eponier: This was not used in the end, but I rebased to avoid even introducing this type.

eponier:
I did some history cleaning, and removed the last commit, so this is just introducing 5 instructions + refactoring cond.

clebreto:
> I did some history cleaning, and removed the last commit, so this is just introducing 5 instructions + refactoring cond.

Neat !

clebreto:
@eponier How and why was the pipeline canceled ? Can we ask to run it again from here ?

eponier:
> @eponier How and why was the pipeline canceled ? Can we ask to run it again from here ?

I cancelled it, because I thought it would fail, since we are in a temporary state. But now I've removed the last commit, it should actually work. Vincent restarted the pipeline anyway, so let's see.

Issue #708, opened 2024-01-29
Register arrays in export functions produce an internal compilation error


vbgl:
Here is an example:

~~~
export
fn main(reg u64[1] x) { }
~~~

Jasmin 2023.06.2 says:

> compilation error in function main:
array expansion: reg arrays are not allowed in parameters of non inlined function

Current `main` says:

> internal compilation error in function main:
  array expansion: expansion would modify signature of an exported function (x (defined at "a.jazz", line 2 (19-20)))
Please report

Looks like a regression.

eponier:
It should be an easy one, I'm taking a look.

vbgl:
Longer-term, I think this might not fail at all. IMHO, we should see arguments as local variables. How this local variable is initialized using caller’s data is mandated by calling conventions. But the case of (non-`ptr`) arrays is tricky I guess.

eponier:
A lot of "internal" errors in this file were actually not internal, and with the help of Benjamin, I discovered some more bugs, so this is taking more time than initially planned.

eponier:
> Longer-term, I think this might not fail at all. IMHO, we should see arguments as local variables. How this local variable is initialized using caller’s data is mandated by calling conventions. But the case of (non-`ptr`) arrays is tricky I guess.

Regarding this, I agree that we should have a way to link nicely the arguments in the source and in the target. Until now, we did not change the signature of export functions, so this was easy. But with https://github.com/jasmin-lang/jasmin/pull/707 (once this is ready), this no longer holds. And if at some point we support also register arrays, this can give a theorem statement that is unreadable if we do it in a straightforward manner.

PR #720, opened 2024-02-08
Improvements to the support of reg arrays

eponier:
In particular, it fixes a bad interaction with reg ptr and turns most of the errors into non-internal ones.

Fixes https://github.com/jasmin-lang/jasmin/issues/708


bgregoir (approved):
I am not sure that remove the "." in the CHANGELOG is a good idea.
But I do not really care.
Except this point, it is very good.

proofs/compiler/array_expansion.v
```diff
@@ -1,4 +1,4 @@
-(* ** Imports and settings *)
+  (* ** Imports and settings *)
```
vbgl, line 1: Is this change intended?

vbgl:
I don’t like the last commit.

proofs/compiler/array_expansion.v
eponier: No.

eponier:
I dropped the last commit to avoid blocking the PR.

vbgl (approved):

Issue #693, opened 2024-01-19
Using mmx on ARM produces strange errors


eponier:
Compiling for ARM the program
```
export fn main () -> reg u32 {
  #mmx reg u32 x;
  reg u32 y;

  y = 0;
  x = y;
  y = x;
  return y;
}
```
produces
```
compilation error:
register allocation: no more free register to allocate variable:
    x.180 (defined at "file.jazz", line 6 (2-3))
Conflicts with:

```
Note that the list of conflicts is empty, so the message is a bit confusing.

Compiling this other program
```
export fn main () -> reg u32 {
  #mmx reg u32 x;
  reg u32 x2 y z;

  y = 0;
  x = y;
  x2 = y;
  y = x;
  z = x2;
  y += z;
  return y;
}
```
raises an exception
```
compilation error:
register allocation: no more free register to allocate variable:
    x.192 (defined at "file.jazz", line 6 (2-3))
Conflicts with:
Fatal error: exception Not_found
```

eponier:
Maybe related to https://github.com/jasmin-lang/jasmin/issues/601?

vbgl:
I guess the annotation should be ignored, with a warning.

eponier:
The second example reveals a real problem in the allocator if I'm not mistaken: the fact that registers and extra registers can conflict. The problem does not happen with large registers because they have different types from scalar register. When dealing with a class of registers, we have at our disposal the table of the variables corresponding to that class, so printing a conflict with a variable from another class fails.

eponier:
For example 1, it depends whether we want to be able to write code working both for ARM and x86. If we do not care, there is no problem rejecting programs that use mmx. We just need a better error message.

bgregoir:
I think this is fix with the last commit of #692

eponier:
No, the conflict may arise also if there are extra registers available. If we use all of them, we end up in this case too. This is just a bit more difficult to craft an example.

vbgl:
Thanks. This crash (`Fatal error: exception Not_found`) of the second example is addressed by #694.

bgregoir:
Can we close this issus or we still have a pb with the first example.

eponier:
I think we need a dedicated error message when the bank is empty. Having an error explaining that it conflicts with nothing is really confusing.

PR #721, opened 2024-02-09
Reg alloc: error message when a bank is empty

eponier:
Fixes https://github.com/jasmin-lang/jasmin/issues/693

### DIFF
--- a/compiler/src/regalloc.ml
+++ b/compiler/src/regalloc.ml
@@ -743,6 +743,17 @@ let two_phase_coloring
   let schedule =
     if !Glob_options.lazy_regalloc then lazy_scheduling variables a
     else schedule_coloring size variables cnf a in
+  (* Give a specific error message if the bank is empty: there is no way the
+     variables can be allocated. We pick one of the variables to illustrate
+     the error message. *)
+  begin match schedule, registers with
+  | i :: _, [] ->
+      let x = List.hd (Hashtbl.find variables i) in
+      hierror_reg ~loc:Lnone "unable to allocate %a: bank “%s” is empty on this architecture"
+        (Printer.pp_dvar ~debug:true) x
+        (string_of_kind (kind_of_type Arch.reg_size x.v_kind x.v_ty))
+  | _, _ -> ()
+  end;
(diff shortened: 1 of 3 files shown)


eponier:
If there is no conflict, this probably means that the bank is empty, so another fix could be to check if the set of conflicts is empty.

compiler/tests/fail/register_allocation/arm-m4/no_mmx.jazz
```diff
@@ -0,0 +1,10 @@
+export fn main () -> reg u32 {
+  reg u32 res;
+  #mmx reg u32 tmp;
```
vbgl, line 3: I disagree that this test should fail. The annotation could be ignored, with a warning.

compiler/src/regalloc.ml
```diff
@@ -743,6 +743,17 @@ let two_phase_coloring
   let schedule =
     if !Glob_options.lazy_regalloc then lazy_scheduling variables a
     else schedule_coloring size variables cnf a in
+  (* Give a specific error message if the bank is empty: there is no way the
+     variables can be allocated. We pick one of the variables to illustrate
+     the error message. *)
+  begin match schedule, registers with
```
vbgl, line 749: There is no point in computing a schedule. You may pick any `x` in the `variables` hash table (e.g. using `to_seq` and then `uncons`).

compiler/tests/fail/register_allocation/arm-m4/no_mmx.jazz
bgregoir: Hi Vincent, I am relatively agree with you. If a mmx annotation does not make sens it is certainly better to fail/ignore it in pretyping instead. Question: why do you think it should be a warning (portability ?)

vbgl: No, it’s the opposite of portability. For portability, you would like any `mmx` annotation to be silently ignored on platforms where it does not make sense. So maybe no warning would be better.

vbgl (approved):

PR #725, opened 2024-02-12
opam: ez-conf-lib is broken

vbgl:
This fixes the `opam-compiler` CI job.

### DIFF
--- a/compiler/jasmin.opam
+++ b/compiler/jasmin.opam
@@ -32,3 +32,6 @@ depends: [
   "yojson" {>= "1.6.0"}
   "ocamlfind" { build }
 ]
+conflicts: [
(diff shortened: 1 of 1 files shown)


eponier:
(The installation with opam does not work at the moment on some MacOS systems because it fails to detect ppl and mpfr, needed by Apron. The switch to ez-conf-lib was supposed to solve this.)

vbgl:
nix works well on macos. There is no point in seeking trouble and using opam.

eponier:
I agree that nix is the solution to use at the moment, but if we claim to support opam, it would be good to support it on all reasonable platforms (MacOS is not an exotic thing, it is supposed to be correctly supported by opam).

PR #730, opened 2024-02-15
x86: fix lowering of large immediate store

vbgl:


bgregoir:
Just a side not, I wonder if we should emit warning for this particular case.
   imagine this code:
     while (i < n) { 
          [p + i * 8 ] = 2u32[1,0];
          i += 1;
     } 

  many move will be introduced silently, while it is maybe better to move 2u32[1,0] one before the loop.
  A warning can be welcome.

bgregoir (approved):

PR #731, opened 2024-02-15
Pretyping: accept all calls to export functions

### LINKED ISSUE #729: Non-inline call to export function should be a simple warning
Type-checking the following program produces a type error. This is twice wrong:

  1. the program is well-typed
  2. there is a canonical fix to make thy compiler happy (namely, tag the call to `id` as `inline`)

~~~
export fn id(reg u64 x) -> reg u64 { x = x; return x; }

inline
fn whynot(reg u64 a) -> reg u64 {
  a = id(a);
  return a;
}
~~~

vbgl:
Just emit a warning if said call is not annotated as inline

Fixes #729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,11 @@
   ([PR #710](https://github.com/jasmin-lang/jasmin/pull/710);
    fixes [#709](https://github.com/jasmin-lang/jasmin/issues/709)).
 
+- Type-checking warns about calls to export functions that are not explicitly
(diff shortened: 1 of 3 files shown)


bgregoir:
Not a big fan of than, I will prefers a solution where we did not fail, if the #inline is added at call cite but doit it silently ...

bgregoir (commented):
I missing the fact that a warning is now printed. Forgot my last comment.

bgregoir (approved):

PR #733, opened 2024-02-16
Catch required files that do not exist

### LINKED ISSUE #383: Exception when importing non-existing file
We do not do any check about the files we import (whether it is a valid path, whether it is a directory, ...). Maybe we can factorize the tests with those about the main input file (cf. CLI_errors.ml).

sarranz:
Closes #383.

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -2079,7 +2079,12 @@ and tt_file arch_info env from loc fname =
   | None -> env, []
   | Some(env, fname) ->
     let ast   = Parseio.parse_program ~name:fname in
-    let ast   = BatFile.with_file_in fname ast in
+    let ast =
+      try BatFile.with_file_in fname ast
+      with Sys_error(err) ->
+        let loc = Option.map_default (fun l -> Lone l) Lnone loc in
+        hierror ~loc ~kind:"typing" "error reading file %S (%s)" fname err
+    in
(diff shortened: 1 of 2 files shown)


compiler/src/pretyping.ml
```diff
@@ -2079,7 +2079,12 @@ and tt_file arch_info env from loc fname =
   | None -> env, []
   | Some(env, fname) ->
     let ast   = Parseio.parse_program ~name:fname in
-    let ast   = BatFile.with_file_in fname ast in
+    let ast =
+      try BatFile.with_file_in fname ast
+      with Sys_error(err) ->
+        let loc = Option.map_default (fun l -> Lone l) Lnone loc in
+        hierror ~loc ~kind:"typing" "error reading file \"%s\" (%s)" fname err
```
vbgl, line 2086:
```suggestion
        hierror ~loc ~kind:"typing" "error reading file %S (%s)" fname err
```

vbgl:
Bug fixes may deserve an entry in the changelog.

compiler/src/pretyping.ml
sarranz: Thanks, done

sarranz:
> Bug fixes may deserve an entry in the changelog.

You're right, done.

vbgl (approved):

PR #736, opened 2024-02-23
Let's DOIT: Add support for DOIT-instruction list in (S)CT checkers

J08nY:
This PR will add support for ensuring that secret values only go into guaranteed constant-time instructions (Intel DOITM, ARM DIT).

Currently it does the following:
 - Adds a script for semi-automated extraction of the [Intel DOIT](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html)/[ARM DIT](https://developer.arm.com/documentation/ddi0595/2021-12/AArch64-Registers/DIT--Data-Independent-Timing?lang=en) lists. This script checks, in the Intel case, whether the opcodes on the list cover all the instruction encodings of the mnemonic on the list.
 - Adds a CLI option `-help-instructions` to the compiler that is similar to the `-help-instrinsics` one but prints all of the instructions for the given architecture that Jasmin understands along with their expanded variants. This is necessary for the automated extraction of the DIT/DOIT instructions that Jasmin supports. Example of output:
>    VPSUB:VPSUBB
>    VPSUB:VPSUBD
>    VPSUB:VPSUBQ
>    VPSUB:VPSUBW

 - Adds a CLI option `-doit` to the `jazzct` tool that **switches** the CT checker list of constant-time instructions from "*all but DIV/MOD*" to "*only those in the DOIT/DIT*" lists. This CLI option also checks that the compiler pass selected is during or after lowering, because the mode otherwise does not make sense.

### Which compiler pass to run with
It needs to be done at least after lowering, to have the instructions.

When run right after lowering, the CT checker can have false positives in case of lines like:
`if (_LT(of, cf, sf, zf)) {`
because it will consider all of the flags, even though the _LT operator only actually considers some.
If run after inline variable propagation this problem disappears, because the expression will be replaced with one that contains only the used flags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@
   returned first and in the same order in the list of results.
   ([PR #707](https://github.com/jasmin-lang/jasmin/pull/707)).
 
+- The (speculative) constant-time checker (`jazzct`) can optionally
+  (when the `-doit` flag is used) check that secrets are only used with
+  guaranteed constant time instructions (DOIT for Intel, DIT for ARM). This option
+  only makes sense when done after the lowering compiler pass, which can be ensured
+  with the new `-after` option.
+  ([PR #736](https://github.com/jasmin-lang/jasmin/pull/736)).
+
 - Add spill/unspill primitives allowing to spill/unspill reg and reg ptr
   to/from the stack without need to declare the corresponding stack variable.
   If the annotation #spill_to_mmx is used at the variable declaration the variable
--- a/compiler/CCT/fail/H_L_L.jazz
+++ b/compiler/CCT/fail/H_L_L.jazz
@@ -1,8 +1,9 @@
-fn add_H_L_L(#secret reg u64 x, #public reg u64 ya) -> #public reg u64 {
+export fn add_H_L_L(#secret reg u64 x, #public reg u64 ya) -> #public reg u64 {
 
  reg u64 t;
 
  t = x + ya;
+ t = t;
 
  return t;
 }
--- a/compiler/CCT/fail/strict.jazz
+++ b/compiler/CCT/fail/strict.jazz
@@ -1,4 +1,4 @@
-fn add(#public reg u64 p, #secret reg u64 s) -> reg u64 {
+export fn add(#public reg u64 p, #secret reg u64 s) -> reg u64 {
    p += s;
    return p;
 }
--- a/compiler/config/tests.config
+++ b/compiler/config/tests.config
@@ -14,6 +14,13 @@ kodirs = !safety/fail
 bin    = ./scripts/check-cct
 okdirs = !CCT/success
 kodirs = !CCT/fail
+exclude = !CCT/fail/doit
+
+[test-CCT-DOIT]
+bin    = ./scripts/check-cct
+args   = --doit --after=propagate
+okdirs = !CCT/success
+kodirs = !CCT/fail !CCT/fail/doit
 
 [test-SCT]
 bin    = ./scripts/check-cct
--- a/compiler/src/arch_full.ml
+++ b/compiler/src/arch_full.ml
@@ -35,7 +35,9 @@ module type Core_arch = sig
   val known_implicits : (Name.t * string) list
 
   val is_ct_asm_op : asm_op -> bool
+  val is_doit_asm_op : asm_op -> bool
   val is_ct_asm_extra : extra_op -> bool
+  val is_doit_asm_extra : extra_op -> bool
 
 end
 
@@ -71,7 +73,7 @@ module type Arch = sig
 
   val arch_info : (reg, regx, xreg, rflag, cond, asm_op, extra_op) Pretyping.arch_info
 
-  val is_ct_sopn : (reg, regx, xreg, rflag, cond, asm_op, extra_op) Arch_extra.extended_op -> bool
+  val is_ct_sopn : ?doit:bool -> (reg, regx, xreg, rflag, cond, asm_op, extra_op) Arch_extra.extended_op -> bool
 end
 
 module Arch_from_Core_arch (A : Core_arch) :
@@ -198,9 +200,9 @@ module Arch_from_Core_arch (A : Core_arch) :
       flagnames = List.map fst known_implicits;
     }
 
-  let is_ct_sopn (o : (reg, regx, xreg, rflag, cond, asm_op, extra_op) Arch_extra.extended_op) =
+  let is_ct_sopn ?(doit = false) (o : (reg, regx, xreg, rflag, cond, asm_op, extra_op) Arch_extra.extended_op) =
    match o with
-   | BaseOp (_, o) -> is_ct_asm_op o
-   | ExtOp o -> is_ct_asm_extra o
+   | BaseOp (_, o) -> (if doit then is_doit_asm_op else is_ct_asm_op) o
+   | ExtOp o -> (if doit then is_doit_asm_extra else is_ct_asm_extra) o
 
 end
--- a/compiler/src/ct_checker_forward.ml
+++ b/compiler/src/ct_checker_forward.ml
@@ -172,7 +172,7 @@ let pp_kind fmt k =
   if k = Flexible then Format.fprintf fmt "#%s " (string_of_lvl_kind k)
 
 let pp_arg fmt (x, (k, lvl)) =
-  Format.fprintf fmt "%a%a %a" pp_kind k  Lvl.pp lvl (Printer.pp_var ~debug:false) x
+  Format.fprintf fmt "%a%a %a" pp_kind k  Lvl.pp lvl (Printer.pp_var ~debug:!Glob_options.debug) x
 
 let pp_signature prog fmt (fn, { tyin ; tyout }) =
   Format.fprintf fmt "@[<h>@[%s(@[%a@]) ->@ @[%a@]@]@]@."
@@ -264,7 +264,7 @@ end = struct
       if not (Lvl.equal lvl lvlx) then
         error ~loc:(L.loc x)
              "%a has type #%s %a it cannot receive a value of type %a"
-             (Printer.pp_var ~debug:false) (L.unloc x)
+             (Printer.pp_var ~debug:!Glob_options.debug) (L.unloc x)
              Lvl.sstrict Lvl.pp lvlx
              Lvl.pp lvl
       else env
@@ -282,7 +282,7 @@ end = struct
       | Secret ->
         error ~loc
           "%a has type secret it needs to be public"
-             (Printer.pp_var ~debug:false) (L.unloc x)
+             (Printer.pp_var ~debug:!Glob_options.debug) (L.unloc x)
       | Public -> env, Public
       | Poly s ->
         let poly = Svl.filter Vl.is_poly s in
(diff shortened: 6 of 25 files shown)


J08nY:
I have just now noticed #608, which reorganizes the (S)CT checkers and introduces some of the functionality that this PR duplicates, namely the "checkCTafter" CLI option. What is the plan with that PR @bgregoir? Will it get merged soon? Should this get rebased on top of it?

bgregoir (commented):
I have not taken the time to read all of the PR. 
But two small remarks:
   - Can you propagate the change also in SCT ?
   - Clearly the changelog needs to be updated.

bgregoir:
Forgot about the first I was stupid, it is done

compiler/src/arch_full.ml
```diff
@@ -35,6 +36,7 @@ module type Core_arch = sig
   val known_implicits : (Name.t * string) list
 
   val is_ct_asm_op : asm_op -> bool
+  val is_doit_asm_op : asm_op -> bool
   val is_ct_asm_extra : extra_op -> bool
```
bgregoir, line 40: I think is_doit_asm_extra need to be added too. 
asm_extra are then compiled into sequence of asm instruction, but this is done at during the last pass of the compilation.

compiler/src/x86_arch_full.ml
```diff
@@ -47,6 +48,140 @@ module X86_core = struct
     | DIV _ | IDIV _ -> false
     | _ -> true
 
+  let is_doit_asm_op (o : asm_op) =
```
bgregoir, line 51: This function should not use | _ -> false.
All the cases should be explicite. Else in the future 
if we add new instruction the Ocaml compiler will not 
complains that we have not added a case. 
Also why this function is not automatically generated ?

bgregoir (commented):
For the option checkCTafter should we use it always when we want to do doit?

bgregoir:
Thank a lot for this PR, I am globally impressed by what you did. 
I have some small remarks but maybe I misunderstood some parts.
I think the best will be to discuss of it next week.

compiler/src/arch_full.ml
J08nY: Hmm good point, I will have to look at that more. Couldn't the solution also be to run the CT checker after that compiler pass, so there are no asm_extra ops? If the transformation of asm_extra -> sequence of asm instructions is non-trivial then this `is_doit_asm_extra` would have to essentially do the job ofthe CT checker for that sequence. I could imagine that some asm_extra turns into a sequence that contains a mix of DOIT and non-DOIT instructions, but secrets only get passed to one of the two classes or both.

Anyway, I will investigate more.

compiler/src/x86_arch_full.ml
J08nY: I will ditch the `_ -> false` and make it generate the false cases.

The function is semi-automatically generated. That is what the `extract_doit.py` script does. It goes like 95% of the way. The remaining 5% is very tricky to automate. Jasmin is not fully consistent in how it includes instructions and when it has all or only some of their variants. Also for example `MOV` or `PUSH` and `POP` do not have all of their opcodes on the DOIT list. The opcodes that manipulate (push or pop) the segment registers are non-DOIT, but those will never be used by Jasmin so it is fine. But essentially things like this need to be argued when constructing the list. Or for example the naming `MULx_lo_hi` or the fact that `CMOVcc` doesn't have the condition info in the asm_op type, etc. I could spend more time making it fully automated but it would just be full of manual special cases anyway and thus not future proof, which is like the whole point of automating it.

J08nY:
> For the option checkCTafter should we use it always when we want to do doit?

Yeah, it does not make sense to use `-doit` unless you are checking CT after `lowering`. And as discussed in the PR message, doing it after `propagate` is even better.

compiler/src/x86_arch_full.ml
J08nY: This now makes all cases explicit.

compiler/src/arch_full.ml
J08nY: Added this now. I manually investigated the cases of extra OPs and they seem OK, but this will need to be done again wen new are introduced.

J08nY:
I implemented changes based on the review:
 - There is a changelog now.
 - Added a check that gives the user an error if DOIT is run without specifying a (S)CT checking pass that is not after lowering.
 - Added `is_doit_asm_extra` that tests extra ops for being constructed out of DOIT instructions.
 - Made the DOIT lists to be explicit pattern matches and removed the `_ -> false`.

One thing that needs to be done (and I do not know how to do it) is to properly print the ARM instructions in the `-help-instructions`. The x86_64 variant of the code should be cleaned up as well. The intended format of the output is as I described in the PR description:

>  VPSUB:VPSUBB
>  VPSUB:VPSUBD
>  VPSUB:VPSUBQ
>  VPSUB:VPSUBW

compiler/CCT/fail/strict.jazz
```diff
@@ -1,4 +1,4 @@
-fn add(#public reg u64 p, #secret reg u64 s) -> reg u64 {
+export fn add(#public reg u64 p, #secret reg u64 s) -> reg u64 {
```
bgregoir, line 1: It is not related to this PR. But I don't understand why this test should fail.

compiler/config/tests.config
```diff
@@ -15,6 +15,12 @@ bin    = ./scripts/check-cct
 okdirs = !CCT/success
 kodirs = !CCT/fail
 
+[test-CCT-DOIT]
```
bgregoir, line 18: Should we add also a SCT DOIT?

CHANGELOG.md
```diff
@@ -11,6 +11,13 @@
   returned first and in the same order in the list of results.
   ([PR #707](https://github.com/jasmin-lang/jasmin/pull/707)).
 
+- The (speculative) constant-time checker (`jazzct`) can optionally
+  (when the `-doit` flag is used) check that secrets are only used with
+  guaranteed constant time instructions (DOIT for Intel, DIT for ARM). This option
+  only makes sense when done after the lowering compiler pass, which can be ensured
+  with the new `-after` option.
```
bgregoir, line 18: for -doit, did we have a default pass ?

J08nY: What do you mean by a default pass? There are two ways I guess:
 - Passing the `-doit` option also sets the compiler pass to be after inline propagation. If someone also uses the `-after` option they get an error? Or maybe they get an error only if the pass is before lowering?
 - Passing the `-doit` option does nothing to the compiler pass set, it needs to be set using the `-after` option and that will check that the set pass is after lowering. This is currently implemented (and I like it more, but I am open to suggestions).

compiler/config/tests.config
J08nY: Sure, let me add that.

compiler/CCT/fail/strict.jazz
J08nY: Running `jazzct --infer` gives me:
```
"CCT/fail/strict.jazz", line 2 (3-4):
constant type checker: p has type #strict #public it cannot receive a value of type #secret
```

CHANGELOG.md
bgregoir: I was more in favor of the first one, with
the solution "Or maybe they get an error only if the pass is before lowering?". I think the user don't need 
to know the name of the pass.

compiler/CCT/fail/H_L_L.jazz
```diff
@@ -1,8 +1,9 @@
-fn add_H_L_L(#secret reg u64 x, #public reg u64 ya) -> #public reg u64 {
+export fn add_H_L_L(#secret reg u64 x, #public reg u64 ya) -> #public reg u64 {
```
bgregoir, line 1: Can you explain why you have change all the tests.
What is the motivation of adding the export.

compiler/CCT/fail/strict.jazz
bgregoir: Thank I forgot that by default type annotated variable
are strict.

compiler/src/ct_checker_forward.ml
```diff
@@ -172,7 +172,7 @@ let pp_kind fmt k =
   if k = Flexible then Format.fprintf fmt "#%s " (string_of_lvl_kind k)
 
 let pp_arg fmt (x, (k, lvl)) =
-  Format.fprintf fmt "%a%a %a" pp_kind k  Lvl.pp lvl (Printer.pp_var ~debug:false) x
+  Format.fprintf fmt "%a%a %a" pp_kind k  Lvl.pp lvl (Printer.pp_var ~debug:!Glob_options.debug) x
```
bgregoir, line 175: This seems to be not the correct fix.
Debug may print many other infos that are usually used 
to debug the compiler. 
For doit, I am agree that it is important to print the info
(it is done after lower spill), name clashes can occur. But either replace this by just a true or a boolean 
depending on the pass.

bgregoir (commented):
I have added some extra small comment. Excepted those I think we will be ok.

compiler/CCT/fail/H_L_L.jazz
J08nY: If they are not marked `export` they are not present at the later compiler passes where DOIT checking happens :)

CHANGELOG.md
J08nY: Ok, makes sense.

compiler/CCT/fail/H_L_L.jazz
vbgl: Then those programs are not suited for testing the new functionality.

J08nY: Well, AFAIK the `export` does not change anything about them with regards to the CT checking (the ordinary tests would have caught that) so I believe these can be reused for DOIT checking. They should be reused to see that unless specific non-DOIT instructions are used DOIT checking is the same as CT checking.

This brings me to a broader point that DOIT-specific tests should be added, which I will work on.

vbgl: Did you manually inspect the error messages of the CT checker before and after your change to validate that they are still the same? Do you expect the reviewers of your PR to do this boring work?

J08nY: I would expect the tests (and some understanding of what the `export` keyword does) to ensure that. Nevertheless I just manually checked and there is no change in the ordinary CT checker output.

compiler/src/ct_checker_forward.ml
J08nY: I see, if this is set depending on the pass then the pass would need to be passed through to the checker. I think this will be quite cumbersome.

vbgl: Indeed, fixing this is non trivial and mostly unrelated to this PR. Therefore please:
1. open an issue
2. drop this commit from this PR.

vbgl: Done: #810.

vbgl:
I’ve taken to PR #811 some changes from this PR. Notable changes are that for armv7, instructions LDR, STR, and MOVT have been marked as secure. The rationale for memory accesses is that the documentation explicitly mentions that timing may only depend on the address (which is generically handled by the checker otherwise) and not on the exchanged data. For MOVT, the guess is that this instruction is not listed as a secure instruction of armv8 because it is not an instruction of armv8 in the first place.

vbgl:
I believe that all the tooling proposed in this PR, namely data extraction from manufacturer’s websites and data extraction from Jasmin, may be moved to a different repository (e.g., attached to a publication related this work).

vbgl:
Future steps might be:

  - move the DOIT data into the instruction descriptors;
  - prove that pseudo-instructions that are marked DOIT are indeed implemented using DOIT instructions only;
  - implement an independent auditing tool to validate that instructions that are marked secure in the Jasmin implementation are indeed secure according to the manufacturer’s documentation.

vbgl:
Now that PR #736 has been merged, I suggest to close this one.

PR #737, opened 2024-02-26
Fix let instead of Let in x86_instr_decl

sarranz:
We were protected from this being a problem by pretyping. `grep` says we don't have this pattern anywhere else...


vbgl (approved):
Nice catch.

PR #740, opened 2024-02-28
writeP_neq: remove redundant lemma

vbgl:
The removed lemma has too strong hypotheses (that may not hold when memory accesses are not aligned…).

### DIFF
--- a/proofs/compiler/arm_stack_zeroization_proof.v
+++ b/proofs/compiler/arm_stack_zeroization_proof.v
@@ -361,6 +361,7 @@ Proof.
     by apply hvalid.
   + move=> p hp.
     rewrite (writeP_neq hm'); first by apply hdisj.
+    apply: disjoint_range_alt.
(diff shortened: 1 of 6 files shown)


eponier:
Can you elaborate on why the hypotheses are too strong? It seems to me that the only difference between the two lemmas is related to overflows.

vbgl:
Exactly: I don’t want to prove that the pointer addition “p + i” does not overflow when it does not matter.

vbgl:
To give a bit more of context, I am currently considering a world in which unaligned memory accesses may be well defined: https://gitlab.com/vbgl/jasmin/-/commits/relax-alignment

eponier (approved):
The `writeP_neq` that remains was not masked by the one you removed (and personally, I always use the version you removed), but since it's only a matter of applying `disjoint_range_alt` and because dealing with alignment is much more interesting than arguing about this lemma, let's do it!

PR #741, opened 2024-02-28
Split large memory accesses in stack alloc

sarranz:
When stack allocation replaces `x = e` by `[rsp + off] = e`, the offset may be too large. I introduced a check at every `Copn` that splits this memory access into several instructions in ARM. I also added a warning. Note that when `x` is an array set, this can be fixed manually with `reg ptr`s, but if it's a variable there's no way of doing so (but this scenario is probably far-fetched).


eponier:
opam strikes back. You used a feature (symmetry of disequality) that is not available before Coq 8.18, and opam tests 8.16. The stupid fix is to call a lemma instead of symmetry, but maybe it is time to increase the lowest supported version to 8.18?

sarranz:
I will use the lemma, thanks!

bgregoir (dismissed):
I strongly suspect that we will have to wrote the same code for risc-V. 
So maybe we should take the time to make the code a little bit more modular.

compiler/src/compile.ml
```diff
@@ -183,10 +183,22 @@ let compile (type reg regx xreg rflag cond asm_op extra_op)
...
-     warning UseLea loc "%a" Printer.pp_warning_msg msg);
+  let warning ii w =
+    let o =
+      match w with
+      | Compiler_util.Use_lea ->
+          if not !Glob_options.lea
+          then Some (UseLea, "LEA instruction is used")
+          else None
+      | Split_memory_access ->
+          let msg =
```
bgregoir, line 194: Can you add an option like it is done with Glob_options.lea, to say if we print the warning or not ?

sarranz: Would this flag control the warning only or whether we split accesses as well?

PR #742, opened 2024-02-29
Cleaning tests for randombytes

vbgl:
There was a bit of duplication and unclear comments.


compiler/tests/success/syscall/x86-64/test_randombytes.jazz
```diff
@@ -4,24 +4,23 @@ export fn foo1() -> reg u64 {
    reg ptr u64[1] p;
    p = r;
    p = #randombytes(p);
-   r = p;
    res = r[0];
```
eponier, line 7: `p` is not used anymore, is it on purpose?

vbgl: No, it is a mistake.

eponier (approved):

PR #749, opened 2024-03-12
Regalloc: liveness annotatons

vbgl:


bgregoir (commented):
Thank a lot. I have try to use it. 
For me there is two main problems:
   - The variable names do not correspond to the one of the source program.
   -  I would like also to print variables depending on their type (and maybe the number of variables of each categories)

fdupress (dismissed):

vbgl:
Thanks Benjamin. I’ve cleaned up the history & commit message. I think this is ready.

vbgl:
ping @eponier Is there anything wrong?

eponier (approved):
Yes, something is wrong, I'm too slow to review.

PR #753, opened 2024-03-13
Attach more precise meta-data to variables introduced at compile-time

### LINKED ISSUE #718: Compiler pass "add array init" creates a lval with a dummy location
The instruction itself has a location, but not the lval. I do not know what to do here, maybe there is no good solution.

vbgl:
Fixes #718.

Partly taken from #736.

cc @J08nY


J08nY:
This is nice!

bgregoir (approved):

PR #765, opened 2024-03-21
Explicit uses of Z.of_nat & related simplifications

vbgl:
#702 was an experiment. This is a result of this experiment.


eponier (approved):
Looks good to me.

PR #768, opened 2024-03-25
EasyCrypt: depend on dune-site

vbgl:
A recently merged PR (https://github.com/EasyCrypt/easycrypt/pull/519) broke our CI!

### DIFF
--- a/scripts/easycrypt.nix
+++ b/scripts/easycrypt.nix
@@ -44,6 +44,7 @@ stdenv.mkDerivation rec {
   buildInputs = with ocamlPackages; [
     batteries
     dune-build-info
+    dune-site
     inifiles
     why3
     yojson


eponier (approved):

vbgl:
Are you suggesting to increase coupling so as to decrease synchronization issues?

eponier:
I don't understand much on these nix thing, so probably what I suggest does not make sense. I have the impression that easycrypt.nix explains how to install easycrypt, but I suppose easycrypt knows how to install itself, so there seems to be some duplication here.

eponier:
@vbgl I am still interested by an answer to my question.

vbgl:
It would make sense, sure. But I don’t see what kind of problem it is supposed to solve.

eponier:
The idea would be to avoid duplication, have this `dune-site` specified only in one file, so that if this file changes again in the future, it does not break our CI.

PR #770, opened 2024-03-27
Tests: use “reg bool” instead of “inline bool”

vbgl:


bgregoir (commented):
This is cool, did you plan to do more ?

vbgl:
No, I think I’m done. I’m not sure about https://github.com/jasmin-lang/jasmin/blob/main/compiler/tests/fail/slh/x86-64/if_wrong_update.jazz

eponier:
What is the rule of thumb with `bool` then? Always use`reg`?

vbgl:
My rule of thumb is that if it’s a value known at compile-time AND you want the compiler to resolve/eliminate computations of this value, then you flag it as `inline`.

bgregoir:
> No, I think I’m done. I’m not sure about https://github.com/jasmin-lang/jasmin/blob/main/compiler/tests/fail/slh/x86-64/if_wrong_update.jazz

I think the objective of this test is simply to check that b is used in the argument of update_msf. I don't like it because it is hard to understand that it is the expected tested feature (if msf was needed has an MSF then the false branch need to be update). 
Here since msf is returned by the export function, it will be perfectly correct to accept this program.

bgregoir:
I means here we fail, but it will possible to say if the current typing rule 
fail for #update_msf, if that case we apply the default typing rule for operator.
In that case this program will become well typed. (I don't say we should do this)

bgregoir (approved):

PR #785, opened 2024-04-17
take into account #spill/#unspill in sct checker

bgregoir:


compiler/src/sct_checker_forward.ml
```diff
@@ -1412,6 +1458,12 @@ let init_constraint fenv f =
 
   let venv = List.fold_left do_local venv (Sv.elements (locals f)) in
 
+  let do_spill venv x =
+    let sx = Env.add_spill env x in
+    do_local venv sx in
+
+  let venv = List.fold_left do_spill venv (Sv.elements (spilled f)) in
```
vbgl, line 1465: Isn’t there a `Sv.fold` operator?

vbgl:
Why don’t we just get rid of these fishy spill/unspill stuff?

compiler/src/sct_checker_forward.ml
bgregoir: yes I will change that.

bgregoir: done

vbgl:
I really feel that this spill/unspill feature is poorly designed and that there is wide room for improvement.

PR #794, opened 2024-05-11
Make SLH lowering the last pass before Stack Allocation

sarranz:
We need to put SLH lowering as late as possible so that the SCT checker can run as late as possible. In particular this means that it goes after propagate inline, which is important for the DOIT checker #736.
This introduces an annoyance, which is that propagate inline might introduce double negations (e.g. if we have `inline bool b = !x`). These get solved later by `assemble_cond` I believe. But the SLH pass does not understand them.
I now use constant propagation in the SLH pass (before we kept the last condition we saw in the environment, now we keep the simplified version of it, and then before comparing we simplify the expression in the `update_msf`).

I believe this doesn't mean we accept more programs, since constant propagation was running on all expressions anyways. Furthermore I only use the propagation to checker whether two expressions evaluate to the same value, I don't change the code.
The SCT checker didn't change.


compiler/src/glob_options.ml
```diff
@@ -141,8 +141,8 @@ let print_strings = function
   | Compiler.RemoveGlobal                -> "rmglobals", "remove globals variables"
   | Compiler.MakeRefArguments            -> "makeref"  , "add assignments before and after call to ensure that arguments and results are ref ptr"
   | Compiler.LowerInstruction            -> "lowering" , "lowering of instructions"
-  | Compiler.SLHLowering                  -> "slhlowering" , "selective load hardening lowering of instructions"
   | Compiler.PropagateInline             -> "propagate", "propagate inline variables"
+  | Compiler.SLHLowering                  -> "slhlowering" , "selective load hardening lowering of instructions"
```
bgregoir, line 145: Maybe, "lowering of selective load hardening instructions" will be better. 
Not really related to this PR. but since you touch it.

bgregoir (dismissed):

compiler/src/glob_options.ml
sarranz: Done

bgregoir (approved):

Issue #71, opened 2021-11-17
Lowering should be allowed to fail


eponier:
When lowering does not know what to do for a given instruction, it returns it without touching it. The following passes in the compiler then receive a sometimes nonsensical instruction. It would be cleaner if lowering failed in these cases.

PR #795, opened 2024-05-11
Add smart load immediate operator for ARM

sarranz:
The idea is to move everything in `ARMFopn` to extra ops that the compiler uses transparently and get solved in asm_gen.

This PR does this for loading an immediate in a register. We decided to lower assignments `x = imm` to this operator, even if it can generate two instructions.

### DIFF
--- a/proofs/compiler/arm_lowering.v
+++ b/proofs/compiler/arm_lowering.v
@@ -55,7 +55,17 @@ Section ARM_LOWERING.
 Context
   (fv : fresh_vars).
 
-Notation lowered_pexpr := (option (arm_op * seq pexpr)) (only parsing).
+(* TODO: When this pass is allowed to fail, this should be inside [cexec]. *)
+Definition low_expr : Type := option (sopn * seq pexpr).
+Definition le_skip : low_expr := None.
+Definition le_issue_sopn op es : low_expr := Some (op, es).
+Definition le_issue_extra op := le_issue_sopn (Oasm (ExtOp op)).
+Definition le_issue_aop aop := le_issue_sopn (Oasm (BaseOp (None, aop))).
+Definition le_issue_opts mn opts := le_issue_aop (ARM_op mn opts).
+Definition le_issue mn := le_issue_opts mn default_opts.
+
+Definition no_pre (ole : low_expr) : option (seq instr_r * sopn * seq pexpr) :=
+  let%opt (aop, es) := ole in Some ([::], aop, es).
 
 Definition chk_ws_reg (ws : wsize) : option unit :=
   oassert (ws == reg_size)%CMP.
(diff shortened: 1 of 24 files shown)


proofs/compiler/arm_params_proof.v
```diff
@@ -684,6 +686,227 @@ Proof.
...
+Proof. by case: res => [// | [// | [] // ?] ?] [-> ->]. Qed.
+
+Lemma uncons_wconstP ii les imm les' :
+  uncons_wconst ii les = ok (imm, les') ->
+  exists ws, les = rconst ws imm :: les'.
+Proof.
+  case: les => [// | [//|]] [] // [] // ? [] // ?? [-> ->]. by eexists.
+Qed.
+
+Lemma sem_fopn_equiv o s :
```
sarranz, line 711: I need to remove this and use the one in arm_params_core.

bgregoir (commented):
Hi Santiago, can you provide more info on the idea of this PR.
By taking a quick look to it, it does not seems to simplify the code. It seems that lowering of x = i is better.

proofs/compiler/arm_params_common_proof.v
```diff
@@ -167,7 +167,7 @@ Qed.
 (* FIXME try to remove the usage of this lemma, use sem_fopn_args version instead *)
 Lemma movi_eval_instr {lp ls ii imm xname vi} :
   let: (xi, x) := mkv xname vi in
-  (is_expandable_or_shift imm \/ is_w16_encoding imm) ->
+  (is_expandable imm \/ is_w16_encoding imm) ->
```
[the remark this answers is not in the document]
sarranz: Because it is wrong. Currently,
```
export
fn main(reg u32 x) {
    x = 0x000cb000;
    [x] = x;
}
```
fails, because lowering sees that this immediate has kind "shift" and lowers it to a `MOV`, but then `pp_arm_m4` fails with `printing: invalid immediate (831488 is too large).`.
We can't use `MOV` with immediates of kind "shift" because they set the carry flag and we are not capturing that in our semantics yet.
Printing has the correct version of the check: the immediate needs to be either less than 2^8, in which case we issue `MOV`, or less than 2^16, in which case we issue `MOVw`.
If we merge this PR then this is not an issue because the smart LI will take care of all immediates.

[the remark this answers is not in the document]
sarranz: We have the above program in the `fail` test directory, but with the `MOV` intrinsic. Currently we are making a mistake by lowering the instruction to a `MOV`, but reject it at printing time so in the end we don't print invalid instructions. I think that lowering should not issue invalid instructions however.

bgregoir (approved):

vbgl:
This looks good. Is it ready?

sarranz:
Yes

proofs/compiler/arm_lowering.v
```diff
@@ -55,7 +55,17 @@ Section ARM_LOWERING.
 Context
   (fv : fresh_vars).
 
-Notation lowered_pexpr := (option (arm_op * seq pexpr)) (only parsing).
+(* FIXME: This should be inside [cexec]. *)
```
vbgl, line 58: If this is a FIXME, then please fix it. Otherwise mark it as TODO (and clarify).

sarranz: I marked it as TODO, it's https://github.com/jasmin-lang/jasmin/issues/71

vbgl: Thanks.

vbgl (approved):
I ’ll merge as soon as CI completes.

PR #796, opened 2024-05-16
pexpr_ind: simpler proofs

vbgl:
Trying to better understand the `pexpr_ind` induction principle, I got the impression that it is generally less convenient to use than the mutual induction principle `pexprs_ind_pair`.


eponier (approved):
Nice cleaning!

eponier:
Maybe we should remove pexpr_ind and use only the mutual version...

PR #797, opened 2024-05-17
do not use SSR have in extracted code

gares:
Backward compatible, unblocks https://github.com/coq/coq/pull/18449


eponier (approved):
The right fix might be to move the proofs to a different file so that they are not extracted, but in the meantime, this does not hurt.

PR #802, opened 2024-05-22
Rework a few proofs about word.lsr

vbgl:
This will allow to easily change the implementation of `lsr`.

The `opam` CI job has been failing a lot recently (by needing too much memory) and does not know about development versions of `mathcomp-word`: I’ve thus disabled it.


eponier (approved):
There would probably be ways to fix the opam issue (e.g., have a dev version), but indeed it seems to consume too much memory, and it is not a big problem if the opam file is outdated from time to time.

PR #807, opened 2024-05-23
Move Uint63 to its own library

vbgl:
There are two implementations:
  - native, the default one, that assumes that build and host environments have same word sizes;
  - js, that only works with 32-bit host (suitable when compiling the Jasmin compiler to javascript)


eponier:
I find it strange that we have to do such a hack. I discussed with Maxime who suggested to raise a question on the Zulip, what I did [here](https://coq.zulipchat.com/#narrow/stream/240550-Dune-devs-.26-users/topic/Primitive.20integers.20and.20.28modes.20js.29). If the discussion does not converge, then I'll merge the PR, but I hope there is a cleaner solution.

vbgl:
Why do you call this a hack? I think what is in this PR is no more dirty than what is currently implemented.

bgregoir (approved):

Issue #808, opened 2024-05-24
Reject ARM intrinsics with an S when they don't support it


sarranz:
`CMPS`, `LDRS`, etc are accepted and issued by the compiler but rejected by the assembler.

PR #809, opened 2024-05-24
Reject ARM intrinsics with an S when they do not support it

sarranz:
Solves #808.


sarranz:
I added an entry in the changelog, and updated the `check-arm-m4` script. All the test cases I added were accepted by the compiler but rejected by the assembler so they wouldn't have caught this.

vbgl (dismissed):

PR #811, opened 2024-05-28
Add support for DOIT

vbgl:
This is derived from PR #736. It adds to the architecture descriptions data about which instructions have “Data Operand Independent Timing” (DOIT). The CT checkers can use this data to verify that sensitive (secret) information is only operated through such instructions.


J08nY (approved):
Looks good to me. Though long term, the automation for extraction of the lists should live somewhere, because it would get lost otherwise.

bgregoir (approved):

PR #813, opened 2024-05-29
Checksafety: minimal support of arrays in export functions

vbgl:
Do not crash; assume arrays are initialized

### DIFF
--- a/compiler/safety/fail/array-export.jazz
+++ b/compiler/safety/fail/array-export.jazz
@@ -1,0 +1,6 @@
+export
+fn main(reg ptr u8[4] a, reg u64 i) -> reg u8 {
+  reg u8 x;
+  x = a[i];
+  return x;
+}
--- a/compiler/safety/success/array-export.jazz
+++ b/compiler/safety/success/array-export.jazz
@@ -1,0 +1,7 @@
+export
+fn main(reg ptr u8[4] a, reg u64 i) -> reg u8 {
+  reg u8 x;
+  i &= 3;
+  x = a[i];
+  return x;
+}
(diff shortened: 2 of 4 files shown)


bgregoir (commented):
I don't remember, but did we have annotations for the safety checker. 
I have the impression that the assumption here is strong

vbgl:
Unfortunately no, there are no contracts for the safety checker yet.

bgregoir:
I don't know what to do with that. I am really in trouble with assuming that the array is fully initialised, mostly because the assumption will be false in many cases.
Did we do the same for memory ?

vbgl:
> the assumption will be false in many cases.

“Citation needed”.

Most programming language won’t let you create any uninitialized buffer.

bgregoir:
It is my point, most of them will not allow it. Which means that some of them will allow it.
I think some important target languages like rust and C allow it. 
You also did not answer to my question about how it is done for memory.
At the end, if there is no easy solution, I think we should merge this since it is better than nothing.

vbgl:
> languages like rust and C allow it

From the [Rustonomicon](https://doc.rust-lang.org/nomicon/unchecked-uninit.html):

> Safe Rust doesn't permit you to partially initialize an array.

I don’t care about C.

vbgl:
About the current handling of memory, here is an example:

~~~
export
fn main(reg u64 o a i) {
  reg u8 x;
  i &= 3;
  x = (u8)[a + i];
  (u8)[o] = x;
}
~~~

Running `jasminc -checksafety -safetyparam 'main>o,a,i;' deref.jazz` (with latest release version, 2023.06.3) returns:
~~~
*** No Safety Violation

* Rel:
{mem_o ≥ 0, mem_a ≥ 0, mem_a ≤ 4, mem_o ≤ 1, mem_i = 0}
mem_a ∊ [0; 4]
mem_i ∊ [0; 0]
mem_o ∊ [0; 1]
~~~

No distinction is made between input pointers (here `a`) that are expected to be initialized and output pointers (here `o`) that are only used for writing.

bgregoir (approved):
Still think it is not the best, but since we do not have a simple and better solution. Let's do that.

PR #817, opened 2024-05-29
ARM: improve error message for invalid addressing modes

### LINKED ISSUE #667: exception Failure("TODO_ARM: pp_reg_address_aux")
If we try to use an invalid addressing mode on ARM, we get an exception. We should get an error, but a proper one.

This can be observed for instance if we try to access a stack array at a position hold in a register. For instance:
```
export fn main () -> reg u32 {
  stack u32[1] s1 s2;
  reg u32 p;
  reg u32 aux;
  reg u32 res;

  aux = 0;
  s1[0] = aux;
  s2[0] = aux;
  p = 0;
  aux = s1[0];
  res = s2[(int) p]; // this is the problematic line, we need to take a reg ptr of s2 to do that
  res += aux;

  return res;
}
```

vbgl:
Fixes #667


sarranz (approved):
Thanks!

PR #818, opened 2024-05-30
Make-reference-arguments also applies to #swap

### LINKED ISSUE #815: Swapping arrays
The following program leads to an internal compilation error: “stack allocation: variable a should be a reg ptr”.

~~~
export
fn swapa(reg u64 x) -> reg u64 {
  stack u64[1] a b;
  a[0] = x;
  b[0] = x;
  a, b = #swap(a, b);
  x = a[0];
  return x;
}
~~~

I see two possible fixes: either reject this program early (e.g., during pre-typing) or strengthen make-reference-arguments to also take care of pseudo-operators.

vbgl:
Fixes #815


proofs/compiler/makeReferenceArguments.v
```diff
@@ -134,10 +134,21 @@ Definition get_syscall_sig o :=
...
   let (ii,ir) := i in
   match ir with
-  | Cassgn _ _ _ _ |  Copn _ _ _ _  => ok [::i]
+  | Copn xs tg op es =>
+      if is_swap_op op is Some ty then
+        let sig := (true, "__swap__"%string, ty) in
+        let sig := [:: sig; sig ] in
+        Let: (prologue, es) := make_prologue ii X 0 sig es in
+        Let: (xs, epilogue) := make_epilogue ii X sig xs in
+        ok (prologue ++ MkI ii (Copn xs tg (Opseudo_op (pseudo_operator.Oswap ty)) es) :: epilogue)
```
bgregoir, line 149: Can we add a code like this
 let tg := if Nat.eqb (size prologue) 2 && 
                     Nat.eqb (size epilogue) 2 then AT_none else tg in
 (* I don't know why but I have a syntax error if I use == instead of Nat.eqb *)

bgregoir: The idea is that it is clearly a case where we compute address to swap them, but where the swap instruction will be never used.

vbgl: Done.

bgregoir: Did you understand the syntax error ?

bgregoir (approved):

proofs/compiler/makeReferenceArguments.v
vbgl: No. Coq notations dark magic.

eponier: We do not include `all_ssreflect` so you need to import `eqType` and probably `ssrnat` to be able to use `==` here.

PR #824, opened 2024-06-05
Add x86 instructions RORX, SARX, SHRX, SHLX

vbgl:
Shifts & rotations that:

1. do not affect flags
2. may have different source & destination

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 
 ## New features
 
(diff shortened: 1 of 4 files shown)


bgregoir (approved):
By curiosity, did you run your fuzzing on it ?

vbgl:
No. Randomized test of the semantics of instructions currently only works for ARM. Moreover, there is no satisfying method for sampling arguments that are immediate values (such as last operand of `RORX`).

Issue #823, opened 2024-06-04
large_stack is untested


eponier:
It turns out that `make check` does not run the tests in `tests/success/arm-m4/large_stack/`.

The semantics of paths in `config/tests.config` can be surprising:
- without '!', it is the common glob interpretation
- with '!', it is recursive but globs are not interpreted

For the moment, it is not easy to express, e.g. "all the subdirs of tests called arm-m4 **and all their subfolders**".

eponier:
Ok, I have just tested, `tests/**/arm-m4/**/` should do it. Not sure about performance, thought, but probably it does not matter.

PR #820, opened 2024-05-31
Refactor runtest and improve output readability.

cassiersg:
As part of my work I needed to add new tests and ended up refactoring the `compiler/scripts/rutest` python script.

I also modified a bit the output format to be quiter about tests that pass (only kept a status line at the bottom with current progress and last test run).

Example output:
```
scripts/runtest --jobs="2" config/tests.config --report=report.log safety CCT x86-64-ATT x86-64-Intel x86-64-print x86-64-nolea arm-m4 arm-m4-print x86-64-stack-zero-loop x86-64-stack-zero-unrolled arm-m4-stack-zero-loop arm-m4-stack-zero-unrolled
Failed test ./scripts/check-x86-64 -ATT tests/success/x86-64/aes.jazz
--------------- Stdout ---------------

--------------- Stderr ---------------
+ dirname ./scripts/check-x86-64
+ ./scripts/../jasminc -g -o jasminAkFKxD/jasmin.s -ATT tests/success/x86-64/aes.jazz
"tests/success/x86-64/aes.jazz", line 50 (17-26)
warning: unused variable next_rkey.191
+ echo -ATT tests/success/x86-64/aes.jazz
+ grep -q fail
+ cc -c -o jasminAkFKxD/jasmin.o jasminAkFKxD/jasmin.s
jasminAkFKxD/jasmin.s: Assembler messages:
jasminAkFKxD/jasmin.s:130: Error: operand type mismatch for `vaeskeygenassist'
+ rm -r jasminAkFKxD


Failed test ./scripts/check-x86-64 -intel tests/success/x86-64/aes.jazz
--------------- Stdout ---------------

--------------- Stderr ---------------
+ dirname ./scripts/check-x86-64
+ ./scripts/../jasminc -g -o jasminmxWlZj/jasmin.s -intel tests/success/x86-64/aes.jazz
"tests/success/x86-64/aes.jazz", line 50 (17-26)
warning: unused variable next_rkey.191
+ echo -intel tests/success/x86-64/aes.jazz
+ grep -q fail
+ cc -c -o jasminmxWlZj/jasmin.o jasminmxWlZj/jasmin.s
jasminmxWlZj/jasmin.s: Assembler messages:
jasminmxWlZj/jasmin.s:130: Error: operand type mismatch for `vaeskeygenassist'
+ rm -r jasminmxWlZj

Tests:     453/2200 | Failed:    2 | [SUCCESS] ./scripts/parse-print-parse tests/success/x86-64/test_mem.jazz
```
[...]

PR #825, opened 2024-06-05
Do not ignore test subdirectories

eponier:
Make sure all the subdirectories are tested. Also renames some template files into *.jinc. Before the rename, they were tested in isolation while they do not make sense alone.

Fixes https://github.com/jasmin-lang/jasmin/issues/823


eponier:
The PR is a draft since the newly tested files are broken, so this PR depends on the fix of the tests.

eponier:
New implementation on top of https://github.com/jasmin-lang/jasmin/pull/820. `!` was removed. We are now recursive by default (i.e. when we give a path, we test the files in all the subdirs of this path).

eponier:
We now run `jasminc` on `examples/gimli/x86-64/ecproof/gimli_ref{,1}.jazz`, is this what we want?

vbgl (approved):

PR #831, opened 2024-06-10
x86: add VAES instructions

### LINKED ISSUE #630: VAES support
On x86_64, the `[V]AES{DEC,DECLAST,ENC,ENCLAST}` intrinsics only support `u128`operands. Same for `PVCLMULQDQ`.

This is sad, because many Intel and Intel-compatible CPUs support `2u128` operands with the same throughput/latency.

Would it be possible to add support for VAES in a future release?

Thank you!

vbgl:
The AES-NI AVX instructions are extended to 256-bit operands.

Fixes #630.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 
 ## New features
 
+- The instructions of the VAES extension are available through a size suffix
+  (e.g., `VAESENC_256`)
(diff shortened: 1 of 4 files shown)


bgregoir (approved):
At some point it will become urgent to flag the instruction with the name of the extension.

jedisct1:
Thank you!

PR #832, opened 2024-06-11
Add a conflicts field to instruction descriptors

eponier:
and use it in Oarm_add_large_imm.

This allows to go back to a situation where RSP is not written when entering register allocation.


proofs/lang/sopn.v
```diff
@@ -29,6 +29,7 @@ Record instruction_desc := mkInstruction {
   i_in     : seq arg_desc;
   tout     : list stype;
   i_out    : seq arg_desc;
+  conflicts: seq (arg_desc * arg_desc);
```
vbgl, line 32: It is surprising to have `arg_desc` here. Moreover there is non-trivial code in regalloc to translate these arg-desc into positions in one of the argument or ℓ-value lists. Would it be possible to directly put the relevant index here? (e.g., moving the `arg_position` data-type from regalloc here.)

eponier: Let's try!

eponier:
I made a try. I used `nat` in Coq, but I didn't want to use `nat` everywhere in OCaml, so I duplicated type `arg_position`.

The code is definitely simpler, but I found that the previous description was easier to read. But it's not a strong opinion.

vbgl:
I don’t buy the argument about `int` and `nat`.

eponier:
What is your idea? Use `nat` everywhere? I can also use `Uint63.t` instead, but maybe not worth.

vbgl:
Yes. `nat` is the right data-structure here. Let me give it a try.

vbgl (dismissed):

vbgl:
Jean-Christophe, are you OK with my changes?

compiler/src/regalloc.ml
```diff
@@ -92,10 +90,12 @@ let find_equality_constraints (id: instruction_desc) : arg_position list list =
...
   match ap with
-  | APout n -> (List.nth outs n |> function Lvar v -> Some v | _ -> None)
+  | APout n ->
+     Option.bind (Oseq.onth outs n)
+       (function Lvar v -> Some v | _ -> None)
   | APin n ->
-     (List.nth ins n |>
-        function
+     Option.bind (Oseq.onth ins n)
+       (function
```
eponier, line 98: Why do you use options there? If `outs` or `ins` are not long enough, I think this means there is an error in the instruction description, which we would like to report rather than ignore?

compiler/src/regalloc.ml
```diff
@@ -68,19 +68,19 @@ let types_cannot_conflict reg_size kx x ky y : bool =
   | _, _ -> true
 
 let find_equality_constraints (id: instruction_desc) : arg_position list list =
-  let tbl : (Datatypes.nat, arg_position list) Hashtbl.t = Hashtbl.create 17 in
+  let tbl : (int, arg_position list) Hashtbl.t = Hashtbl.create 17 in
```
eponier, line 71: Out of curiosity, what was the problem with `Datatypes.nat` there?

vbgl: This change does not exist.

compiler/src/regalloc.ml
vbgl: You’re right. Let me fix that.

vbgl: Done. The error message is not very informative.

compiler/src/regalloc.ml
eponier: I know this is not in the final version, but at some point you turn the `int` into `Datatypes.nat` and then back. I was just curious to know why.

vbgl: This is called a mistake.

eponier: I tought it was a change made on purpose that turned out not to work, if it was just a mistake from the beginning, my question does not make sense indeed.

vbgl: Let me tell you the story of a function whose parameters are called `n` and `p`, in this order, and then a few lines below said function is invoked with arguments `p` and `n`, in this order. Then the hasty reader went lost wondering whether they changed the type of `n` or the type of `p`. (Correct answer is: none, and both).

compiler/src/regalloc.ml
eponier: In general, our internal errors are not very informative, that's ok.

eponier:
Ok to me.

PR #833, opened 2024-06-11
Raise lower bound on batteries in opam files

eponier:
I submitted https://github.com/ocaml/opam-repository/pull/26051 to add the latest Jasmin release to opam, and it revealed that we are not compatible with batteries < 3.5. I think we have been incompatible for a long time, so I don't know why opam's CI that is really good at testing lower bounds did not catch this problem for previous releases.


vbgl (approved):
Thanks.

PR #835, opened 2024-06-12
Use Coq 8.19

vbgl:
The first commit is unrelated but might not deserve its own PR.


eponier (approved):
I trust you on the docker thing, I couldn't find what the change implies.

PR #844, opened 2024-06-21
Regalloc: improve liveness analysis

### LINKED ISSUE #634: Register allocation doesn't track conflicts between lvals
This program
```
export
fn main() {
    reg u64 x y;
    x = 0;
    y = 0;

    _, [y] = #MULX_lo_hi(x, y);
}
```
gives `"ble.jazz", line 7 (4-31): internal compilation error in function main: Variable allocation: y (defined at "ble.jazz", line 5 (4-5)) and RAX : variable already set`.

Because register allocation chooses the first available register for `_` (`RAX` in this case), without realizing it's read in the second destination (`[RAX]`).
```
/* After register allocation */
export
fn main () -> () {
  reg u64 RDX;
  reg u64 RAX;
  
  RDX = #MOV_64(((64u) 0));
  RAX = #MOV_64(((64u) 0));
  (RAX, (u64)[RAX + ((64u) 0)]) = #MULX_lo_hi_64(RDX, RAX);
  return ();
}
```

I could not find where this is dealt with in Register allocation/liveness analysis.
I know this program won't have an x86 compilation because `MULX` does not allow memory destinations, but it's the simplest example I could think of. In OTBN and ARMv7 there are instructions that do this.

vbgl:
During register-allocation (i.e., when performing “weak” liveness analysis), values that are used in the left-hand side of an assignment (assgn, opn, call, syscall, etc.) are considered live after the instruction.

Fixes #634

### DIFF
--- a/compiler/src/liveness.ml
+++ b/compiler/src/liveness.ml
@@ -10,16 +10,19 @@ let dep_lv s_o x =
 let dep_lvs s_o xs =
   List.fold_left dep_lv s_o xs
 
-let writev_lval s = function
+(** Adds to [s] variables that are used as destination or to compute a
+destination (array index, memory offset) *)
+let weak_dep_lv s = function
   | Lnone _     -> s
   | Lvar x      -> Sv.add (L.unloc x) s
-  | Lmem _      -> s
-  | Laset(_, _, _, x, _)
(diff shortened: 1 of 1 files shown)


vbgl:
Just added a comment and renamed a function.

bgregoir (approved):

bgregoir:
much better function name :-)

PR #846, opened 2024-06-24
[WIP] Refactor extraction to build an easycrypt AST, then pretty-print it.

cassiersg:
This allows to simplify the code and make it easier to extend.

The main commit is the last one (`31b149b7`) and does not change the extracted `.ec` for all test cases (besides the amount of whitespace).

The first 4 commits prepare for testing the changes, then `9a26005f` fixes a bug in the extraction, `d0b25cdb` removes the unused safety extraction, and `4957f20f` changes the parentheses and whitespace (the code from that commit is completely over-written in `31b149b7`, but that allows to check that `31b149b7` is a no-op).

I left the option for setting the pretty-printing width (`205e83c2`)  and the comparison test scripts (`0db786d3`), but they can be removed if deemed not useful.

cc @bgregoir


vbgl:
The fixes to the testing infrastructure should be moved to a different PR (and merged quickly).

The removal of extraction for safety deserves a changelog entry and misses some cleaning in the `eclib/` directory. It cannot be merged before the next release (2024.06.0).

cassiersg:
@lyonel2017 Thanks for spotting this, it should be removed.

The back-story is that the current change tries very hard to replicate the output of the previous extraction in order to make sure to not break the logic somewhere, since I have not seen a way to test the extraction. The newline is an artifact of this, but should indeed be removed (along with a few idiosyncrasies in the pretty-printing).

lyonel2017:
Some additional remarks.

- For the constructor for lambda ([`Efun1`](https://github.com/cassiersg/jasmin/blob/2dfcc41fac2ea5f173ca6e520bfc4aaab1b56645/compiler/src/toEC.ml#L480)) in `ec_expr`, to avoid currying, I would propose to replacing the string (for the bound variable) with a list of strings.

- The constructor for lambda (`Efun1`) in `ec_expr` takes a string for the bound variable which is for me unnatural; I would expect a variable. The reason for that, I think, is the [`ec_ident`](https://github.com/cassiersg/jasmin/blob/2dfcc41fac2ea5f173ca6e520bfc4aaab1b56645/compiler/src/toEC.ml#L473) type which is a list of strings. Maybe it would be interesting to have a more sophisticated types for identifier of variables and record, tuples, and modules access.

- I have the impression that using a functor to manage the Normal and ConstantTime models would result in cleaner code ([example](https://github.com/cassiersg/jasmin/blob/2dfcc41fac2ea5f173ca6e520bfc4aaab1b56645/compiler/src/toEC.ml#L1020)) and simplify the addition of new models in the future.

lyonel2017:
- For `Efun1`, yes exactly.
- For the type of identifiers and the `ec_expr` type, perhaps something like this (to be discussed):

``` ocaml

type ec_type = ...
type ec_op = ...

type ec_var = {
   var_name = string
   var_type = ec_type
}

type ec_expr =
   ...
   | Efun   of (ec_var list) * ec_expr
   | Eproj  of ec_expr * int              (* projection of tuple *)
   | Evar   of ec_var
   | Eop    of ec_op
   | Eapp   of ec_expr * ec_expr list
   ...
```

Record access uses a projection operator in EC, if I've understood correctly (maybe wrong). The case is therefore taken into account by `Eapp`. The printer for `ec_expr` needs to be adapted, and some information needs to be added to the `ec_op` type so that we print the right thing.

- For the functor part, I had something a little different in mind. For example, for `toec_intr`:

``` ocaml
module type Model = sig
....
val assign : env -> lval -> expr -> instr
...
end.

module Normal : Model = struct
....
let assign env l e =
            let e = toec_cast env (ty_lval lv) e in
            [toec_lval1 env lv e]
....
end.

module ConstantTime : Model = struct
....
let assign env l e =
            let tys = [ty_expr e] in
            (ec_leaks_e env e) @
            ec_call env [lv] tys tys (toec_expr env e)
....
end.

Module EC (M: Model) = struct
.....
and toec_instr asmOp env i =
 match i.i_desc with
    | Cassgn (lv, _, _, (Parr_init _ as e)) ->
        (ec_leaks_e env e) @
        [toec_lval1 env lv (ec_ident "witness")]
    | Cassgn (lv, _, _, e) -> M.assign env lv e
    | .......
....
end
````

cassiersg:
@vbgl This has now been rebased following last week's discussion (first step of the plan).

I think that it's best to merge this PR in its current state, and defer further cleanups/improvements to future PRs.

For future reference, possible improvements are:
- more unification of the `Normal` and `ConstantTime` code paths (that would slightly change the extracted easycrypt code)
- making the easycrypt AST a bit more general (see https://github.com/jasmin-lang/jasmin/pull/846#issuecomment-2298467326)
- investigate the move to functors for handling the various extraction modes (see https://github.com/jasmin-lang/jasmin/pull/846#issuecomment-2298467326). Doing this without too much code duplication would likely require de above clean-ups to be performed first.

vbgl (approved):

PR #851, opened 2024-06-26
Improve test for easycrypt extraction

cassiersg:
Split out from #846

### DIFF
--- a/compiler/scripts/runtest
+++ b/compiler/scripts/runtest
@@ -264,7 +264,7 @@ def _dump_report(results, out):
         grouped.setdefault(result.config.group, []).append(result)
 
     hostname = socket.gethostname()
-    timestamp = datetime.datetime.utcnow().isoformat()
+    timestamp = datetime.datetime.now(datetime.UTC).isoformat()
     for gname, group in grouped.items():
         ko = [x for x in group if not x.success]
         node = cl.OrderedDict()
--- a/compiler/scripts/runtest
+++ b/compiler/scripts/runtest
@@ -359,7 +359,7 @@ def _main():
                     file=sys.stderr,
                 )
             ANSITerm.progress_write(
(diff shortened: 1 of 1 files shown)


compiler/scripts/extract-and-check
```diff
@@ -2,24 +2,28 @@
...
-trap "rm -r ${DIR}" EXIT
+# trap "rm -r ${DIR}" EXIT
+
+ECLIB=$(dirname $0)/../../eclib
 
 set -x
 
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$@"
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -CT -oec ${CT} "$@"
+echo $(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$2"
```
vbgl, line 23: '$2' seems wrong here.

CI says: `Error: File  not found.`

cassiersg: The issue was coming from the `shift` above, from an incorrect merge.

cassiersg:
@vbgl The issue should be fixed.

vbgl:
Well, there seems to be exactly the same issue with the `extract-golden` file.

vbgl:
> The fixes to the testing infrastructure should be moved to a different PR (and merged quickly).

This applies to this PR too…

The first two commits are OK. The third one is a new feature. The last one is a draft for a new feature mixed up with a fix for an unconfirmed issue (please file a bug report if you think there is a real problem).

cassiersg:
I left only the two first commits - the rest will follow later when I have time to refactor the commits.

BTW, is there any test to ensure that the extraction is correct, beyond being accepted by easycrypt ?

vbgl (approved):

PR #873, opened 2024-08-01
Cleaning

eponier:
Two independent patches.


vbgl:
Thanks. I’ve just added my unrelated stuff ;-)

vbgl:
CI suggests future work (not to be addressed here):

> Error: Cannot extract nontrivial sort polymorphism

vbgl (approved):
Before merging, it would be nice if someone could give a look to the third commit.

eponier:
> Before merging, it would be nice if someone could give a look to the third commit.
Trivial removal of unused lemmas, so ok to me. I was sure that they were used, but apparently not anymore.

PR #880, opened 2024-08-21
Add support for slices in array #copy

### LINKED ISSUE #842: Assertion failed in `insert_copy_and_fix_length`
This program
```
export fn fail() -> reg u64
{
  stack u64[1] a b;
  a[0] = 1;
  b[0:1] = #copy(a);
  reg u64 r;
  r = b[0];
  return r;
}
```
gives
```
Fatal error: exception File "src/insert_copy_and_fix_length.ml", line 75, characters 86-92: Assertion failed
```

vbgl:
Fixes #842.

### DIFF
--- a/compiler/tests/fail/x86-64/unaligned_slice_copy.jazz
+++ b/compiler/tests/fail/x86-64/unaligned_slice_copy.jazz
@@ -1,0 +1,10 @@
+// Cannot copy by chunks of 4 bytes starting at offset 3 bytes
+export fn main() -> reg u32 {
+   stack u8[8] s;
+   s[u64 0] = 0;
+   stack u32[1] d;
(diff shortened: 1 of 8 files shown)


compiler/tests/fail/x86-64/unaligned_slice_copy.jazz
```diff
@@ -0,0 +1,10 @@
+// Cannot copy by chunks of 4 bytes starting at offset 3 bytes
```
bgregoir, line 1: Why this should fail ? We can now have unaligned access.

vbgl: Currently, the `copy` operator only supports aligned accesses.

bgregoir (approved):

PR #883, opened 2024-08-28
CI: fix libjade jobs

vbgl:
Libjade is undergoing heavy refactoring. This is a hacky fix to keep things working meanwhile.


eponier (approved):

eponier (changes requested):
Actually it seems that the artifact is not produced anymore (in both jobs libjade-compile-to-asm and libjade-extract-to-ec). Do we care?

eponier (approved):
Thanks

PR #884, opened 2024-08-29
eclib: remove redundant dependencies in JModel_x86

vbgl:


fdupress (approved):
Checked with https://github.com/EasyCrypt/easycrypt/pull/605.

PR #886, opened 2024-09-03
Modifying jasmin2tex integer representation to fit source integer representation better

MrDaiki:
# Issue

jasmin2tex program always convert integer from source file to base 10. In certain cases, it makes readability of the program harder. For example : 

```jazz
tab[i] = 0xab;
```
is converted to :

```tex
\jasminindent{2}tab[i] = 172;\\
```

# Solution 

We implement a new Syntax node `pbinteger` that store the raw integer string coming from source file : 
```ml
type pbinteger = {zvalue:Z.t; raw:string}
```

At pretyping, we convert this type back to `Z.t`, which mean that it does not affect the behavior of the compiler for all next compilation steps. Since `latex_printer` is executed before pre-typing, we can use the raw string for generating latex file. The new output is now : 

```tex
\jasminindent{2}tab[i] = 0xab;\\
```

# Changelog 
* Add new Syntax Node `pbasedinteger` which dissapear at pretyping
* Changing parser and lexer accordingly
* Changing latex_printer integer handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,10 @@
 
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
   ([PR #869](https://github.com/jasmin-lang/jasmin/pull/869)).
+
 - The checker for S-CT accepts copies of outdated MSF
   ([PR #885](https://github.com/jasmin-lang/jasmin/pull/885)).
+  
 - Preserve formatting of integer literals in the lexer and when pretty-printing to LATEX
   ([PR #886](https://github.com/jasmin-lang/jasmin/pull/886)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
   ([PR #869](https://github.com/jasmin-lang/jasmin/pull/869)).
 - Modification of jasmin2tex to support source file integer representation.
+- Adding new integer syntax
 
 # Jasmin 2024.07.0 — Sophia-Antipolis, 2024-07-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
 
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
   ([PR #869](https://github.com/jasmin-lang/jasmin/pull/869)).
-- Modification of jasmin2tex to support source file integer representation.
+
+- Preserve formatting of integer literals in the lexer and when pretty-printing to LATEX
+  ([PR #886](https://github.com/jasmin-lang/jasmin/pull/886)).
 
 # Jasmin 2024.07.0 — Sophia-Antipolis, 2024-07-09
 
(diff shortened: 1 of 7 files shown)


compiler/src/lexer.mll
```diff
@@ -135,10 +135,10 @@ rule main = parse
 
   (* Why this is needed *)
   | ((*'-'?*) digit+) as s   
-      { INT (Z.of_string s) } 
+      { INT {zvalue=Z.of_string s;raw=s}} 
 
   | ('0' ['x' 'X'] hexdigit+) as s
-      { INT (Z.of_string s) }
+      { INT {zvalue=Z.of_string s;raw=s} }
```
vbgl, line 142: Why do you compute the Z value here? Isn’t it simpler to only keep the string here and compute the value in pre-typing?

MrDaiki: I wasn't very used to the process of the compiler so I wanted the modification to affect the smallest possible part of the code. If you think it is preferable that the Z type building is done in pre-typing, I can change it (it will indeed simplify code readability)

compiler/src/pretyping.ml
```diff
@@ -2144,14 +2144,15 @@ let tt_global pd (env : 'asm Env.env) _loc (gd: S.pglobal) : 'asm Env.env =
...
+
 (* -------------------------------------------------------------------- *)
 let rec tt_item arch_info (env : 'asm Env.env) pt : 'asm Env.env =
   match L.unloc pt with
   | S.PParam  pp -> tt_param  arch_info.pd env (L.loc pt) pp
   | S.PFundef pf -> tt_fundef arch_info env (L.loc pt) pf
   | S.PGlobal pg -> tt_global arch_info.pd env (L.loc pt) pg
   | S.Pexec   pf ->
-    Env.Exec.push (L.loc pt) (fst (tt_fun env pf.pex_name)).P.f_name pf.pex_mem env
+    Env.Exec.push (L.loc pt) (fst (tt_fun env pf.pex_name)).P.f_name (List.map (fun ((x,y):string*string) -> (Z.of_string x),(Z.of_string y)) pf.pex_mem) env
```
vbgl, line 2155:
```suggestion
    Env.Exec.push (L.loc pt) (fst (tt_fun env pf.pex_name)).P.f_name (List.map (fun (x, y) -> Z.of_string x, Z.of_string y) pf.pex_mem) env
```

vbgl (commented):
Considering your future plans, it might be worth introducing an abstract type (say `int_notation`) right now.

compiler/src/pretyping.ml
MrDaiki: You're totally right, type annotation was mainly for debugging purposes, I can remove it now

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
   ([PR #869](https://github.com/jasmin-lang/jasmin/pull/869)).
+- Modification of jasmin2tex to support source file integer representation.
```
vbgl, line 18:
```suggestion

- Preserve formatting of integer literals in the lexer and when pretty-printing to LATEX
  ([PR #886](https://github.com/jasmin-lang/jasmin/pull/886)).
```

vbgl (dismissed):
Thanks. I’ll merge as soon as CI completes.

vbgl (approved):

PR #889, opened 2024-09-04
Minor cleaning

vbgl:


proofs/lang/memory_model.v
```diff
@@ -580,7 +580,7 @@ Qed.
 (** Pointer arithmetic *)
 
 #[ global ]
-Instance Pointer : pointer_op pointer.
+Instance PointerW : pointer_op pointer.
```
bgregoir, line 583: No problem with this change, but what is the motivation?

bgregoir (approved):

proofs/lang/memory_model.v
vbgl: Motivations are spelled out in the commit message.

bgregoir: Thank, I forgot to click on the "..." button.

PR #892, opened 2024-09-05
varalloc: system calls do modify the stack pointer

### LINKED ISSUE #870: Cannot use #randombytes if there is no stack
The following program is rejected by the Jasmin compiler:

~~~
export
fn rand(reg ptr u8[32] io) -> reg ptr u8[32] {
  io = #randombytes(io);
  return io;
}
~~~

> "tmp/rand.jazz", line 1 (0) to line 5 (1):
compilation error in function rand:
linearization: bad save-stack

Reported by Aaron (@Rixxc).

vbgl:
Fixes #870.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
   ([PR #888](https://github.com/jasmin-lang/jasmin/pull/888));
   fixes [#887](https://github.com/jasmin-lang/jasmin/issues/887)).
 
+- Fix compilation of functions with system calls but not making other use of
+  the stack
+  ([PR #892](https://github.com/jasmin-lang/jasmin/pull/892));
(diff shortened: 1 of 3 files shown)


eponier (commented):
I understand why you add a new condition, but not why you remove `ra_on_stack`. Can you explain why it is not there anymore?

vbgl:
You’re right. This change is unrelated, sorry. I’ve removed the condition because it is redundant with the previous one: `ra_on_stack` implies the stack size being non-zero.

eponier (approved):
Indeed, thanks!

eponier:
(I have no problem shipping the cleaning in this PR, it was just not clear to me that it was correct or on purpose)

PR #896, opened 2024-09-06
pretyping: small cleaning

vbgl:
Remove a dead function, a legacy comment, and the now obsolete `tt_ws` function.


bgregoir (approved):
Second commit is very nice

PR #898, opened 2024-09-11
Tests: fix output as tests progress

vbgl:
Do not move the cursor up: this eats previous output

Compute the terminal width when needed (it may change during the course of the test-suite).

Also print something in CI log.

Thanks Alexandre for spotting fishy ANSI codes.


eponier (approved):
I don't understand the patch, but I checked the output, and the glitches that I observe from time to time are seemingly gone. For me, it's good to go. Unless you want a more informed review from @cassiersg ?

vbgl:
Thanks for testing.

Issue #743, opened 2024-02-29
Initialisation at declaration site


eponier:
This is a feature request to have the option to initialize a variable at declaration time, e.g. `reg u64 res = 1;`.

eponier:
One advantage of allowing that could be to be more strict about const arrays. For instance, for now, we allow to write
`r = g`, where `r` is a `reg mut ptr` and `g` is a `global`. We cannot define `r` as `reg constr ptr`, otherwise `r = g` is rejected. With initialization at declaration, we could write `reg const ptr u64[2] r = g` and refuse `r = g`.

eponier:
(While writing that, I realize it is not clear in my mind if `const` means that `r` is a constant pointer, a pointer to constant arrays, or both.)

vbgl:
`const` means that `r` has partial ownership (aka read-only) on the array.

eponier:
For now, we reject both `r = s` and `r[i] = ...`. Maybe we should just reject the latter.

vbgl:
Agreed (this is unrelated to the topic of this issue, though).

MrDaiki:
In term of proper implementation, could we consider this feature to be syntactic sugar for initialize reg + affect to the right expression value ?. 

For exemple : 
```jazz
reg u64 res = 1;
```

Would be replaced by : 
```jazz
reg u64 res;
res = 1;
``` 
and then compiled normally. 

Another interrogation is the specification in the case where multiples var are declared as the same time : 

for example 
```jazz
reg u64 res1,res2 = 1;
```
In my opinion, this kind of syntax can create problems if the right expression is a pointer. Maybe a preferable syntax would be : 
```jazz
reg u64 res1 = 1, res2 = 3 ... ;
```

Another solution would be to forbid multiple declarations in this case

vbgl:
Done in #908.

PR #899, opened 2024-09-12
Authorizing variable initialisation at declaration

MrDaiki:
# Issue 
describe in Issue #743. The idea of this new feature is to authorize initialization of variables at their declaration. For example:
```jazz
reg u64 x=1;
```
# Implementation

With the proposed implementation, we can also initialize multiple variables at the same time: 
```jazz
reg u64 x=1,y=2,z,t=3;
```

The current implementation doesn't support array initializing with this feature because array can currently be explicitly created only at top level. Allowing it would require reworking a large part of the grammar and pre-typing analysis. 

In terms of semantic, the proposed implementation works as follow : 
```jazz
reg u64 x=3;
```
is abstractly equivalent to :
```jazz
reg u64 x;
x=3;
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ compiler/src/CIL/*.mli
 *.glob
 *.aux
 *.lock
+*.vscode
 _build
 *.run.out
 *.run.err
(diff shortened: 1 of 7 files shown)


vbgl:
Subsumed by #908.

PR #901, opened 2024-09-16
Refactoring of tt_instr

MrDaiki:
Creation of a tt_assign function to separate the pre-typing of assign nodes from tt_instr function (which is already a big function)


vbgl:
Subsumed by #905.

PR #903, opened 2024-09-17
Remove dummy locations from errors before printing them

vbgl:
Best would be to avoid these dummy locations in the first place. Meanwhile, this is an attempt at saving the user from seeing scary locations.


bgregoir (commented):
I don't know if we should do this.

bgregoir:
Sorry, I was thinking that current situation can help to find them. But a grep will be simpler ...

bgregoir (approved):

PR #906, opened 2024-09-18
Mark gvar OCaml type as covariant

vbgl:
This allows polymorphic constant values of type 'a gvar.


eponier:
Ok, it needs the annotation because of `private`. Without it, the type is inferred to be contravariant by the compiler. I wonder whether this is a bug of the OCaml compiler.

eponier (approved):

Issue #871, opened 2024-07-31
Invalid assembly when several globals have the same name


vbgl:
The compiler produces invalid assembly when compiling the following program:

~~~
inline fn get(inline u32 i) -> reg u32 {
  reg u32 r;
  global u32 g;
  g = i;
  r = g;
  return r;
}

export
fn main() -> reg u32 {
  reg u32 a b;
  a = get(0xa);
  b = get(0xb);
  b += a;
  return b;
}
~~~

Processing the generated assembly yields:

> bug_871.s: Assembler messages:
bug_871.s:21: Error: symbol `G$g' is already defined

vbgl:
Fixed by #907.

PR #907, opened 2024-09-18
Debug of issue #871

MrDaiki:
The `PrintASM.format_glob_data` function was able to write the same label multiple times, causing the produced assembly code to be incorrect. 

By adding a Map to count occurrences of the same variable, the problem is corrected.

There is a test to write (for example the file written by @vbgl in the issue). I wasn't sure where to put it so I didn't (We will need to add it before merging).

### DIFF
--- a/compiler/src/printASM.ml
+++ b/compiler/src/printASM.ml
@@ -25,11 +25,18 @@ let print_asm_lines fmt lns =
 let string_of_label name p =
   Format.asprintf "L%s$%d" (escape name) (Conv.int_of_pos p)
 
-let string_of_glob x =
-  Format.asprintf "G$%s" (escape x.v_name)
+let string_of_glob occurrences x =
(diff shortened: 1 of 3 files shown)


compiler/src/printASM.ml
```diff
@@ -25,11 +25,18 @@ let print_asm_lines fmt lns =
...
+  let map = Map.add x.v_name (count+1) !occurences in
+  occurences:= map;
+  (*Adding the number of occurences to the label to avoid names conflict*)
+  Format.asprintf "G$%s$%d" (escape x.v_name) (count)
+
 
 (* -------------------------------------------------------------------- *)
 let format_glob_data globs names =
+  (* Creating a map to count occurences of a label name*)
+  let occurences = ref Map.empty in
```
vbgl, line 39: Please fix the spelling of occurrences and use a mutable data-structure.

vbgl (approved):
Fixes #871.

PR #908, opened 2024-09-19
Variable intialization at declare time

MrDaiki:
# Issue 

This is a duplicate of issue #899 based on #901 refactoring of pre-typing.
The feature requested is to add a new syntax that authorize variable initialization when they are declared.

The syntax is as follow :

```
reg u64 x=1,y=2,z,t=3;
```

# Implementation

With the proposed implementation, we can also initialize multiple variables at the same time: 
```jazz
reg u64 x=1,y=2,z,t=3;
```

The current implementation doesn't support array initializing with this feature because array can currently be explicitly created only at top level. Allowing it would require reworking a large part of the grammar and pre-typing analysis. 

In terms of semantic, the proposed implementation works as follow : 
```jazz
reg u64 x=3;
```
is abstractly equivalent to :
```jazz
reg u64 x;
x=3;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - Extraction as EasyCrypt code targets version 2024.09
   ([PR #910](https://github.com/jasmin-lang/jasmin/pull/910)).
 
+- Local variables may be initialized when declared, as for instance
+  `reg u32 x = 1, y, z = 2; stack u64 s = 1, u, t = z;`
+  ([PR #908](https://github.com/jasmin-lang/jasmin/pull/908)).
+
 ## Bug fixes
 
 - Easycrypt extraction for CT : fix decreasing for loops
--- a/compiler/src/latex_printer.ml
+++ b/compiler/src/latex_printer.ml
@@ -235,6 +235,18 @@ let pp_args fmt (sty, xs) =
     pp_sto_ty sty
     (pp_list " " pp_var) xs
 
+let pp_decl fmt (x: vardecl L.located) =
+  let x, e = L.unloc x in
+  let pp fmt = F.fprintf fmt " = %a" pp_expr in
+  F.fprintf fmt "%s%a" (L.unloc x) (pp_opt pp) e
+
+let pp_decls fmt (sty, vd) =
+  F.fprintf
+    fmt
+    "%a %a"
+    pp_sto_ty sty
+    (pp_list " " pp_decl) vd
+
 let pp_rty =
   pp_opt
     (fun fmt tys ->
@@ -263,8 +275,8 @@ let pp_eqop fmt op =
 let pp_sidecond fmt =
   F.fprintf fmt " %a %a" kw "if" pp_expr
 
-let pp_vardecls fmt d =
-  F.fprintf fmt "%a;" pp_args d
+let pp_vardecls fmt (d:vardecls) =
+  F.fprintf fmt "%a;" pp_decls d
 
 let rec pp_instr depth fmt (annot, p) =
   if annot <> [] then F.fprintf fmt "%a%a" indent depth pp_top_annotations annot;
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -1674,11 +1674,18 @@ let mk_call loc inline lvs f es =
 
   P.Ccall (lvs, f.P.f_name, es)
 
-let tt_annot_vardecls dfl_writable pd env (annot, (ty,vs)) = 
+let assign_from_decl (decl: S.vardecl L.located) =
+  let v, e = L.unloc decl in
+  Option.map (fun e ->
+      let d = L.mk_loc (L.loc decl) (S.PLVar v) in
+      (None, [d]), `Raw, e, None
+    ) e
+
+let tt_annot_paramdecls dfl_writable pd env (annot, (ty,vs)) =
   let aty = annot, ty in
   let vars = List.map (fun v -> aty, v) vs in
   tt_vardecls_push dfl_writable pd env vars 
-  
+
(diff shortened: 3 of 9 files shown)


compiler/src/latex_printer.ml
```diff
@@ -235,6 +235,19 @@ let pp_args fmt (sty, xs) =
     pp_sto_ty sty
     (pp_list " " pp_var) xs
 
+  let pp_decl fmt (x:vardecl )= 
+    match x with
+    | NotInitVarDecl v -> F.fprintf fmt "%s" (L.unloc v)
+    | InitVarDecl (v,e) ->  F.fprintf fmt "%s" (L.unloc v); pp_expr fmt e
```
vbgl, line 241: This line looks wrong.

compiler/src/latex_printer.ml
```diff
@@ -235,6 +235,19 @@ let pp_args fmt (sty, xs) =
     pp_sto_ty sty
     (pp_list " " pp_var) xs
 
+  let pp_decl fmt (x:vardecl )=
```
vbgl, line 238: You may take a located `vardecl` here.

vbgl (commented):
Please rebase on top of recent `main` branch.

compiler/src/latex_printer.ml
MrDaiki: I was doing L.unloc before calling pp_decl, I change it so it is done in the function

vbgl:
I did some cleaning and spotted a bug (see the fresh test case): checking of the LHS should occur before the RHS is added to the environment.

MrDaiki:
> I did some cleaning and spotted a bug (see the fresh test case): checking of the LHS should occur before the RHS is added to the environment.

To solve this case, I think we will need to call `tt_expr` instead of `tt_assign` (or we consider that tt_assign checks can be redundant)

vbgl (commented):
Just add a changelog entry and this will be ready.

compiler/src/latex_printer.ml
```diff
@@ -235,6 +235,18 @@ let pp_args fmt (sty, xs) =
...
+  match L.unloc x with
+  | NotInitVarDecl v -> F.fprintf fmt "%s" (L.unloc v)
+  | InitVarDecl (v, e) -> F.fprintf fmt "%s = %a" (L.unloc v) pp_expr e
+
+let pp_decls fmt (sty, vd) =
+  F.fprintf
+    fmt
+    "%a %a"
+    pp_sto_ty sty
+    (pp_list " " pp_decl) vd
```
vbgl, line 248: Is it OK to have a blank separator when there is an initializing expression?

MrDaiki: To be honest, I have no opinion about this matter. Would you prefer it without the blank separator ?

vbgl: Yes it is ;-)

vbgl: I was not asking about opinion but about correctness. It is correct: I’ve thus enriched the example file.

MrDaiki: Sorry, I didn't understand your question ^^'

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 - Extraction as EasyCrypt code targets version 2024.09
   ([PR #910](https://github.com/jasmin-lang/jasmin/pull/910)).
 
+- Local variables may be initialized when declared
+  ([PR #908](https://github.com/jasmin-lang/jasmin/pull/908)).
```
bgregoir, line 16: Add the syntax:
reg u32 x = 1, y, z = 2;
stack u64 s = 1, u, t = z;

compiler/src/syntax.ml
```diff
@@ -214,7 +214,15 @@ type align = [`Align | `NoAlign]
 
 type plvals = annotations L.located option * plvalue list
 
-type vardecls = pstotype * pident list
+
+type vardecl = InitVarDecl of pident * pexpr | NotInitVarDecl of pident
```
bgregoir, line 218: I think the definition
type vardecl = pident * pexpr option
will be more convenient.
You can do the change if you want.

compiler/tests/fail/common/var_initialize_undef.jazz
```diff
@@ -0,0 +1,4 @@
+export fn main() -> reg u32 {
+  reg u32 x = x;
```
bgregoir, line 2: rename into var_initialize_undecl

compiler/src/pretyping.ml
```diff
@@ -1812,31 +1819,45 @@ let rec tt_instr arch_info (env : 'asm Env.env) ((annot,pi) : S.pinstr) : 'asm E
...
       let e' = oget ~exn:(tyerror ~loc exn) (P.expr_of_lval x) in
-      let c = tt_expr_bool arch_info.pd env cp in
-      env, mk_i (P.Cassgn (x, AT_none, ty, Pif (ty, c, e, e'))) :: is
+      let c = tt_expr_bool arch_info.pd env_rhs cp in
+      mk_i (P.Cassgn (x, AT_none, ty, Pif (ty, c, e, e'))) :: is
   in
+  let tt_annot_decl env (vd: S.vardecl L.located) (aty: A.annotations * S.pstotype) =
+    (* remember the environment prior to the declaration:
+      it will be used to type-check the initializing expression, if any *)
+    let old_env = env in
```
bgregoir, line 1839: maybe old_env can be renamed in env_rhs and env in env_lhs.
But it is no important

bgregoir (dismissed):
This is nice, I approve it. 
If you don't want to do the small suggested changes, I will merge it as it is.

vbgl:
I’ve done the small changes you suggested. I let Alexandre decide whether he wants to change the data-type.

vbgl (approved):
Seems ready!

Issue #501, opened 2023-07-11
Support for typedef


swarnpriya:
This is a **feature request** to support `typedef` in Jasmin. This will help the developers to create short aliases of long-type names which can be easily used throughout the program. 

An example:
Type like reg ptr u32[12] can be written as typedef reg ptr u32[12] MyArrayPtr.

After this type definition, the identifier MyArrayPtr can be used as an abbreviation for the type reg ptr u32[12].

For example: `MyArrayPtr ap1`.

MrDaiki:
If I understand correctly, what you would like is a form of type aliasing. There is several possibilities of syntax for it : 
```jazz
alias x = reg ptr u32[12];
typedef x = reg ptr u32[12];
alias reg ptr u32[12] as x;
typedef reg ptr u32[12] as x;
typedef reg ptr u32[12] x;
```

I personally prefer the first two solutions because of their readability. 

In term of semantic, it would be useful to decide whether or not we authorize alias of alias. I personally don't think that would be a good idea because it could ironically make user comprehension of a program harder.

bgregoir:
I think rust syntax is 
type x = reg ptr u32[12];
Since we try to be clause to the rust syntax maybe it is better.
The other question is: does reg ptr really part of the type.
Maybe a first step will be to only allows type definition
type x = u32[12];
I think the change in the parser will be much smaller.

vbgl:
That’s right. Here in a use case, in which a single type is used with different storage attributes:

~~~
type Key = u8[33];

fn check(reg ptr Key key) // …

  stack Key key; // …
  check(key);
~~~

vbgl:
Bejamin is right about Rust syntax: https://doc.rust-lang.org/book/ch19-04-advanced-types.html#creating-type-synonyms-with-type-aliases

Can we avoid introducing a new key-word?

MrDaiki:
I agree that rust syntax is really good. I also agree that storage type should not be part of the alias (at least for the moment). @vbgl presented a good example of alias with multiple storage types.

Another question I would like to ask about the feature is how do we resolve it's semantic. If I understand correctly how jasmin work, for the moment, we can't declare an array with size defined by a variable. For example :
```jazz
fn test(reg u32[N] x) {
    reg u32 N;
    N=3;
    reg u32[N] x;
}
``` 

This code is not valid because N is not a top level `param`. With type alias, we could however encounter an issue. By extending this example a little : 
```jazz 
param int N = 32;
[...]

PR #911, opened 2024-09-23
Adding support for type definition

MrDaiki:
# Feature

As described in #501, the objective of this feature is to allow declaration of type alias within jasmin syntax. The chosen keyword 'type' follow rust language specification. 

You can declare a type alias as follow :
``` 
type typename = u64;
type typename = u64[4];
type typename = othertypename;
``` 

types can then be called during variable declaration :
```
reg typename var;
reg typename[5] var;
``` 
This feature increase portability of programs. For example : 
``` 
// test.jazz
require "arch.jinc";
export fn inc(reg arch_type x) -> reg arch_type {
    reg arch_type y;
    y=x;
    y+=1;
    return y;
}
```
we can write a required header file that define arch_type (or even making it a part of compilation reserved types) for each targeted architecture.
```
// arch.jinc for x86_amd64 target
type arch_type = u64;
```
```
// arch_armv7.jinc armV7 target
type arch_type = u32;
```

avoiding rewriting of code

# Implementation details : 

The declaration of storage type cannot be aliases, making the following declaration invalid (raising TypeError):
```
type typename = reg u64 x;
```
We can consider that 'reg' 'inline' 'stack' are equivalent to 'let' or 'var' keyword in other languages, so it would not make sense to make it aliasable (at least in my point of view).

Types semantic is resolved during pre-typing step of compilation. For the moment, alias are resolved at the moment they are read by the compiler, making the following declarations invalid : 
```
type type1 = type2;
type type2 = u64;
```

We could modify the implementation to allow such semantic (by treating type resolution lazily).

### DIFF
--- a/compiler/src/latex_printer.ml
+++ b/compiler/src/latex_printer.ml
@@ -203,15 +203,18 @@ and pp_mem_access fmt (al, ty,x,e) =
     | Some (`Add, e) -> Format.fprintf fmt " + %a" pp_expr e 
     | Some (`Sub, e) -> Format.fprintf fmt " - %a" pp_expr e in
   F.fprintf fmt "%a[%a%a%a]" (pp_opt (pp_paren pp_ws)) ty pp_aligned al pp_var x pp_e e
-  
+
+
 and pp_type fmt ty =
   match L.unloc ty with
   | TBool -> F.fprintf fmt "%a" ptype "bool"
   | TInt -> F.fprintf fmt "%a" ptype "int"
   | TWord w -> pp_ws fmt w
-  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (string_of_wsize w) pp_expr e
+  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (Syntax.string_of_sizetype w) pp_expr e
+  | TAlias id -> F.fprintf fmt "%a" ptype (L.unloc id)
 
 and pp_ws fmt w = F.fprintf fmt "%a" ptype (string_of_wsize w)
+
 and pp_expr fmt e = pp_expr_rec Pmin fmt e
 
 and pp_arr_access fmt al aa ws x e len=
@@ -419,6 +422,9 @@ let pp_global fmt { pgd_type ; pgd_name ; pgd_val } =
 let pp_path fmt s =
   F.fprintf fmt "%S " (L.unloc s)
 
+let pp_typealias fmt id ty =
+  F.fprintf fmt "%a %a = %a;" kw "type" dname (L.unloc id) pp_type ty
+
 let rec pp_pitem fmt pi =
   match L.unloc pi with
   | PFundef f -> pp_fundef fmt f
@@ -439,6 +445,7 @@ let rec pp_pitem fmt pi =
      List.iter (pp_pitem fmt) pis;
      F.fprintf fmt eol;
      closebrace fmt ()
+  | PTypeAlias (id,ty) -> pp_typealias fmt id ty (**)
 
 let pp_prog fmt =
   F.pp_print_list ~pp_sep:(fun fmt () -> F.fprintf fmt eol) pp_pitem fmt
--- a/compiler/src/lexer.mll
+++ b/compiler/src/lexer.mll
@@ -18,6 +18,7 @@
       raise (Syntax.ParseError (loc, Some (Format.asprintf "ill-formed string (%s)" msg)))
 
   let _keywords = [
+    "type"  , TYPE   ;
     "u8"    , T_U8   ;
     "u16"   , T_U16  ;
     "u32"   , T_U32  ;
--- a/compiler/src/parser.mly
+++ b/compiler/src/parser.mly
@@ -42,6 +42,7 @@
 %token FN
 %token FOR
 %token FROM
+%token TYPE
 %token <Syntax.sign>GE
 %token GLOBAL
 %token <Syntax.sign>GT
@@ -166,6 +167,10 @@ utype:
 | T_U128 { Wsize.U128 }
 | T_U256 { Wsize.U256 }
 
+utype_array:
+| ws=utype {TypeWsize ws}
+| id=ident {TypeSizeAlias id}
+
 ptype_r:
 | T_BOOL
     { TBool }
@@ -176,8 +181,9 @@ ptype_r:
 | ut=utype
     { TWord ut }
 
-| ut=utype d=brackets(pexpr)
+| ut=utype_array d=brackets(pexpr)
     { TArray (ut, d) }
+| x=ident {TAlias x}
 
 ptype:
 | x=loc(ptype_r) { x }
@@ -503,6 +509,8 @@ top:
 | x=pglobal  { Syntax.PGlobal x }
 | x=pexec    { Syntax.Pexec   x }
 | x=prequire { Syntax.Prequire x}
+| TYPE name = ident EQ ty = ptype SEMICOLON
+    { Syntax.PTypeAlias (name, ty)}
 | NAMESPACE name = ident LBRACE pfs = loc(top)* RBRACE
     { Syntax.PNamespace (name, pfs) }
 (* -------------------------------------------------------------------- *)
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -31,6 +31,9 @@ type tyerror =
   | InvalidArgCount     of int * int
   | InvalidLvalCount    of int * int
   | DuplicateFun        of A.symbol * L.t
+  | DuplicateAlias      of A.symbol * P.pty L.located * P.pty L.located
+  | TypeNotFound        of A.symbol
+  | InvalidTypeAlias    of A.symbol * P.pty
   | InvalidCast         of P.pty pair
   | InvalidTypeForGlobal of P.pty
   | NotAPointer         of P.plval
@@ -158,6 +161,24 @@ let pp_tyerror fmt (code : tyerror) =
         "The function %s is already declared at %s"
         f (L.tostring loc)
 
+  | DuplicateAlias (id, newtype, oldtype) ->
+      F.fprintf fmt
+        "Type '%s' (ie: '%a') is already declared at %s (with type : '%a')"
+        id
+        Printer.pp_ptype (L.unloc newtype)
+        (L.tostring (L.loc oldtype))
+        Printer.pp_ptype (L.unloc oldtype)
+
+  | TypeNotFound (id) ->
+      F.fprintf fmt
+      "Type '%s' not found"
+      id
+
+  | InvalidTypeAlias (id,typ) ->
+      F.fprintf fmt
+      "Type '%s' (ie: '%a') is not allowed as array element. Only machine words (u8, u16 ...) allowed"
+      id Printer.pp_ptype typ
+
   | EqOpWithNoLValue ->
       F.fprintf fmt
         "operator-assign requires a lvalue"
@@ -252,6 +273,11 @@ module Env : sig
     val clear_locals : 'asm env -> 'asm env
   end
 
+  module TypeAlias : sig
+    val push : 'asm env -> A.pident -> P.pty -> 'asm env
+    val get : 'asm env -> A.pident -> P.pty L.located
+  end
+
   module Funs : sig
     val push : 'asm env -> (unit, 'asm) P.pfunc -> P.pty list -> 'asm env
     val find : A.symbol -> 'asm env -> ((unit, 'asm) P.pfunc * P.pty list) option
@@ -272,6 +298,7 @@ end  = struct
     } 
 
   type 'asm global_bindings = {
+      gb_types : (A.symbol, P.pty L.located) Map.t;
       gb_vars : (A.symbol, P.pvar * E.v_scope) Map.t;
       gb_funs : (A.symbol, (unit, 'asm) P.pfunc * P.pty list) Map.t;
     }
@@ -295,7 +322,7 @@ end  = struct
     ; from = Map.empty
     }
 
-  let empty_gb = { gb_vars = Map.empty ; gb_funs = Map.empty }
+  let empty_gb = { gb_vars = Map.empty ; gb_funs = Map.empty; gb_types = Map.empty }
 
   let empty : 'asm env =
     { e_bindings = [], empty_gb
@@ -325,20 +352,24 @@ end  = struct
         let n = qualify ns n in
         begin match Map.find n dst with
         | exception Not_found -> ()
-        | (k, _) -> on_duplicate n (fst v) k end;
+        | k -> on_duplicate n v k end;
         Map.add n v dst)
 
-  let warn_duplicate_var name v v' =
+  let warn_duplicate_var name (v, _) (v', _) =
     warning DuplicateVar (L.i_loc0 v.P.v_dloc)
       "the variable %s is already declared at %a"
       name L.pp_loc v'.P.v_dloc
 
-  let err_duplicate_fun name v fd =
+  let err_duplicate_fun name (v, _) (fd, _) =
     rs_tyerror ~loc:v.P.f_loc (DuplicateFun(name, fd.P.f_loc))
+
+  let err_duplicate_type name t1 t2 =
+    rs_tyerror ~loc:(L.loc t2) (DuplicateAlias (name,t1,t2))
 
   let merge_bindings (ns, src) dst =
     { gb_vars = merge_bindings warn_duplicate_var ns src.gb_vars dst.gb_vars
     ; gb_funs = merge_bindings err_duplicate_fun ns src.gb_funs dst.gb_funs
+    ; gb_types = merge_bindings err_duplicate_type ns src.gb_types dst.gb_types
     }
 
   let exit_namespace env =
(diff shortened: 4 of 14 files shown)


vbgl (commented):
Please add tests  involving name spaces.

compiler/src/pretyping.ml
```diff
@@ -2146,6 +2258,7 @@ let rec tt_item arch_info (env : 'asm Env.env) pt : 'asm Env.env =
      let env = List.fold_left (tt_item arch_info) env items in
      let env = Env.exit_namespace env in
      env
+  | S.PTypeAlias (id,ty) -> tt_typealias arch_info env id ty (*TODO : Implement type alias, when a proper semantic will be decided*)
```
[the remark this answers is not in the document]
MrDaiki: I forgot to remove this one

compiler/src/lexer.mll
```diff
@@ -18,6 +18,7 @@
       raise (Syntax.ParseError (loc, Some (Format.asprintf "ill-formed string (%s)" msg)))
 
   let _keywords = [
+    "type"  , TYPE   ;
```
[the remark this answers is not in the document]
vbgl: I meant there is no need for any keyword. And the following becomes valid:
~~~
a = u64;
a a = 42;

b = int;
param b b = 33;
~~~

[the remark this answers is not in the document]
MrDaiki: Do we really want these statements to be considered valid ? In my opinion, having a keyword make the language more clean, thus easier to understand. It is not a heavy modification to do and I can remove the keyword if you really want to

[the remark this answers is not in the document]
bgregoir: In am agree with you, I clearly prefer the version with type. 
@Vincent, adding a key word is really a problem ?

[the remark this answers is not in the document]
vbgl: Hej. Don’t ping random people. Apologies to Vincent (the other one) for the mistake.

MrDaiki:
> Please add tests involving name spaces.

Done in the last commit

MrDaiki:
Before being able to merge this PR, we need to discuss one particular. I implemented this feature by adding a map that store type alias ( id -> P.pty mapping). When we encounter an id during the pre-typing, it is substituted by the corresponding pty if it exist.

The problem is that annotation system also permit the declaration of ids as parameter, meaning that :
```
type x = u32; // Or just x = 32; depending on the syntax we will choose

#annot #param=x
fn ....
``` 
is currently ambiguous. The choice of substituting x by it's type could depend on the context, making the possibility of tricky behaviors appearing. 

The solution I would suggest would be to explicit whether a parameter of annotation is a type or an id. For example with the following syntaxes : 
```
#annot #param= type x 
#annot #param= `x
```
Another solution would be to consider that we try to substitute the id by a type, and if we can't, we let it like that (it is the current version approach). It work for the moment but would clearly create weird behavior if we decide to resolve type lazily. For example : 

``` 
#param #annot=x
fn ....

type x = u32;
```

What is the expected behavior. Will it not cost a lot in terms of implementation ?

compiler/src/latex_printer.ml
```diff
@@ -386,6 +389,9 @@ let pp_global fmt { pgd_type ; pgd_name ; pgd_val } =
 let pp_path fmt s =
   F.fprintf fmt "%S " (L.unloc s)
 
+let pp_typealias fmt id ty =
+  F.fprintf fmt "type %s = %a;" (L.unloc id) pp_type ty
```
vbgl, line 393: You must use `kw` and `dtype` here.

compiler/src/pretyping.ml
```diff
@@ -479,6 +529,39 @@ end  = struct
...
+      match typea with 
+      | None -> 
+        rs_tyerror  ~loc:(L.loc id) (TypeNotFound (L.unloc id))
+      | Some e -> e
+
+    let get_opt (env: 'asm env) (id:A.pident) : (P.pty L.located) option = 
+      match get env id with 
+      | exception _ -> None 
+      | e -> Some e
+
```
vbgl, line 562: This is `find`.

vbgl (commented):
Here are a few cosmetic comments.

compiler/src/latex_printer.ml
```diff
@@ -185,15 +185,18 @@ and pp_mem_access fmt (al, ty,x,e) =
...
   F.fprintf fmt "%a[%a%a%a]" (pp_opt (pp_paren pp_ws)) ty pp_aligned al pp_var x pp_e e
+
   
 and pp_type fmt ty =
   match L.unloc ty with
   | TBool -> F.fprintf fmt "%a" ptype "bool"
   | TInt -> F.fprintf fmt "%a" ptype "int"
-  | TWord w -> pp_ws fmt w
-  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (string_of_wsize w) pp_expr e
+  | TWord w -> F.fprintf fmt "%s" (string_of_wsize w)
```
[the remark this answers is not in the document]
MrDaiki: It is probably wrong (I think I had one version of the types where I introduced type alias inside of utype grammar rule)

compiler/src/pretyping.ml
```diff
@@ -158,6 +161,21 @@ let pp_tyerror fmt (code : tyerror) =
...
+  | TypeNotFound (id) -> 
+      F.fprintf fmt 
+      "Type %S not found"
+      id
+  
+  | InvalidTypeAlias (id,typ) -> 
+      F.fprintf fmt 
+      "Type %S (ie: %a) is not allowed as array element. Only machine words (uXX...) allowed"
+      id Printer.pp_ptype typ
+
```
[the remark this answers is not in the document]
MrDaiki: It is just a way to separate well the name of the type

compiler/src/pretyping.ml
MrDaiki: I did it this way in case that the Env.Typealias.get function is modified in the future

compiler/src/pretyping.ml
```diff
@@ -325,20 +351,43 @@ end  = struct
...
+  
+  let warn_duplicate_var_merge name (v,_) (v',_) =
+    warn_duplicate_var name v v'
+  let err_duplicate_fun_merge name (v,_) (fd,_) =
+    err_duplicate_fun name v fd
+
+  let err_duplicate_type name t1 t2 = 
+    rs_tyerror ~loc:(L.loc t2) (DuplicateAlias (name,L.loc t1))
+
+  let merge_types_bindings ns src dst =
```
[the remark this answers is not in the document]
MrDaiki: I am not sure this is right, I will look it

compiler/src/pretyping.ml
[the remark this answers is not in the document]
vbgl: This is not the purpose of `%S`.

compiler/src/pretyping.ml
```diff
@@ -325,20 +351,43 @@ end  = struct
...
   let err_duplicate_fun name v fd =
     rs_tyerror ~loc:v.P.f_loc (DuplicateFun(name, fd.P.f_loc))
+  
+  let warn_duplicate_var_merge name (v,_) (v',_) =
+    warn_duplicate_var name v v'
+  let err_duplicate_fun_merge name (v,_) (fd,_) =
+    err_duplicate_fun name v fd
+
+  let err_duplicate_type name t1 t2 = 
+    rs_tyerror ~loc:(L.loc t2) (DuplicateAlias (name,L.loc t1))
```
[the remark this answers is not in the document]
MrDaiki: Currently the error message look like this : 
```
Type "x" is already declared at "tests/fail/common/typealias_duplicate.jazz", line 2 (5-6)
```
You would prefer something like that ? 
```
Type "x" (ie: int) is already declared at "tests/fail/common/typealias_duplicate.jazz", line 2 (5-6)
```
I am not sure to see the use of having the type of the object. Or you would prefer that an error is raised if and only if 
`t1 != t2` ? I am not sure to understand well your point

compiler/src/latex_printer.ml
```diff
@@ -185,15 +185,18 @@ and pp_mem_access fmt (al, ty,x,e) =
...
+  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (Syntax.string_of_sizetype w) pp_expr e
```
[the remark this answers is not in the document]
MrDaiki: I think it is necessary here because TArray is now : 
AliasSizeType pident | SizeType wsize (the type of TArray can be an alias pointing to a wsize type. I could put it in a separate function if you want

compiler/src/pretyping.ml
MrDaiki: For example if we resolve type lazily, we would need to add another map to track unfound alias, making `get` behavior different

compiler/src/pretyping.ml
```diff
@@ -2128,6 +2236,29 @@ let tt_global pd (env : 'asm Env.env) _loc (gd: S.pglobal) : 'asm Env.env =
...
+        let talias = Env.TypeAlias.get env alias in 
+        match L.unloc talias with 
+        | P.Bty (P.U ws) -> P.Arr (arch_info.pd, fst (tt_expr ~mode:`OnlyParam ws env e))
+        | _ -> rs_tyerror  ~loc:(L.loc id) (InvalidTypeAlias ((L.unloc id), L.unloc talias))
+        end 
+      in Env.TypeAlias.push env id arr
+    end
+  | S.TAlias alias_id -> 
+    let talias = Env.TypeAlias.get env alias_id in 
+    Env.TypeAlias.push env id (L.unloc talias)
```
vbgl, line 2260: Can’t you call `tt_type` instead of reimplementing it?

compiler/src/pretyping.ml
```diff
@@ -252,6 +273,13 @@ module Env : sig
     val clear_locals : 'asm env -> 'asm env
   end
 
+  module TypeAlias : sig 
+
+    val push : 'asm env -> A.pident -> P.pty -> 'asm env
+    val get : 'asm env -> A.pident -> (P.pty L.located)
+
+    val get_opt : 'asm env -> A.pident -> (P.pty L.located) option
```
vbgl, line 281: This function (`get_opt`) is never called.

MrDaiki:
I did modifications according to your comment. 

The final thing that need to be decided before merging will be the syntax of the feature : 
I personally consider the following syntaxes to be discussed: 
1. `param type x = u64;`
2. `param x = u64;`
3. `type x = u64;`
4. `x = u64;`

The default of adding a new keyword to the language is that it could potentially broke some already written program (that declare a variable `type`). On the other hand, it make the language easier to understand by making type explicitly declared. It also follow the syntax of Rust (and also Ocaml), which is an important inspiration for Jasmin. 

The question of adding `param` keyword should be discuss if we consider the implementation of a future feature of parametric modules, giving a more general semantic behind keyword `param`.

I personally prefer option 1 or 3 for the syntax.

compiler/src/pretyping.ml
MrDaiki: corrected

bgregoir:
I clearly prefer 3. 
1/ First I think that adding the keyword "type" is not so problematic. 
2/ We should try to follow rust syntax.

compiler/src/pretyping.ml
MrDaiki: Done

compiler/src/pretyping.ml
MrDaiki: This function doesn't exist anymore so this is closed by default

compiler/src/pretyping.ml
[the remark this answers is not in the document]
MrDaiki: Corrected (the error message became long though, maybe we could do multiline errors in this case)

compiler/src/latex_printer.ml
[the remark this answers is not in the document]
MrDaiki: It is corrected

compiler/src/latex_printer.ml
```diff
@@ -185,15 +185,18 @@ and pp_mem_access fmt (al, ty,x,e) =
...
+  | TAlias id -> F.fprintf fmt "%s" (L.unloc id)
```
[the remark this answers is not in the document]
MrDaiki: done

compiler/src/latex_printer.ml
MrDaiki: Did you meant `ptype` ?

vbgl: No, `dname`. Sorry about that.

compiler/tests/success/x86-64/typealias_amd64.jazz
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    return (target,source);
+}
+
+inline fn f(reg u64 p) {
+    stack u64 s;
+    s = 1;
+    [p] = s;
+}
```
vbgl, line 30: What is this?

MrDaiki: Something I didn't have before (probably a typo)

MrDaiki: it doesn't test the feature so I will remove it

compiler/src/latex_printer.ml
```diff
@@ -438,7 +444,9 @@ let rec pp_pitem fmt pi =
      F.fprintf fmt eol;
      List.iter (pp_pitem fmt) pis;
      F.fprintf fmt eol;
-     closebrace fmt ()
+     closebrace fmt ();
+     F.fprintf fmt eol
```
vbgl, line 448: This looks wrong.

MrDaiki: Yes : Git automerge it very weirdly, looking if there is other problems

MrDaiki: it is corrected

compiler/src/pretyping.ml
```diff
@@ -272,6 +298,7 @@ end  = struct
     } 
 
   type 'asm global_bindings = {
+      gb_types : (A.symbol, P.pty L.located) Map.t;
```
vbgl, line 301: Is the location needed here?

compiler/src/latex_printer.ml
```diff
@@ -202,16 +202,20 @@ and pp_mem_access fmt (al, ty,x,e) =
...
   | TBool -> F.fprintf fmt "%a" ptype "bool"
   | TInt -> F.fprintf fmt "%a" ptype "int"
   | TWord w -> pp_ws fmt w
-  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (string_of_wsize w) pp_expr e
+  | TArray (w, e) -> F.fprintf fmt "%a[%a]" ptype (Syntax.string_of_sizetype w) pp_expr e
+  | TAlias id -> F.fprintf fmt "%a" ptype (L.unloc id)
 
-and pp_ws fmt w = F.fprintf fmt "%a" ptype (string_of_wsize w)
+and pp_ws fmt (w) = F.fprintf fmt "%a" ptype (string_of_wsize w)
+and pp_sizetype fmt s = F.fprintf fmt "%a" ptype (string_of_sizetype s)
```
bgregoir, line 217: This function is not used

compiler/src/parser.mly
```diff
@@ -166,6 +171,14 @@ utype:
 | T_U128 { Wsize.U128 }
 | T_U256 { Wsize.U256 }
 
+utype_array:
+| ws=utype {TypeWsize ws}
+| id=ident {TypeSizeAlias id}
+
+utype_cast:
+| ws=utype {TypeWsize ws}
+| id=typealiascall {TypeSizeAlias id}
```
bgregoir, line 180: Why BACKQUOTE is needed ?

bgregoir (commented):
I have done some very small comment. 
Do you consider that the PR is now ready?

vbgl:
No, this is not ready: we are back to the design phase.

MrDaiki:
We are currently deciding whether or not we should include alias for sign extension, which is not clear for the moment. As the feature is now, we could integrate it but it would be incomplete, thus creating confusion

compiler/src/parser.mly
MrDaiki: In this case the backquote is needed to avoid ambiguity in the grammar. The problem arise from the fact that the parser cannot distinguish a typealias from an expression. We tried several solutions and came to the conclusion that the backquote was the simplest. We could probably avoid it, but it would mean rewriting a large part of the grammar, and probably also modifying pre-typing

compiler/src/pretyping.ml
MrDaiki: Yes it is for producing more precise error (with correct location from duplicated variables)

MrDaiki:
So, to summarize what we decided with @vbgl, we reverted the pull requests changes on array and memory access type cast. We still need to decide of a design for these particular cases. With this, we can merge this version of the feature. It includes :
* alias definition : `type regtype = u64`
* alias usage in basic cases : `reg regtype v = 3; reg regtype[4] tab;`
* support for alias namespaces : `reg A::regtype v ....`

What will need to be address in future pull requests :
- type alias in annotation  (we still lack a design to address the ambiguity between annotation identifiers and type aliases)
- Memory and array access (having alias in array type cast creates ambiguity in the grammar)
- Sign extension (extensions are not considered as types in the grammar but we could imagine extending the definition of alias to authorise alias in extensions)

vbgl (approved):

bgregoir (approved):

PR #912, opened 2024-09-24
Cleanup tests

cassiersg:
- Do not clutter the compiler/ directory with temp files: the test results are now in deterministic locations, in a file hierarchy under compiler/tests/results
- Tell easycrypt to use eclib from this repo.

### DIFF
--- a/compiler/scripts/extract-and-check
+++ b/compiler/scripts/extract-and-check
@@ -1,25 +1,27 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -e
 
-DIR=$(mktemp -d jasminXXXXXX)
+ROOT_DIR=tests/results/extraction
+DIR=$ROOT_DIR/${2%.jazz}/$1
+ECLIB=$(dirname $0)/../../eclib
 EC=${DIR}/jasmin.ec
 CT=${DIR}/jasmin_ct.ec
 
-if [ "$1" = "arm" ]
-then
-	shift
+if [ -d "${DIR}" ]; then rm -r ${DIR}; fi
+mkdir -p $DIR
+
+if [ "$1" = "arm" ]; then
 	ARCH="-arch arm-m4"
 else
 	ARCH="-arch x86-64"
 fi
 
-trap "rm -r ${DIR}" EXIT
-
 set -x
+set -o xtrace
 
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$@"
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -CT -oec ${CT} "$@"
+$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$2"
+$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -CT -oec ${CT} "$2"
 
-easycrypt ${ECARGS} ${EC}
-easycrypt ${ECARGS} ${CT}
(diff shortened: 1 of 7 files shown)


vbgl:
Why do you want to keep the intermediate files produces when running the test suite?

I’m not sure that the directory names are sufficiently unique so that there are no data-races.

cassiersg:
> Why do you want to keep the intermediate files produces when running the test suite?

I have sometime been using these files to investigates test failures.

Another motivation is that sometimes, the temp directories are not removed (I didn't investigate exactly when, but it happened often enough to be annoying). That could be solved by simply moving the `mktemp` to some sub-directory.

I removed `mktemp` as to ensure name stability across test executions, which is more convenient for debugging test failures (after re-running a test, the investigation can be re-run from shell command history without having to look a test logs to determine the new path). This ensures that the amount of temp files does not increase each time the tests are run.

> I’m not sure that the directory names are sufficiently unique so that there are no data-races.

Indeed. I see two possible solutions:
1. add other command-line parameters as a new directory name (e.g. `compiler/tests/results/check-x86-64/name/of/test/file/P-<param1>-<param2>`.
2. hash all command-line parameters (including name of test file) and having that as directory (e.g. `compiler/tests/results/check-x86-64/<hash of test file name and other params>`).

Any opinion on that ? (I don't have a strong opinion, I see minor pros and cons with both solutions.)

vbgl:
What about?

 3. use `mktemp`

compiler/scripts/extract-and-check
```diff
@@ -1,25 +1,27 @@
...
 
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$@"
-$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -CT -oec ${CT} "$@"
+$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -oec ${EC} "$2"
+$(dirname $0)/../jasminc ${ARCH} -oecarray ${DIR} -CT -oec ${CT} "$2"
 
-easycrypt ${ECARGS} ${EC}
-easycrypt ${ECARGS} ${CT}
+easycrypt -I Jasmin:${ECLIB} ${ECARGS} ${EC}
+easycrypt -I Jasmin:${ECLIB} ${ECARGS} ${CT}
```
vbgl, line 27: Together with this change, you can clean the CI scripts (e.g.: https://github.com/jasmin-lang/jasmin/blob/main/.gitlab-ci.yml#L184).

cassiersg:
> What about?
> 
>     3. use `mktemp`

It's a possibility, but it looses the stability: the same test will have different directories across different test runs.

Downsides:
- each test run generates new files, at some point we need to collect garbage
- after re-running the tests, all paths have changed, hence any command to debug the test resuts (e.g.  `cat` a temp file) need to change, which is annoying

We can generate the same kind of directory names as `mktemp` using a hash of the test parameters, which solves these two issues.

vbgl:
You’re right. Then I don’t have any preference.

compiler/scripts/extract-and-check
cassiersg: Done (can be done in only one place in the CI file: the other places to not run the `extract-and-check` script).

cassiersg:
I implemented solution 1. above.

For scripts with fixed parameter number, it's very simple. For variable number, I concatenate the parameters (except the test file path, still used as a prefix).

vbgl (approved):

PR #913, opened 2024-09-25
Solve TODO for conditional expressions in the SCT checker

### LINKED ISSUE #909: Solve TODO for conditional expressions in the SCT checker
Replace this TODO https://github.com/jasmin-lang/jasmin/blob/ac57159133058234583177bcc439fd728daed6fd/compiler/src/sct_checker_forward.ml#L732 by a comment clarifying that this is the correct behavior and add a test case.

ParJavadian:
fix #909


bgregoir (commented):
Can you also add a positive test ?

bgregoir:
Did you run dune promote in tests/sct-checker/ ?
I think it is why the check did not work.

vbgl (approved):

PR #917, opened 2024-09-27
Allow protecting in MMX

sarranz:


proofs/compiler/x86_extra.v
```diff
@@ -60,7 +60,7 @@ Variant x86_extra_op : Type :=
 | Ox86SLHinit
 | Ox86SLHupdate
 | Ox86SLHmove
-| Ox86SLHprotect of wsize
+| Ox86SLHprotect of bool & wsize (* Indicates if it's MMX. *)
```
vbgl, line 63: What about using `reg_kind` instead of `bool`?

proofs/compiler/x86_instr_decl.v
```diff
@@ -91,6 +91,8 @@ Variant x86_op : Type :=
 
   (* MMX instructions *)
 | MOVX  of wsize 
+| POR
```
vbgl, line 94: Is it really a different instruction than `VPOR`? Probably yes.

proofs/compiler/x86_instr_decl.v
```diff
@@ -301,6 +303,7 @@ Let prim_range range (f: wsize → A) : prim_constructor A :=
   PrimX86 (map PVp range)
         (fun s => if s is PVp sz then Some (f sz) else None).
 
+Definition prim_64 op := prim_range [:: U64 ] (fun _ => op).
```
vbgl, line 306: There is `primM` for monomorphic instructions.

proofs/compiler/x86_instr_decl.v
```diff
@@ -1118,6 +1125,9 @@ Notation mk_instr_w_w name semi ain aout nargs check prc pp_asm :=
  ((fun sz =>
   mk_instr (pp_sz name sz) (w_ty sz) (w_ty sz) ain aout (reg_msb_flag sz) (semi sz) (check sz) nargs [::] (pp_asm sz)), (name%string,prc)) (only parsing).
 
+Notation mk_instr_w2_w sz name semi ain aout nargs check prc pp_asm :=
+  (mk_instr (pp_sz name sz) (w2_ty sz sz) (w_ty sz) ain aout (reg_msb_flag sz) semi check nargs [::] (pp_asm sz), (name%string,prc)) (only parsing).
```
vbgl, line 1129: I believe POR should have `MSB_MERGE`.

proofs/compiler/x86_extra.v
sarranz: Good idea, I'll do this

proofs/compiler/x86_instr_decl.v
sarranz: Thanks

sarranz: Just to confirm: I think that for the semantics this does not matter since the size is 64. Is this right?

vbgl: That’s true. It is only relevant when targeting an xmm register, the upper half of the corresponding ymm is preserved. So there is no importance if you restrict the instruction to 64-bit and mmx registers.

proofs/compiler/x86_instr_decl.v
vbgl: I think it is better as you have done: one two-operand instruction POR and one three-operand instructions VPOR.

sarranz: Great, then I think I've addressed all the comments

proofs/compiler/x86_instr_decl.v
vbgl: I think this `prim_64` constructor is never useful and should be removed.

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@
   `reg u32 x = 1, y, z = 2; stack u64 s = 1, u, t = z;`
   ([PR #908](https://github.com/jasmin-lang/jasmin/pull/908)).
 
+- The SCT checker now allows protecting in MMX registers
+  ([PR #917](https://github.com/jasmin-lang/jasmin/pull/917)).
+
```
vbgl, line 21: It might be worth mentioning the availability of the MMX instruction POR.

sarranz: Good idea

proofs/compiler/x86_instr_decl.v
sarranz: Of course

sarranz:
After double checking, `POR` is indeed DOIT, so I fixed that as well.

vbgl (approved):

PR #921, opened 2024-10-02
Support far global variable accesses in ARM

### LINKED ISSUE #619: ARM - dealing with large programs / global arrays - invalid constant (XXXX) after fixup
I found the following issue while developing [SHA256](https://github.com/formosa-crypto/libjade/blob/feature/armv7m/src/crypto_hash/sha256/armv7m/ref/sha256.jinc) for ARM. It isn't an issue right now for that particular implementation, but I will share my notes here as they might be of use and to see if there is a different way of handling these cases or if the language/compiler needs improvement.

Consider the following program:

```
param int L = 2042; // works for 2041, fails for 2042
u32[1] H = {1};

export fn main() -> reg u32
{
  inline int i;
  reg ptr u32[1] Hp;
  reg u32 t v;

  Hp = H;
  t = Hp[0];

  v = 0;
  for i=0 to L
  { v += t; }

  return v;
}
```

We have a global array `H`, from which we get the address into `Hp` and then use it once. Between ` Hp = H;` and the `glob_data:` section, there are 1 load, 1 move, 2042 add instructions, and a pop instruction. When I try to run `arm-linux-gnueabi-gcc -g -static -mcpu=cortex-m4 -o example example.s`, where `example.s` contains the result of compiling the previous program, I get:

```
globals.s: Assembler messages:
globals.s:8: Error: invalid constant (101c) after fixup
```

The assembly instruction corresponding to `Hp = H;` is `ADR r0, glob_data+0`. For `L = 2041`, the program compiles and when observing the result of `objdump` one can see:

```
f60f 70fc       addw    r0, pc, #4092   ; 0xffc
```

Accordingly to p.197 of this [manual](https://documentation-service.arm.com/static/5f8fef3af86e16515cdbf816) it seems that the T2 encoding is used, and we get 12 bits (4095) to use (I compared the bits from `f60f 70fc` with the ones from the pdf).

Going back to `L = 2042`. The reported value by the compiler is 0x101c, corresponding to 4124 in decimal. The calculation goes as follows, if I'm not mistaken: 4 bytes for `Hp = H;`, 6 bytes for `t = Hp[0]; v=0;`, 4084 bytes corresponding to the for loop (2 bytes per instruction), and 2 bytes for the last `pop`. Total of 4096 bytes (not considering the assembly code before `Hp = H;`)

Due to the `.p2align 5` directive before `glob_data:`, this seems to put `glob_data` at a distance of 4124 from the ADR instruction. Hence, this constant does not fit possible encodings for the instruction.

---
### Questions:
* Is it necessary to align global constants to 32 bytes? (I believe doing smaller alignments does not make a difference for this case, but it might help in other cases).
* How to deal with large programs / large sets of global variables/arrays in ARM? (I imagine that branching to a code section closer to the data section could work, but I'm unsure if this is the best solution for this problem).

---
### Some final notes regarding the SHA256 implementation:
* The initial ARM implementation of SHA256 was obtained by copy-pasting the amd64 implementation with some minor adjustments (large implementation; previously described problem observed);
* Since having a small implementation was the final goal, I wrote some temporary functions to initialize the corresponding constants directly into a stack array to have a working implementation;
* Refactored the ARM implementation to be "small", which made the previous error disappear. Nonetheless, this issue may appear in future implementations.

sarranz:
Temporary patch for #619. I assume that `ADR` instructions are only generated for accesses to the data section, which seems to be true. I don't think it's a robust solution, but I don't see how to do it otherwise. Also, I believe we can't tell whether we can perform the relative access or not, so as to generate two instructions only when needed. The other options discussed in the issue didn't work, and I can't explain why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,10 @@
 - Add support for the MMX instruction `POR`
   ([PR #917](https://github.com/jasmin-lang/jasmin/pull/917)).
 
+- ARM now emits two instructions instead of `ADR` to load the low and high
+  parts of global addresses
+  ([PR#921](https://github.com/jasmin-lang/jasmin/pull/921)).
+
 ## Bug fixes
 
 - Easycrypt extraction for CT : fix decreasing for loops
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -224,6 +224,23 @@ end = struct
     | _ -> ""
 end
 
+(* Split an [ADR] instruction to a global symbol into a [MOVW]/[MOVT] pair. *)
+let pp_ADR pp opts args =
+  let name_lo = pp_mnemonic_ext (ARM_op(MOV, opts)) "w" args in
+  let name_hi = pp_mnemonic_ext (ARM_op(MOVT, opts)) "" args in
+  let args =
+    List.filter_map (fun (_, a) -> pp_asm_arg a) pp.pp_aop_args
+  in
+  let args_lo, args_hi =
+    match args with
+    | dst :: addr :: rest ->
+        let lo = "#:lower16:" ^ addr in
+        let hi = "#:upper16:" ^ addr in
+        (dst :: lo :: rest, dst :: hi :: rest)
+    | _ -> assert false
+  in
+  [ LInstr(name_lo, args_lo); LInstr(name_hi, args_hi) ]
+
 let pp_instr fn i =
   match i with
   | ALIGN ->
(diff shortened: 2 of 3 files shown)


bgregoir:
I don't know what to think about this.
In one side it maybe is better than nothing. 
On the other this is not certified. 
Maybe you should merge this and then do the proper patch.
Any opinion on that ?

sarranz:
I did this because I thought we couldn't verify much about this, but you're right (as usual :)). I guess what you have in mind  is to define the MOV lower and upper as extra ops? If so, I can do it, no problem. But I can't assure when, so I leave the question about merging or waiting to you.

bgregoir:
Hi Santiago,
  have you been able to progress on this ?
B

sarranz:
No. I'm at a loss because I can reproduce certain behaviors only in some assemblers, in particular this method with `upper16` and `lower16`. I need to do a deep dive in the documentation next week when I have some time.

sarranz:
Ok, JC has helped me out. Apparently the disassembler was showing me incomplete information and this approach is fine. I will cleanup the pretty printer for the moment like we discussed and we can plan a cleaner patch next week.

sarranz:
Benjamin suggested we check that the argument is `Addr (Arip _)`, so now the assumption about issuing `ADR` only for global accesses is not necessary.

bgregoir (approved):

PR #926, opened 2024-10-08
CI/coq-master: use elpi 1.20.0

vbgl:
Compiling Coq-elpi master with elpi 1.18.2 has been failing with:

~~~
File "src/coq_elpi_builtins_synterp.ml", line 1220, characters 21-28:
1220 |       ~scope ~dbname clauses ~depth ~options state)),
                            ^^^^^^^
Error: This expression has type
         (string option *
          ([> `After | `Before | `Remove | `Replace ] * string) option *
          Elpi.API.Data.term)
         list
       but an expression was expected of type
         (string option * ([ `After | `Before | `Replace ] * string) option *
          Elpi.API.Data.term)
         list
       The second variant type does not allow tag(s) `Remove
~~~

Updating elpi to 1.20.0 works around this issue.

cc @gares (for your information).

### DIFF
--- a/scripts/coq-elpi.nix
+++ b/scripts/coq-elpi.nix
@@ -2,7 +2,7 @@
 
 let elpi =
   coq.ocamlPackages.elpi.override {
-    version = "v1.18.2";
+    version = "v1.20.0";
   }
 ; in
 


bgregoir (approved):

gares:
@vbgl, in theory it should also work with older versions *if* elpi correctly sets its version, see:

https://github.com/LPCIC/coq-elpi/pull/693/files#diff-7853acab41580965fbc8811e3fe0a83049c33a99c8892aabfa61ca848c4bbbc3

Since you had fixed that issue in nix, you should not have had this issue here.

vbgl:
Ok. Thanks for the explanation.

So updating `nixpkgs` might be a better fix. Sorry for the noise.

PR #934, opened 2024-10-18
Risc v constants if while

bgregoir:


compiler/src/glob_options.ml
```diff
@@ -140,6 +140,7 @@ let print_strings = function
   | Compiler.RegArrayExpansion           -> "arrexp"   , "expansion of register arrays"
   | Compiler.RemoveGlobal                -> "rmglobals", "remove globals variables"
   | Compiler.MakeRefArguments            -> "makeref"  , "add assignments before and after call to ensure that arguments and results are ref ptr"
+  | Compiler.PreLowering                 -> "prelowerign", "arch-specific transformations before lowering"
```
eponier, line 143:
```suggestion
  | Compiler.PreLowering                 -> "prelowering", "arch-specific transformations before lowering"
```

compiler/tests/fail/warning/load_constant_warning.jazz
```diff
@@ -0,0 +1,10 @@
+export fn foo(reg u32 x) -> reg u32 {
+
+  while (x < 10) {
+    x = x + 1;
+  }
+
+  x = x;
+  return x;
+
+}
```
eponier, line 10: I would move the file to a new subdir `risc-v`, because it is only on this arch that there is a warning.

proofs/compiler/arch_params.v
```diff
@@ -1,3 +1,4 @@
+Require Import Uint63.
```
eponier, line 1:
```suggestion
```
We don't need that line anymore.

proofs/compiler/arm_params_proof.v
```diff
@@ -937,7 +937,6 @@ Proof.
   exact: (word_uincl_zero_ext w' hws').
 Qed.
 
-
```
eponier, line 940: Please don't remove lines for no reason.

proofs/compiler/riscv_params.v
```diff
@@ -151,7 +151,6 @@ Definition riscv_liparams : linearization_params :=
 
 End LINEARIZATION.
 
-
```
eponier, line 154: Idem, don't remove this line. It is on purpose that we have two blank lines here.

proofs/compiler/riscv_params_proof.v
```diff
@@ -600,7 +601,6 @@ Proof.
   exact: (word_uincl_zero_ext w' hws').
 Qed.
 
-
```
eponier, line 603: Idem.

proofs/compiler/load_constants_in_cond_proof.v
```diff
@@ -0,0 +1,420 @@
...
+  t_xrbindP => z Hupdate_c Hwith_body.
+  subst x => /=.
+  have [||x Hevms2 Hsem] := (Hs2 _ _ Hupdate_c _ (evm s1)) => //; first by SvD.fsetdec.
+  rewrite with_vm_same in Hsem.
+  eapply EcallRun ; try by eassumption.
+  rewrite -Hvres -!(sem_pexprs_get_var _ (p_globs p)).
+  symmetry; move : Hevms2; rewrite -read_esE; apply : read_es_eq_on.
+Qed.
+
+Local Lemma Hsyscall : sem_Ind_syscall p Pi_r.
```
eponier, line 366: I'm picky, but please move this lemma between `Hopn` and `Hif_true`.

proofs/compiler/load_constants_in_cond_proof.v
```diff
@@ -0,0 +1,420 @@
...
+  case: (write_lvals_eq_on _ hw hvm1); first by SvD.fsetdec.
+  move=> vm2  ? eq_s2_vm2; exists vm2.
+  + by apply: (eq_onI _ eq_s2_vm2); SvD.fsetdec.
+  apply/sem_seq1/EmkI; constructor.
+  rewrite /sem_sopn Let_Let -(read_es_eq_on _ _ (s := X)); last first.
+  + by rewrite read_esE; apply: (eq_onI _ hvm1); SvD.fsetdec.
+  by rewrite Hsem_pexprs /= Hexec_sopn.
+Qed.
+
+(* FIXME move this lemma also in makereference_argument_proof *)
```
eponier, line 212: Please fix this FIXME before we merge.

eponier (changes requested):
I have a few minor comments.

Another question is whether the name of the pass "prelowering" is a good one. Maybe we should just call it "load constants in cond" with a comment explaining it is called only on RISC-V?

proofs/compiler/load_constants_in_cond_proof.v
bgregoir: done.

proofs/compiler/arm_params_proof.v
bgregoir: Done, but we lost our time with this.

proofs/compiler/load_constants_in_cond_proof.v
bgregoir: This is indeed important

proofs/compiler/arch_params.v
bgregoir: It is needed line 25.

proofs/compiler/load_constants_in_cond_proof.v
bgregoir: done

eponier:
I took the liberty to rename the pass ("prelowering" was unclear). If you agree with the changes I did, I'll merge.

eponier:
One side note: maybe it would make sense not to print the program after the pass when actually the pass is not called (on x86 and arm), but this is not so clear to me.

clebreto:
@bgregoir @eponier 

What's the status of this MR ?

bgregoir:
For me it is ok, we can merge it.
I agree that the printing can be removed is the pass is not used (but it is not important).

eponier:
Ok, I'm merging then.

PR #940, opened 2024-10-24
Renaming: report errors due to liveness conflicts

### LINKED ISSUE #483: Regalloc conflict produces internal error when function is inlined
The following gives `internal compilation error in function main: Variable allocation: i (defined at "ble.jazz", line 16 (12-13)) and i (defined at "ble.jazz", line 16 (12-13)) : variable already set`
```
fn f(reg ptr u8[1] p, reg u64 i) -> reg ptr u8[1] {
    while (i < 1) {
        p[(int) i] = 0;
        i += 1;
    }
    return p;
}

export
fn main () -> reg u8 {
    stack u8[1] c;
    reg ptr u8[1] p;
    p = c;

    reg u64 i;
    i = 0;

    while (i < 1) {
        #inline
        p = f(p, i);
        i += 1;
    }

    reg u8 tmp;
    tmp = p[0];
    return tmp;
}
```
I believe it's because `f` increases `i` but `main` does not see this. If I remove the `#inline`, then I get a register allocation error saying that these two `i`s conflict, this is what I'd expect. Naturally, it works if I introduce `i = i` at the beginning of `f`, or return `i` in `f`.
Is this really an internal error, then?

vbgl:
Fixes #483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
   ([PR #937](https://github.com/jasmin-lang/jasmin/pull/937);
   fixes [#895](https://github.com/jasmin-lang/jasmin/issues/895)).
 
+- Improve error message when renaming is not possible due to liveness conflicts
+  ([PR #940](https://github.com/jasmin-lang/jasmin/pull/940);
+  fixes [#483](https://github.com/jasmin-lang/jasmin/issues/483)).
+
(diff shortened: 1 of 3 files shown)


vbgl:
Bad try. This fails too often. There is no issue as long as there is no write to the variables… Let’s try again later.

PR #944, opened 2024-10-28
Modified toCL to be able to run add4 ref example and added the add4 ref example

JoaoDiogoDuarte:
This merge request adds:
- add4.jazz example with abstract predicates and asserts/assumes that will result in a valid and correct cryptoline file

And does the following to these abstracts predicates:
- Uncomments `pow` abstract predicate for algebraic expressions
- Adds algebraic abstract predicate `b2i` that converts a boolean to an integer
- Add algebraic abstract predicate `u64i` that converts a u64 to an integer
- Predicate that allows comparing two boolean variables (uses `Obeq`)

### DIFF
--- a/compiler/examples/jazzline/curve25519/common/add4/add4.jazz
+++ b/compiler/examples/jazzline/curve25519/common/add4/add4.jazz
@@ -1,0 +1,72 @@
+// h = f + g
+// h = 2**0*f0 + 2**64*f1 + 2**128*f2 + 2**192*f3 +
+//     2**0*g0 + 2**64*g1 + 2**128*g2 + 2**192*g3
+
+abstract predicate bool eqmod(int, int, tuple);
+abstract predicate tuple single(int);
+abstract predicate int b2i(bool);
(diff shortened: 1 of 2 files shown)


tfaoliveira:
Hey! I have a question regarding the following types of statements: 
* `\sum (ii \in 0:4) (pow(2, 64*ii)*u64i(result.0[ii]))` 

Why isn't it proposed an abstraction to reconstruct an array of limbs into an integer? 

I was thinking something like `limbs` from Cryptoline ([page 6 of the tutorial](https://github.com/fmlab-iis/cryptoline/blob/master/doc/tutorial.pdf)) for an example of what I'm referring to.  

If there were an limbs version in Jasmin that would take arrays, the slice of the array to be considered, and the radix, I think it would improve readability.

JoaoDiogoDuarte:
Since we support bits natively, we use the statement you mentioned so that we do not have to rely on Cryptoline's internal representation and of `limbs`.

PR #948, opened 2024-10-29
x86: disallow BT with memory operands

### LINKED ISSUE #931: Wrong semantics of x86 instruction BT when operand is in memory
The following (safe) program produces zero in Jasmin semantics but one when run on hardware after compilation.

~~~
export fn test() -> reg u8 {
  reg bool cf;
  stack u64[2] s;
  s[0] = 0;
  s[1] = 1;
  reg u64 ofs = 64;
  cf = #BT(s[0], ofs);
  reg u8 r = #SETcc(cf);
  return r;
}
~~~

vbgl:
Fixes #931.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
   ([PR #937](https://github.com/jasmin-lang/jasmin/pull/937);
   fixes [#895](https://github.com/jasmin-lang/jasmin/issues/895)).
 
+- Disable x86 `BT` instruction with in-memory operands
+  ([PR #948](https://github.com/jasmin-lang/jasmin/pull/948);
+  fixes [#931](https://github.com/jasmin-lang/jasmin/issues/931)).
+
 ## Other changes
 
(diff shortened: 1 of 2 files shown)


bgregoir:
This problem is not in the printing of the instruction ?
"In 64-bit mode, the instruction’s default operation size is 32 bits. Using a REX prefix in the form of REX.R permits access to additional registers (R8-R15). Using a REX prefix in the form of REX.W promotes operation to 64 bit operands. See the summary chart at the beginning of this section for encoding data and limits."
I think the documentation of this instruction with memory operand is anyway hard to understand.

vbgl:
No, the problem is not in printing. The problem is in the semantics.

If you run (pseudo-code):
~~~
x := 137
cf := BT(p, x)
~~~

this will set `cf` to the value of the bit at position 137 after `p`. This instruction can read a bit very far away from the given address.

bgregoir (approved):

PR #951, opened 2024-10-30
asm-gen: improve “invalid scale” message

### LINKED ISSUE #949: Print scale in "invalid scale" error
The following code
```
u256[1] a = { 0 };

export
fn main() -> reg u256 {
    reg ptr u256[1] p;
    p = a;

    reg u64 i;
    i = 0;

    reg u256 x;
    x = p[i];

    return x;
}
```
gives
`compilation error in function main: asmgen: invalid scale`
which is a bit confusing until you remember that `i` is being scaled by 32. If we access the array instead of a pointer, we get a beautiful `compilation error in function main: asmgen: Invalid global address : (RIP +64u ((((64u) 32) *64u RAX) +64u ((64u) 0)))` where we see the 32.
Can we print the invalid scale in the error?

vbgl:
Fixes #949.

Do not hesitate to suggest more improvements to the error message.

### DIFF
--- a/proofs/arch/asm_gen.v
+++ b/proofs/arch/asm_gen.v
@@ -175,7 +175,7 @@ Definition scale_of_z ii (z: Z) : cexec nat :=
   | 2%Z => ok 1
   | 4%Z => ok 2
   | 8%Z => ok 3
-  | _ => Error (E.error ii (pp_s "invalid scale"))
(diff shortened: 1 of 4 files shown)


sarranz:
I think it would be worth it printing the whole expression like we do for global addresses. Let me push a suggestion.

vbgl (commented):
Very nice. Thanks.

sarranz (approved):

PR #952, opened 2024-11-04
Array extraction: implement operations in eclib instead of generating anonymous functions

cassiersg:
@bgregoir @vbgl @lyonel2017

Progress on this one has been slow, but it's finally there.

(Needs recent easycrypt fixes: https://github.com/EasyCrypt/easycrypt/pull/653)


compiler/src/utils.mli
```diff
@@ -137,6 +137,10 @@ type model =
   | ConstantTime
   | Normal
 
+type amodel =
+  | ArrayOld
+  | ArrayEclib
+
```
vbgl, line 143: This belongs to `toEC.ml`.

compiler/tests/success/subarrays/x86-64/unaligned_cast.jazz
```diff
@@ -0,0 +1,28 @@
+
```
vbgl, line 1: This file has been added to the test-suite of the compiler (the tool that produces assembly). Yet, it is rejected by said compiler. Is that intended?

eclib/JWord_array.ec
```diff
@@ -250,8 +250,187 @@ abstract theory WArray.
...
+(* Particular case of SubArrayCast where both array's word types are the same.
+Still not a simple sub-array since indices may be a number of bytes
+non-multiple of the word size. *)
+abstract theory SubArrayDirect.
+  op sizeW: int.
+  axiom gt0_sizeW: 0 < sizeW.
+
+  clone WT as Word with
+    op size <- 8 * sizeW
+    proof gt0_size by rewrite pmulr_rgt0 1:// gt0_sizeWS.
```
vbgl, line 377: EasyCrypt complains:

> unknown lemma `gt0_sizeWS'

eclib/JWord_array.ec
```diff
@@ -250,8 +250,187 @@ abstract theory WArray.
...
+    proof gt0_r by apply gt0_sizeWS, sizeBrS by rewrite mulzC.
+
+
+  op get_cast_direct (a: WordB.t ArrayWordsB.ArrayN.t) (i: int) =
+    WSu8.pack'R_t (
+      WSu8.Pack.init (fun j => ArrayWordsB.WArrayN.get8 (ArrayWordsB.of_word_array a) (i+j))
+    ).
+
+  op set_cast_direct (a: WordB.t ArrayWordsB.ArrayN.t) (i: int) (b: WordS.t) =
+    (* FIXME: change init8 and get8 to init and _.[_] (breaks compatibility by rewrite) *)
```
vbgl, line 421: Shall this be fixed before the PR is merged? Otherwise please change to TODO.

vbgl (commented):
It's annoying to break compatibility with latest released version of EasyCrypt.

eclib/JWord_array.ec
cassiersg: The goal of not fixing this immediately was to allow to fix all proofs impacted by this change by simply `rewrite`'ing all new definitions. Here, the `rewrite` would need to go in the other direction, since get8 and init8 are just defined as _.[_] and init - or the proof could be fixed in a cleaner way (shouldn't be much harder in general).

Given that we have a backward-compatibility option, I'd tend to just make this additional change now.

compiler/tests/success/subarrays/x86-64/unaligned_cast.jazz
cassiersg: That was a bug.

cassiersg:
For depending on a released version of EasyCrypt, I don't see any alternative except waiting for a release.

@strub Do you have a schedule for the next EasyCrypt release?

vbgl:
I’ve just pushed a workaround regarding the unreleased bugfix in EasyCrypt.

eclib/JArray.ec
```diff
@@ -523,3 +523,28 @@ abstract theory PolyArray.
...
+abstract theory SubArray.
+  op sizeS: int.
+  axiom gt0_sizeS: 0 < sizeS.
+
+  op sizeB: int.
+  axiom gt0_sizeB: 0 < sizeB.
+
+  (* Sub-array *)
+  clone import PolyArray as ArrayS with
+    op size <- sizeS
```
bgregoir, line 537: I think op size <= sizeS will be better here. Because ArrayS will still be compatible with PolyArray.

eclib/JArray.ec
```diff
@@ -523,3 +523,28 @@ abstract theory PolyArray.
...
+    proof ge0_size by rewrite ltzW gt0_sizeS.
+
+  (* Base array *)
+  clone import PolyArray as ArrayB with
+    op size <- sizeB
```
bgregoir, line 542: same comment.

bgregoir (commented):
I need more time to read this.

eclib/JArray.ec
cassiersg: I don't have a good understanding of `<-` vs `<=` in theory cloning, but applying the suggested change leads to the following error:
```
[critical] [: line 1 (0) to line 7 (76)] In external theory SubArray2_3 [<unknown> location]:
unknown operator `Top.Array2.Array2.size'
```

in the following code (generated by jasmin2ec):
```
from Jasmin require import JArray.

require import Array2 Array3.

clone export SubArray as SubArray2_3  with op sizeS <- 2, op sizeB <- 3,
                                           theory ArrayS <= Array2,
                                           theory ArrayB <= Array3.
```

where the Array2 theory is defined as
```
from Jasmin require import JArray.

clone export PolyArray as Array2  with op size <- 2.
```

Maybe the problem comes from the way the `SubArray` or `Array` theories are cloned, I don't really understand how `<-` vs `<=` work. I found the current way to work after trial-and-error...

Besides what you can read in the patches for `JArray.ec` and `JWord_array.ec`, the code generated by `jasmin2ec` does the following (see [these functions](https://github.com/jasmin-lang/jasmin/blob/379462c2a854c613158c47dc31b1a06e6be4f6ba/compiler/src/toEC.ml#L714-L783)):

- `<-` for the array size & word size parameters (function `fmt_op`)
- `<-` for the "word" (W8/W16/...) theory instances (function `fmt_the`)
- `<=` for the other theories  (function `fmt_th`)

I'm open to any other solution (or explanation on what would/wouldn't work).

bgregoir (approved):

eclib/JArray.ec
bgregoir: The semantic is the following
clone T as T'
   x <- t     :    x will be substituted by t everywhere in T', and its definition is removed so T'.x does not exists
   x = t        : T'.x is defined to be t, all occurences of x in T' is unchanged.
   x <= t     : T'.x is defined to be t,  x will be substituted by t everywhere in T'.

PR #955, opened 2024-11-05
x86: improve description of AVX shift instructions

vbgl:
- VPSRA only supports 16 or 32 bit elements
- count argument can be in xmm register or in memory


vbgl:
I’d like to use `W128.t` instead of `W8.t` there: https://github.com/jasmin-lang/jasmin/blob/62b1f0ac2decd23285f045d79998f7b83d007492/eclib/JWord.ec#L2304

Has anyone got a smart suggestion?

vbgl:
I’ve got something working, on top of #957.

bgregoir (approved):

PR #958, opened 2024-11-08
fix #637

### LINKED ISSUE #637: Wrong lval order for UMULL
The lval order should be lower then higher, like the assembly.

bgregoir:


vbgl:
Ping Vincent (@vincentvbh) in case he has got some feedback.

vincentvbh:
Thank you for making the changes. I'll immediately notify people I know that are using umull and smull.

vbgl:
I have a doubt about the change for the signed multiplication. Can someone please double-check?

vbgl:
Because when I compare execution on hardware and in the semantics, it used to be OK and now output words are swapped.

bgregoir:
For me I think, it is ok. And that it is not incompatible with the fact that you checked hardware with
software before.
Before this commit 
  (hi, lo) = SMULL(x, y)    ----> SMULL lo hi x y     semantics (highbit (x * y ), lowbit (x * y))
  (hi, lo) = UMULL(x, y)    ----> UMULL lo hi x y
after : 
  (lo, hi) = SMULL(x, y)    ----> SMULL lo hi x y      semantics (lowbit(x * y), highbit (x * y))

In all cases (before/after and source/assembly)  you have lo <- lowbit(x * y) and hi <- highbit (x * y)
Assuming lo <> hi.

vbgl:
Fausse alerte. Sorry for the noise.

vbgl (approved):

PR #960, opened 2024-11-15
DIV + REM

eponier:
Also:
+ reorder riscv_instr_decl.v for consistency
+ remove GP/TP from the formalization (according to the RISC-V ABI, they are "unallocatable")


eponier:
I rebased on top of latest risc-v. Extraction to EC must be updated to take into account DIV and REM, but I don't know how to do that.

clebreto:
@bgregoir We are in need of a double check regarding the EasyCrypt extraction of DIV and DIVU =)

bgregoir (approved):

PR #961, opened 2024-11-15
Add nowarning option to jasmin-ct, jasmin2tex, and jasmin2ec

sarranz:
There is no way of suppressing these warning in these tools. I also disabled the warnings for SCT checker tests. The tests under CCT don't need this since the warnings are already caught somewhere else.

### DIFF
--- a/compiler/tests/sct-checker/common.ml
+++ b/compiler/tests/sct-checker/common.ml
@@ -9,7 +9,6 @@ module Arch =
        (module Arch_full.Arch_from_Core_arch (C) : Arch_full.Arch))
 
 let load_file name =
-  Utils.nowarning ();
   let open Pretyping in
   try
     name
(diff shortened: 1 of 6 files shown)


sarranz:
That's a good point, thanks

bgregoir:
Hi, I agree that printing the warning for all those tools are not relevant. 
I think it is good to have a nowarning option (also in jasminc).
But I think it will be better that jasmin-ct, jasmin2tex jasmin2ec use nowarning has the default value
while jasminc not.

sarranz:
Good idea, done

compiler/tests/sct-checker/common.ml
```diff
@@ -9,6 +9,7 @@ module Arch =
        (module Arch_full.Arch_from_Core_arch (C) : Arch_full.Arch))
 
 let load_file name =
+  Utils.nowarning ();
```
vbgl, line 12: This looks like a bad idea to me. When the warning about unused local variables was introduced, it helped improving the test suite.

sarranz: Fixed

vbgl (approved):

PR #963, opened 2024-11-21
Fix printing of underscores in jasmin2tex

sarranz:
Underscores need a backslash otherwise they print as a dot.

### DIFF
--- a/compiler/src/latex_printer.ml
+++ b/compiler/src/latex_printer.ml
@@ -291,7 +291,7 @@ let pp_inbraces depth p fmt x =
 
 let pp_lv fmt x =
   match L.unloc x with
-  | PLIgnore -> F.fprintf fmt "_"
+  | PLIgnore -> F.fprintf fmt "\\_"
   | PLVar x -> pp_var fmt x
   | PLArray (al, aa, ws, x, e, len) -> pp_arr_access fmt al aa ws x e len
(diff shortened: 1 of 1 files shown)


vbgl:
Are you sure?

sarranz:
If I write
```
export
fn my_main_function() {
  _ = 0;
}
```
I get
```
\jasminkw{export} \jasminkw{fn} \jasmindname{my_main_function}() \{\\
\jasminindent{1}_ = 0;\\
\}\\
```
So then I compile
```
\documentclass{article}

\usepackage{jasmin}

\begin{document}

\begin{jasmincode}
\jasminkw{export} \jasminkw{fn} \jasmindname{my_main_function}() \{\\
\jasminindent{1}_ = 0;\\
\}\\
\end{jasmincode}

\end{document}
```
and get
Screenshot from 2024-11-21 10-19-35

sarranz:
It might be a problem of my latex installation... Otherwise I'll take a look at the ones in identifiers as well

vbgl:
What happens if you `\usepackage[T1]{fontenc}`?

sarranz:
That fixes it, sorry for the noise. I will add this require to my jasmin.sty then.

vbgl:
> I will add this require to my jasmin.sty then.

This looks like a bad idea: it will mess up any document that uses fonts using different encodings.

PR #965, opened 2024-11-25
Use liveness info to speed up allocation

eponier:
The tables used in allocation can be huge. We remove the dead variables from the tables, to reduce their sizes and speed up the pass. Inspired by CompCert that does something similar to reduce the cost of the value analysis.

Please review carefully the changes to Liveness, because I don't know what I did.

On [this implementation of Keccak](https://github.com/formosa-crypto/formosa-mldsa/tree/01ec03dd53e0fe43abf8c77091fd178560017ce6/src/libjade/keccakf1600), the time taken by the analysis after inlining goes down from 20s to 1.5s. It still feels too much for 4 inlined functions (size 1700, 3800 variables). (More functions are inlined, but these are the ones where the time is spent).

### DIFF
--- a/compiler/src/liveness.ml
+++ b/compiler/src/liveness.ml
@@ -50,7 +50,14 @@ and live_d weak d (s_o: Sv.t) =
     let s2, c2 = live_c weak c2 s_o in
     Sv.union (vars_e e) (Sv.union s1 s2), s_o, Cif(e, c1, c2)
 
-  | Cfor _ -> assert false (* Should have been removed before *)
+  | Cfor (x, (_dir, e1, e2 as r), c) ->
+    let rec loop s_o =
+      let s_i, c = live_c weak c s_o in
+      let s_i = Sv.remove (L.unloc x) s_i in
+      if Sv.subset s_i s_o then s_o, c
+      else loop (Sv.union s_i s_o) in
+    let s_i, c = loop s_o in
+    Sv.union (vars_es [ e1; e2 ]) s_i, s_o, Cfor (x, r, c)
 
   | Cwhile(a,c,e,c') ->
     let ve = (vars_e e) in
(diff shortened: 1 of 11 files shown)


compiler/src/liveness.ml
```diff
@@ -50,7 +50,14 @@ and live_d weak d (s_o: Sv.t) =
     let s2, c2 = live_c weak c2 s_o in
     Sv.union (vars_e e) (Sv.union s1 s2), s_o, Cif(e, c1, c2)
 
-  | Cfor _ -> assert false (* Should have been removed before *)
+  | Cfor (x, r, c) ->
+    let rec loop s_o =
+      let s_o' = if weak then Sv.add (L.unloc x) s_o else s_o in
+      let s_i, c = live_c weak c s_o' in
```
vbgl, line 56: This looks wrong. `s_o'` should be `s_o` and `x` should be *removed* from `s_i`.

vbgl (commented):
This change might deserve an entry in the changelog.

compiler/src/liveness.ml
eponier: Indeed, this was certainly wrong. Is this better?

eponier:
> This change might deserve an entry in the changelog.

I added an item.

compiler/src/liveness.ml
eponier: libjade fails, so I guess it's wrong.

compiler/src/liveness.ml
```diff
@@ -50,7 +50,14 @@ and live_d weak d (s_o: Sv.t) =
...
 
-  | Cfor _ -> assert false (* Should have been removed before *)
+  | Cfor (x, r, c) ->
+    let rec loop s_o =
+      let s_i, c = live_c weak c s_o in
+      let s_i = Sv.remove (L.unloc x) s_i in
+      if Sv.subset s_i s_o then s_i, c
+      else loop (Sv.union s_i s_o) in
+    let s_i, c = loop s_o in
+    s_i, (if weak then Sv.add (L.unloc x) s_o else s_o), Cfor (x, r, c)
```
vbgl, line 60: Why do you look at `weak`?

The variables occurring in the range are live before the `for` instruction.

vbgl (approved):

eponier:
Thanks!

PR #967, opened 2024-11-27
toEC: make leakage part of functions return instead of a global variable

cassiersg:


compiler/src/toEC.ml
```diff
@@ -368,34 +371,28 @@ let add_funcs env fds =
     let s = mkname env fd.f_name.fn_name in
     let funs = 
       Mf.add fd.f_name (s, ((*mk_tys*) fd.f_tyout, (*mk_tys*)fd.f_tyin)) env.funs in
-    { env with funs; alls = Ss.add s env.alls } in
+    { env with funs; alls = ref (Ss.add s !(env.alls)) } in
```
bgregoir, line 461: do you really want to create a new ref or simply update 
env.alls ?

bgregoir (approved):
I let you fix the typo

compiler/src/toEC.ml
cassiersg: We only need to update env.alls, but having a new ref also works.

cassiersg:
@vbgl The gimli proof checks fail with the new extraction due to https://github.com/EasyCrypt/easycrypt/issues/666, which will soon be fixed in easycrypt `main` (https://github.com/EasyCrypt/easycrypt/pull/667).

Another way to fix the proof failure is to revert that test to the previous way of extracting leakage with `--model CTG`. This can perhaps be done only for the easycrypt release CI, so that we don't have to wait for an easycrypt release, but we still test the new code ?

vbgl:
If you can adapt the CI & gimli proof makefile to use CT model with easycrypt-main and CTG model with easycrypt-release, it would be perfect.

If this is not possible, or too hard to implement, we’ll figure out something else.

cassiersg:
I tried that, let's see if it works.

cassiersg:
No, I forgot to disable my editor's yaml auto-formatter when saving. I'll fix that.

vbgl (approved):
I did not check toEC.ml in details.

PR #971, opened 2024-12-05
Fix EC extraction in case on nested loops

lyonel2017:
The EC extraction is wrong In the case of nested loop. Consider following Jasmin program:
```
param int a = 10;

fn __for_nest() -> ()
{
  inline int i;
  inline int j;

  for i = 0 to (a + a)
  {
     for j = 0 to (a * a){}
  }

  return;
}
```
The result is currently :
```
require import AllCore IntDiv CoreMap List Distr.

from Jasmin require import JModel_x86.

import SLH64.

module M = {
  proc __for_nest () : unit = {
    var int:aux;
    var j:int;
    var i:int;
    aux <- (10 + 10);
    i <- 0;
    while ((i < aux)) {
      aux <- (10 * 10);
      j <- 0;
      while ((j < aux)) {
        j <- (j + 1);
      }
      i <- (i + 1);
    }
    return ();
  }
}.
```
It is not correct to use `aux` for the second loop.
This PR fix this. The result for the previous Jasmin example is now :
```
require import AllCore IntDiv CoreMap List Distr.

from Jasmin require import JModel_x86.

import SLH64.

module M = {
  proc __for_nest () : unit = {
    var int:inc;
    var int:inc_0;
    var j:int;
    var i:int;
    inc <- (10 + 10);
    i <- 0;
    while ((i < inc)) {
      inc_0 <- (10 * 10);
      j <- 0;
      while ((j < inc_0)) {
        j <- (j + 1);
      }
      i <- (i + 1);
    }
    return ();
  }
}.
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,6 +19,7 @@ Ján Jančár
 Jean-Christophe Léchenet
 José Bacelar Almeida
 Kai-Chun Ning
+Lionel Blatter
 Lucas Tabary-Maujean
 Manuel Barbosa
 Miguel Quaresma
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 - Fix EasyCrypt semantics of shift operators
   ([PR#973](https://github.com/jasmin-lang/jasmin/pull/973)).
 
+- Fix EC extraction in case on nested loops
+  ([PR #971](https://github.com/jasmin-lang/jasmin/pull/971)).
+
 ## Other changes
 
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -377,6 +377,8 @@ type env = {
     array_theories: Sarraytheory.t ref;
     (* aux variables: intermediate in extraction of jasmin assignments *)
     auxv  : string list Mty.t ref;
+    (* counter for aux variables *)
+    count : int Mty.t ref;
     randombytes : Sint.t ref;
     nesting: int;
   }
@@ -449,6 +451,7 @@ let empty_env arch pd model array_theories randombytes =
     funs = Mf.empty;
     array_theories;
     auxv  = ref Mty.empty;
+    count = ref Mty.empty;
     randombytes;
     nesting = 0;
   }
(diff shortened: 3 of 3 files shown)


vbgl:
Please add a changelog entry, and, if possible, a non-regression test-case.

lyonel2017:
> Please add a changelog entry, and, if possible, a non-regression test-case.

I added a changelog entry, and a non-regression test-case.

cassiersg:
Regarding implementation strategy for this fix, I have something somewhat similar in #967 . 
(See https://github.com/jasmin-lang/jasmin/pull/967/commits/9f42f2cf39f47f10770f42e5d9d74864c63ac05c#diff-039c0771a2cb0a5a50ae2c100a8cf26e1a8880c59122bf613c250e745d0b3d09R468 and https://github.com/jasmin-lang/jasmin/pull/967/commits/9f42f2cf39f47f10770f42e5d9d74864c63ac05c#diff-039c0771a2cb0a5a50ae2c100a8cf26e1a8880c59122bf613c250e745d0b3d09R1385 .)

Would be good to have a single mechanism.

vbgl:
Please decide the fate of this PR (is it subsumed by 967 or is it still worth on its own). If this PR stays, then please clean the history and rebase on top of recent `main` (in particular put the CHANGELOG message in the existing “bug fixes” section).

vbgl:
My opinion is that the bug (present in the current release) should be fixed on its own.

cassiersg:
With @lyonel2017, we concluded that it would be easier to rebase this PR on top of #976, which I'll rebase and finalize today.

cassiersg (approved):
This should fix the bug, but adds a bit of technical debt since we now have two mechanisms for creating aux variables, they should ideally be unified. (The previous one is a bit hacky, while the new one does not allow re-use of aux variables at all.)

lyonel2017:
> we now have two mechanisms for creating aux variables, they should ideally be unified. (The previous one is a bit hacky, while the new one does not allow re-use of aux variables at all.)

These points should now be fixed

eponier:
Regardless of the content of the PR, please add yourself to the AUTHORS file.

vbgl:
And also please address the issue with the changelog that I mentioned earlier.

lyonel2017:
> Regardless of the content of the PR, please add yourself to the AUTHORS file.

> And also please address the issue with the changelog that I mentioned earlier.

These two points should now be ok.

cassiersg (approved):
Looks good to me.

While trying to clean up my changes w.r.t. your improvements, I wrote a follow-up commit that improves the handling of leakage aux variables, and also removes the `pop_aux`/`reset_aux` functions (these seem likely to lead to bugs if forgotten) with a new function that creates a new env when entering an operation that may use new aux vars (that requires aux var usage to be properly nested, which currently holds, and seems reasonable):

https://github.com/jasmin-lang/jasmin/commit/a42ec46394f5662654736cc5512c5b252daf3347#diff-039c0771a2cb0a5a50ae2c100a8cf26e1a8880c59122bf613c250e745d0b3d09R377

It's probably better to merge this PR as-is and that I open a new PR to discuss my changes, though.

cassiersg:
@vbgl  I think this is ready to merge. (I'm not sure if this ping is useful to you or not - let me know for next time.)

PR #976, opened 2024-12-06
Add message in the result of jasmin2tex about location of .sty file

lyonel2017:
Add to the Jasmin project the `jasmin.sty` file needed when using the result of the `jasmin2tex` extraction.

### DIFF
--- a/compiler/src/latex_printer.ml
+++ b/compiler/src/latex_printer.ml
@@ -447,5 +447,14 @@ let rec pp_pitem fmt pi =
      closebrace fmt ()
   | PTypeAlias (id,ty) -> pp_typealias fmt id ty (**)
 
+let pp_info fmt =
+  F.fprintf fmt "@[<v>@[%% The produced LATEX snippet is meant to be included in a@] @ ";
+  F.fprintf fmt "@[%% jasmincode environment provided by the jasmin package@] @ ";
+  F.fprintf fmt "@[%% defined in file: @] @ ";
+  F.fprintf fmt "@[%% https://github.com/jasmin-lang/jasmin/wiki/resources/jasmin.sty@]@]";
+  F.pp_print_newline fmt ();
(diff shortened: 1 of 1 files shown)


vbgl:
This file is already in the wiki repository. Why do you think it belongs here?

vbgl:
This file looks out-of-date.

lyonel2017:
> This file is already in the wiki repository. 

I was not aware of this.

> Why do you think it belongs here?

This is is the place I would have started looking for, since `jasmin2tex` is in the project.

Maybe a better reason is that the file would be synchronized with the development of the `Jasmin` project and not the wiki.

lyonel2017:
> This file looks out-of-date.

Indeed.

compiler/src/latex_printer.ml
```diff
@@ -447,5 +447,14 @@ let rec pp_pitem fmt pi =
      closebrace fmt ()
   | PTypeAlias (id,ty) -> pp_typealias fmt id ty (**)
 
+let pp_info fmt =
+  F.fprintf fmt "@[<v>@[The produced LATEX snippet is meant to be included in a@] @ ";
+  F.fprintf fmt "@[jasmincode environment provided by the jasmin package@] @ ";
+  F.fprintf fmt "@[defined in file: @] @ ";
+  F.fprintf fmt "@[https://github.com/jasmin-lang/jasmin/wiki/resources/jasmin.sty@]@]";
+  F.pp_print_newline fmt ();
+  F.pp_print_newline fmt ()
```
vbgl, line 456: Are these TEX comment?

lyonel2017: No. Since the output is not valid Tex (initially), I thought this is not required.

lyonel2017: I add TeX comments.

vbgl (approved):

PR #977, opened 2024-12-08
subst: small clean-up

vbgl:
Remove redundant generalization over a constant value


eponier:
Wrong, not worth, or I was just too slow to merge it?

vbgl:
What is currently implemented is correct and complex for a reason. The simplification proposed in this PR is probably not incorrect because unique identifiers are globally unique but it makes the substitution table larger (a single one for all functions) instead of resetting it to its initial state before handling any new function.

eponier:
Thanks for the explanation!

PR #979, opened 2024-12-08
Check no for loop in coq

vbgl:
For loops must be gone after unrolling. The compiler relies on this fact. The no-for-loop check makes sure that if this is not the case, the user sees a proper error message.

Currently, this check is only run in the `jasminc` front-end. This PR moves this check to the compiler itself, where it belongs.


eponier (approved):

eponier (approved):
I've only rebased to avoid some conflict.

PR #985, opened 2024-12-11
Make _CoqProject compatible with ProofGeneral

vbgl:
Please someone double-checks this also works with CoqIDE.


eponier:
It seems to work on CoqIDE + VsCoq on Linux.

PR #991, opened 2024-12-13
Update warnings

eponier:
Some do not exist any more, some appeared.

I had too much time, it seems.


vbgl:
This one we may want to disable as well:
[deprecated-from-Coq,deprecated-since-9.0,deprecated,default].

eponier:
Oh. Maybe I should test coq 8.20 also, to verify there is no other warnings.

vbgl:
Right, and it looks serious (i.e., fix the notations rather than turning off the warning).

eponier:
> Right, and it looks serious (i.e., fix the notations rather than turning off the warning).
I fixed it, but I could not fix the one about "postfix notations should be at level 1" because we inherit it from ssreflect.

eponier:
> There is also a warning for inductives declared with annotation `: Type` but that were silently put in `Prop` until now. The fix is either to remove the annotation, and they will stay in `Prop`, or to disable the warning and they will be in `Type` instead. This happens with type `empty` in files `arch/arch_utils.v` and with `h_sh_params` in `compiler/slh_lowering_proof.v`. The second one should really go in `Prop`, but for the first one, I don't know what to do. What is the best move?

I removed the annotation also in the case of `empty`, meaning it will stay in `Prop` as before. I just picked the easiest change, no particular reasoning behing.

eponier:
To prepare for 9.0, I added `From Coq` everywhere it was needed. There could be some polishing to do by making the imports more uniform across the Jasmin codebase. I leave it for another day, if I'm motivated again for this kind of really creative task.

bgregoir:
"I removed the annotation also in the case of empty, meaning it will stay in Prop as before. I just picked the easiest change, no particular reasoning behind."
Does it work if we use Set instead?

eponier:
> "I removed the annotation also in the case of empty, meaning it will stay in Prop as before. I just picked the easiest change, no particular reasoning behind." Does it work if we use Set instead?

Same behaviour and no warning whatever the version of Coq is, so it seems a good solution. I made the change.

eponier:
I mean, not the same behaviour as before, but with `: Set` the behaviour is the same with all Coq versions.

eponier:
Actually, the change to `empty` change the extraction. For the better, I think, since instead of the cryptic `__`, we need to use `Arch_utils.empty`. This should now work.

eponier:
Green and ready!

PR #999, opened 2024-12-18
Ec extraction: cleanup leak aux variables

cassiersg:
Follow-up to #971 

Improve the handling of leakage aux variables, and also removes the `pop_aux/reset_aux` functions (these seem likely to lead to bugs if forgotten), and replace them with a new function that creates a new env when entering an operation that may use new aux vars (this assumes aux var usage to be properly nested, which currently holds, and seems reasonable).

### DIFF
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -412,7 +412,7 @@ module Env: EnvT = struct
         - auxv: for each (prefix, type), the list of all aux (used for variable declaration).
         - count: number of currently live aux variables for each (prefix, type).
         *)
-      auxv: string list Mpty.t ref;
+      auxv: string BatVect.t Mpty.t ref;
       mutable count: int Mpty.t;
       randombytes: Sint.t ref;
     }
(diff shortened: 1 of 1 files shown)


compiler/src/toEC.ml
```diff
@@ -468,39 +470,22 @@ let get_funtype env f = snd (Mf.find f env.funs)
...
-  if 0 < i then
-    env.count := Mty.add (prefix,ty) (i-1) !(env.count)
-
-let reset_aux env prefix tys =
-  let do1 ty =
-    env.count:= Mty.add (prefix, ty) 0 !(env.count)
-  in
-  List.iter do1 tys
+  let l = try Mty.find (prefix, ty) !(env.auxv) with Not_found -> [] in
+  env.count := Mty.add (prefix,ty) (i+1) !(env.count);
```
lyonel2017, line 476: I think increasing the counter of the aux variables should be done only when `fresh` is `true`.

lyonel2017 (changes requested):

compiler/src/toEC.ml
cassiersg: indeed

cassiersg:
@bgregoir All `toEC` fixes we discussed are implemented.

compiler/src/toEC.ml
```diff
@@ -357,156 +357,206 @@ type ec_prog = ec_item list
...
 
-let pop_aux env prefix ty =
-  let i = try Mty.find (prefix, ty) !(env.count) with Not_found -> 0 in
-  if 0 < i then
-    env.count := Mty.add (prefix,ty) (i-1) !(env.count)
+  (* Return the last created aux for (prefix, ty) in this env. *)
+  let reuse_aux env prefix ty =
+    let i = Mpty.find (prefix, ty) env.count in
+    let l = Mpty.find (prefix, ty) !(env.auxv) in
+    BatVect.get l (i-1)
```
lyonel2017, line 544: It might be interesting to add a test for the size of the list `l`, and if the list is empty a call to `create_aux`.

lyonel2017 (changes requested):

compiler/src/toEC.ml
cassiersg: It('s intentionally not done: if `reuse_aux` is called is called and `create_aux` has not been called before, then it's a bug. It's then better to crash than to (probably) generate incorrect output.

bgregoir (approved):

compiler/src/toEC.ml
bgregoir: I am agree with Gaetan

bgregoir: Lyonel are you agree to remove this change request ?

PR #1014, opened 2025-01-22
SCT checker: kill dead vars

vbgl:
There have been reports of performance issues with the SCT checker.

This change is meant to improve run-time performance by dropping internal information about dead variables.

The single data point I have supporting this change is that checking a particular implementation of ML-KEM for DOIT takes 30mn before this change and 5s after.

### DIFF
--- a/compiler/src/sct_checker_forward.ml
+++ b/compiler/src/sct_checker_forward.ml
@@ -409,6 +409,9 @@ module Env : sig
 
   val get_resulting_corruption : venv -> VlPairs.t
 
+  val venv_forget : Sv.t -> venv -> venv
+  (** Restrict knowledge to the given set of variables. *)
+
 end = struct
 
   type env = {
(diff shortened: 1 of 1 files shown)


compiler/src/sct_checker_forward.ml
```diff
@@ -953,7 +961,12 @@ let move_msf ~loc env (msf, venv) mso msi =
 (* --------------------------------------------------------------- *)
 (* [ty_instr env msf i] return msf' such that env, msf |- i : msf' *)
 
-let rec ty_instr is_ct_asm fenv env ((msf,venv) as msf_e :msf_e) i =
+let rec ty_instr is_ct_asm fenv env (msf, venv) i =
+  let venv = Env.venv_forget (fst i.i_info) venv in
+  let msf, venv = ty_instr_r is_ct_asm fenv env (msf, venv) i in
+  msf, Env.venv_forget (snd i.i_info) venv
```
bgregoir, line 967: Why forget is done after. I have the impression that it will be done again at the beginning of the next instruction. But maybe it is not really important.

bgregoir (approved):

compiler/src/sct_checker_forward.ml
vbgl: If you don’t do it, there are assertion failures in the test suite.

Sometimes, the next step is not the analysis of an instruction but a merge, e.g., in a `for` loop.

bgregoir: Ok, let's merge it like that.

PR #1015, opened 2025-01-22
updating printer `pp_iprog` function to print Cwhile Info

MrDaiki:
[PR#969](https://github.com/jasmin-lang/jasmin/pull/969) introduced a bug in `pp_iprog` function which was not priting added info in CWhile. This PR fix it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@
 - While instructions now have two pieces of information (including locations)
   attached to them: one corresponding to the whole instruction, the other one
   corresponding to the guard condition
-  ([PR #969](https://github.com/jasmin-lang/jasmin/pull/969);
+  ([PR #969](https://github.com/jasmin-lang/jasmin/pull/969),
+  [PR #1015](https://github.com/jasmin-lang/jasmin/pull/1015);
   fixes [#902](https://github.com/jasmin-lang/jasmin/issues/902)).
 
 - Add support for x86 `VPABS` instructions
--- a/compiler/src/printer.ml
+++ b/compiler/src/printer.ml
@@ -168,20 +168,26 @@ let rec pp_gi pp_info pp_len pp_opn pp_var fmt i =
       (pp_gvar_i pp_var) i (pp_ge pp_len pp_var) e1 dir (pp_ge pp_len pp_var) e2
       (pp_cblock pp_info pp_len pp_opn pp_var) c
 
-  | Cwhile(a, [], e, _, c) ->
-    F.fprintf fmt "@[<v>%awhile (%a) %a@]"
+  | Cwhile(a, [], e, ((iloc, _), info), c) ->
+    F.fprintf fmt "@[<v>%awhile %a(%a) %a@]"
       pp_align a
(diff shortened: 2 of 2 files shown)


compiler/src/printer.ml
```diff
@@ -168,19 +168,22 @@ let rec pp_gi pp_info pp_len pp_opn pp_var fmt i =
...
+  | Cwhile(a, c, e, ((iloc,_),info), []) ->
+    F.fprintf fmt "@[<v>%a%awhile %a (%a)@]"
       pp_align a
+      pp_info (iloc, info)
       (pp_cblock pp_info pp_len pp_opn pp_var) c (pp_ge pp_len pp_var) e
 
-  | Cwhile(a, c, e, _, c') ->
-    F.fprintf fmt "@[<v>%awhile %a (%a) %a@]"
+  | Cwhile(a, c, e, ((iloc,_),info), c') ->
+    F.fprintf fmt "@[<v>%a%awhile %a (%a) %a@]"
```
vbgl, line 184: It looks strange to print this information before the loop (although the location refers to the condition): it will be hard to distinguish it from the info attached to the whole loop.

MrDaiki: I wasn't sure to where I should put it, I can move it after the condition if you think that's better

vbgl: Let’s consult an Oracle.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@
   process
   ([PR#989](https://github.com/jasmin-lang/jasmin/pull/989)).
 
+- Info introduced in CWhile node in [PR#969](https://github.com/jasmin-lang/jasmin/pull/969) is not printed while using `pp_iprog`
+  ([PR#1015](https://github.com/jasmin-lang/jasmin/pull/1015))
+
```
vbgl, line 54: I think you should just amend the message for PR 969 above.

MrDaiki: I wasn't sure how to handle this case, thank you for your input, will be done shortly

compiler/src/printer.ml
MrDaiki: I made a correction to the display so info appears before the while condition

eponier:
I agree with @vbgl that it's probably best to print it close to the condition, where it belongs. I wanted to look at an example to get an intuition about where it would be best, but it seems we never print it in practice, do we? We define `pp_ifunc` and `pp_iprog`, but they are never called. `pp_info` is defined in Liveness, but never called either.

eponier:
I remembered that we have option `-pliveness`. With it, I can see the impact of this patch. Currently, the info is printed after the condition, not before. And I think the info is valid for before the condition, so it must really be printed before.

vbgl:
`-pliveness` does not use `pp_iprog`.

eponier:
No, indeed, but it is impacted by this PR.

eponier:
It also has an impact on the printing done by `-pall`. What I suggest for this patch:
- ensure this does not make `-pall` harder to read (ideally, it would even have no impact on `-pall`)
- print the new piece of info before the while condition
- check on `-pliveness` that the changes are ok

vbgl (approved):
Rebased and launched CI.

PR #1019, opened 2025-01-23
CI: update nixpkgs

vbgl:
This fixes the `coq-master` job.


eponier:
For my comprehension, before your patch, coq-elpi master was compiled with the last release of elpi, and now with the master version of elpi? Is this correct?

eponier (approved):
I feel I start understanding a bit these nixeries!

PR #1020, opened 2025-01-24
eclib: add [rigid] to hints on 'modulus'

bacelar:
This PR adds a `[rigid]` attribute to hints dealing with `modulus` on the `BitWord` theory. This is a new feature added to EC in https://github.com/EasyCrypt/easycrypt/commit/5f7be80cc247e5ad91606e139b9b20c30be91d3c (although not yet available in a released version), which solves an efficiency problem when instantiating the theory `BitWord` with big sizes (e.g., 512 bits).


jba-uminho:
Its now in sync with the current `main` branch.

Issue #727, opened 2024-02-13
Well-typed program rejected by the type-checker


vbgl:
The following program is well-typed, yet, it is rejected by the pre-typing phase (since version 2022.04).

~~~
fn id(reg u64 x) -> reg u64 { return x; }

inline
fn test() -> reg u64 {
  reg u64 r;
  r = id(42);
  return r;
}
~~~

MrDaiki:
Looking at this issue, I am not sure I agree with you that the program is "well typed". In a sense, I understand that since 42 could be store in a `reg u64`, there is no type error existing in the program. However, this add an implicit instruction : 
```
reg u64 temp = 42;
r = id(temp);
```

Do we really want to make this construction implicit ? A solution would be to raise an error after pre-typing, in a distinct compilation pass. What we need to know in that case is until which pass we wan't to be able to input this program (maybe until the pass where the program can be translated to Easycrypt, I am not familiar with this process)

PR #1023, opened 2025-01-24
Changing some prettyping errors to be able to ignore them (preparation to fix bug #727)

MrDaiki:
# Issue 

Some well typed programs are rejecting by pre-typing because they don't have a valid assembly target (see [#727](https://github.com/jasmin-lang/jasmin/issues/727))

# Solution 

I introduced a global boolean variable (I know it isn't the cleanest but it fit well in this case) that define if some of pre-typing error should be handled as warning or as error. It can be set with the `Utils.set_warn_recoverable` function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
   therefore old intrinsic `#VPBLENDVB` is deprecated
   ([PR #1010](https://github.com/jasmin-lang/jasmin/pull/1010)).
 
+- Add an option to treat some pre-typing error as warning instead. 
+  ([PR #1023](https://github.com/jasmin-lang/jasmin/pull/1023))
+
 ## Bug fixes
 
 - Fix EasyCrypt semantics of shift operators
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -46,7 +46,6 @@ type tyerror =
   | UnknownPrim of A.symbol * string
   | PrimWrongSuffix of A.symbol * Sopn.prim_x86_suffix list
   | PtrOnlyForArray
-  | WriteToConstantPointer of A.symbol
   | PackSigned
   | PackWrongWS of int
   | PackWrongPE of int
@@ -206,9 +205,6 @@ let pp_tyerror fmt (code : tyerror) =
 
   | PtrOnlyForArray -> 
     F.fprintf fmt "Pointer allowed only on array"
-
-  | WriteToConstantPointer v ->
-    F.fprintf fmt "Cannot write to the constant pointer %s" v
 
   | PackSigned ->
     F.fprintf fmt "packs should be unsigned"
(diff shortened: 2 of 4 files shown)


compiler/src/pretyping.ml
```diff
@@ -1721,15 +1717,15 @@ let mk_call loc inline lvs f es =
   | Internal -> ()
   | Export _ ->
     if not inline then
-      warning Always (L.i_loc0 loc) "export function will be inlined"
+      warning RecoverableError (L.i_loc0 loc) "export function will be inlined"
```
vbgl, line 1720: Please leave this warning as-is.

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@
   therefore old intrinsic `#VPBLENDVB` is deprecated
   ([PR #1010](https://github.com/jasmin-lang/jasmin/pull/1010)).
 
+- Add an option to treat some pre-typing error as warning instead. 
+  ([PR #1023](https://github.com/jasmin-lang/jasmin/pull/1023))
+  fixes [#727](https://github.com/jasmin-lang/jasmin/issues/727).
```
vbgl, line 40: I don’t think 727 is fixed yet. But this is indeed necessary preparatory work.

compiler/src/utils.ml
```diff
@@ -426,8 +426,12 @@ type warning =
   | Deprecated
   | Experimental
   | Always
+  | RecoverableError
```
vbgl, line 429: I find the name poorly chosen: these warnings are errors by mistake. They are not error. Calling them error again is wrong.

compiler/src/utils.ml
```diff
@@ -438,18 +442,26 @@ let add_warning (w:warning) () =
 
 let nowarning () = warns := Some []
 
+
 let to_warn w = 
   match !warns with
   | None -> true
-  | Some ws -> w = Always || List.mem w ws 
+  | Some ws -> w = Always || w = RecoverableError || List.mem w ws
```
vbgl, line 449: You don’t need to change this line.

compiler/src/pretyping.ml
MrDaiki: Yes, I modify it by mistake, it is corrected now

compiler/src/utils.ml
MrDaiki: I agree that the name is not the best. Would something like `NotHandledWarning` would be better ?

compiler/src/utils.ml
MrDaiki: If we don't add this line dont think that these warnings will be displayed because we don't have any way to add `RecoverableError` to the `warns` list. In fact we could handle this case by adding a command line option to display those warning (or do imply with this comment that you don't want them to be displayed at all in the case they are treated like warnings ?)

compiler/src/utils.ml
vbgl: `Legacy`, `Fatal`, and `Bug727` come to mind.

compiler/src/utils.ml
vbgl: By default (`warns` is `None`), they will be displayed. Moreover, the `add_warning` function can be used to add it to the list.

compiler/src/utils.ml
vbgl: What about `PedanticPretyping`?

vbgl (approved):

PR #1027, opened 2025-01-27
Regalloc: improve printing of equiv. traces

### LINKED ISSUE #1026: Register allocation error shows useless wrong line information
Compiling the following program (with jasminc 2024.07.2) produces the error message reproduced below. The φ-nodes listed in that message carry a dummy information that should not be displayed.

~~~
export fn main(reg u32 x) -> reg u32 {
  reg u32 y = x;
  if x <s 0 {
    y = -x;
  } else {
    y = x;
  }
  x += y;
  return x;
}
~~~

~~~
"…/bug_1026.jazz", line 3 (2) to line 7 (3):
compilation error:
register allocation: conflicting variables “x.157” and “y.208” must be merged due to:
  at line -1 (-1):
    y.208 = y.206; /* u32:φ */
  at line -1 (-1):
    y.208 = y.207; /* u32:φ */
  at "tests/fail/register_allocation/x86-64/bug_1026.jazz", line 4 (4-11):
    ( _1.201,  _2.202,  _3.203,  _4.204,  _5.205, y.206) =
      #NEG_32(x.157); /*  */
  at "tests/fail/register_allocation/x86-64/bug_1026.jazz", line 8 (2-9):
    ( _6.209,  _7.210,  _8.211,  _9.212,  _10.213, x.214) =
      #ADD_32(x.157, y.208); /*  */
~~~

vbgl:
Fixes #1026.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@
   process
   ([PR#989](https://github.com/jasmin-lang/jasmin/pull/989)).
 
+- Incorrect printing of line information in register allocation has been removed
+  ([PR #1027](https://github.com/jasmin-lang/jasmin/pull/1027);
+  fixes [#1026](https://github.com/jasmin-lang/jasmin/issues/1026)).
+
 ## Other changes
 
 - Adding an annotation to function (`'info` type). Useful to store result of analysis. (see [[#1016](https://github.com/jasmin-lang/jasmin/issues/1016)])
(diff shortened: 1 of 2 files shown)


compiler/tests/fail/register_allocation/x86-64/bug_1026.jazz
```diff
@@ -0,0 +1,10 @@
+export fn main(reg u32 x) -> reg u32 {
```
bgregoir, line 1: I don't understand the interest of the test.
I mean we don't check that printing will not change, so what are we testing here ?

vbgl: You are right. This is just an example that I’ve been using to check these changes. However, there is no need to keep in in the repository. Let me remove it.

vbgl: Done.

bgregoir (approved):

PR #1028, opened 2025-01-27
Fixing type checking of function return storage type

### LINKED ISSUE #18: The storage modifier of the return type of a function is ignored
If we compile the following function in the `glob_array3` branch
```
 inline fn sum (stack u64[2] a) -> stack u64 {
  reg u64 r;
  
  r = a[0];
  r += a[1];
    
  return r;
}
```
the `stack` storage in the type is simply ignored and replace by `reg`. If we replace `stack` with either `reg ptr` or `stack ptr`, we get the same result. This is a bit confusing.

Maybe this is done on purpose to have a uniform syntax, and it is known that the storage modifier in the return type of a function is ignored and instead inferred from the body of the function, in which case this issue can be marked as invalid, but I prefer to ask.

MrDaiki:
# Issue 

Compiler wasn't checking coherence of storage types between declared return storage type and return variable storage type. For more informations see #18 

# Solution 

Defining a new function in `pre-typing` to check for this.

# Implication

This change break retrocompatibility making some old program not valid. (some tests of the test suite had to be modified)

Fixes #18

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -28,6 +28,7 @@ type tyerror =
   | InvalidOperator     of sop
   | NoReturnStatement   of P.funname * int
   | InvalidReturnStatement of P.funname * int * int
+  | InvalidSignatureStorage of P.funname * S.pstorage * A.symbol * W.v_kind
   | InvalidArgCount     of int * int
   | InvalidLvalCount    of int * int
   | DuplicateFun        of A.symbol * L.t
@@ -142,6 +143,13 @@ let pp_tyerror fmt (code : tyerror) =
 
   | InvalidReturnStatement (name, given, expected) ->
       F.fprintf fmt "return statement of function %s has %d values instead of %d (as claimed by the signature)" name.P.fn_name given expected
+
+  | InvalidSignatureStorage (fname, sto, var_name, var_kind) ->
+      F.fprintf fmt "In function “%s” return statement variable “%s” has storage type “%a”, which differ from declared return storage type “%s”"
+        fname.P.fn_name
+        var_name
+        PrintCommon.pp_kind var_kind
+        (S.pp_storage sto)
 
   | InvalidArgCount (n1, n2) ->
       F.fprintf fmt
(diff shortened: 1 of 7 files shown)


compiler/tests/fail/common/storage_signature_mismatch.jazz
```diff
@@ -0,0 +1,8 @@
+inline fn sum (stack u64[2] a) -> stack u64 {
+  reg u64 r;
+  
+  r = a[0];
+  r += a[1];
+    
+  return r;
+}
```
vbgl, line 8: Negative examples should be minimal and focused to avoid them failing for unrelated reasons.

compiler/src/pretyping.ml
```diff
@@ -665,6 +673,23 @@ let check_return_statement ~loc name (declared : P.pty list) (given : (L.t * P.p
...
+      | `Global, W.Global -> ()
+      | `Reg ptr, W.Reg (_, ref) | `Stack ptr, W.Stack ref -> (
+        match ptr,ref with 
+        | `Pointer (Some `Writable), W.Pointer W.Writable -> ()
+        | `Pointer (Some `Constant), W.Pointer W.Constant -> ()
+        | `Pointer None, W.Pointer _ -> ()
+        | `Direct, W.Direct -> ()
+        | _ -> rs_tyerror ~loc (InvalidSignatureStorage(fname, x, (L.unloc y).v_name,(L.unloc y).v_kind))
+      )
+      | _ -> rs_tyerror ~loc (InvalidSignatureStorage(fname, x, (L.unloc y).v_name,(L.unloc y).v_kind))
```
vbgl, line 690: This catch-all pattern is fishy. Please list all possible cases.

compiler/src/pretyping.ml
```diff
@@ -665,6 +673,23 @@ let check_return_statement ~loc name (declared : P.pty list) (given : (L.t * P.p
     (fun ty1 (loc, ty2) -> check_ty_eq ~loc ~from:ty2 ~to_:ty1)
     declared given
 
+let check_return_storage ~loc fname (sto_list:S.pstorage list) (retvar_list: 'len P.gvar_i list) = 
+  List.iter (
+    fun (x,y:(S.pstorage*'len P.gvar_i)) -> 
+      match x,(L.unloc y).v_kind with
+      | `Inline, W.Inline -> ()
+      | `Global, W.Global -> ()
```
vbgl, line 681: This is strange.

vbgl:
This unveils issues in the libjade source code. Unfortunately, the source code of libjade that we use in CI is a frozen snapshot.

It is time to fix these libjade CI jobs.

vbgl:
Here is a fix for libjade: https://github.com/formosa-crypto/libjade/pull/131

eponier (approved):

Issue #994, opened 2024-12-13
Add non-regression test-case for the Easycrypt extraction


lyonel2017:

vbgl:
Is this issue still relevant?
ping @cassiersg

cassiersg:
We could certainly have more tests, but I won't have time in the coming months to develop such tests myself... maybe we should close this and try to have tests added whenever the extraction code is modified?

PR #1032, opened 2025-01-30
EC: add unit tests for extraction

vbgl:
This is a first step addressing #994 and an invitation to contribute more unit tests.

This PR adds two cases related to bugs recently fixed.

Writing EasyCrypt proof scripts that are compatible with current and next versions of EasyCrypt is challenging.


bgregoir (commented):
A question: nix-shell also install easycrypt ?
If not should we do it?
Else, I am fully in agree with this PR.

vbgl:
Yes. There is all that is needed to install latest release or current development version of EasyCrypt. This is what CI uses.

bgregoir (approved):
Sorry to have been so long

PR #1034, opened 2025-01-31
check-risc-v: avoid non-POSIX (bash) constructions

vbgl:
This makes it possible to run the RISC-V checks without Bash.

Side-note: changing the name of the assembler depending on the OS kernel is wrong.

Reported by Alexandre (@MrDaiki).

### DIFF
--- a/compiler/scripts/check-risc-v
+++ b/compiler/scripts/check-risc-v
@@ -19,16 +19,16 @@ ASSEMBLY_CHECKER_OPTIONS=${ASSEMBLY_CHECKER_OPTIONS:---triple=riscv32 --mcpu=gen
 UNAME_S=$(uname -s)
 
 # Conditionally set ASSEMBLY_CHECKER for Darwin (macOS)
-if [ "$UNAME_S" == "Darwin" ]; then
+if [ "$UNAME_S" = "Darwin" ]; then
     ASSEMBLY_CHECKER="riscv64-unknown-elf-as"
     ASSEMBLY_CHECKER_OPTIONS=""
-elif [ "$UNAME_S" == "Linux" ]; then
+elif [ "$UNAME_S" = "Linux" ]; then
     ASSEMBLY_CHECKER="llvm-mc"
     ASSEMBLY_CHECKER_OPTIONS="--triple=riscv32 --mcpu=generic-rv32 -mattr=+m --filetype=obj"
 fi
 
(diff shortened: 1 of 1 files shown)


clebreto (approved):

clebreto:
> changing the name of the assembler depending on the OS kernel is wrong.

Could you please elaborate on the rationale behind the statement ? What would be the alternative to facilitate the execution of the tests (down to assembly generation) for a maximum of users ?

vbgl:
> Could you please elaborate on the rationale behind the statement ?

Rationale? WDYM. It is just wrong. Like looking at the latitude to know whether you should wear a hat.

vbgl:
There is a `default.nix` file that describes in a machine readable way how to get the dependencies to run the test suite. In particular this description picks a precise version of LLVM and puts a `llvm-mc` binary in PATH irrespective of the nickname of your os kernel.

clebreto:
Thanks for reaching back.

> Rationale ? WDYM.
By rationale, I meant reasons / logic.

> There is a `default.nix` file that describes in a machine readable way how to get the dependencies to run the test suite. In particular this description picks a precise version of LLVM and puts a `llvm-mc` binary in PATH irrespective of the nickname of your os kernel.

This does imply that all users rely on nix environments, which is not enforced by jasmin development policies : uncertain weather -> some would wear the hat while other wouldn't =).


I am nonetheless more that willing to make the changes to use "llvm-mc" regardless of the OS / distribution / environment and document how to "alias" an homebrew-installed llvm assembler to be found by the testing suite. 

Would that fit better in the current state of the Jasmin testing suite ?

vbgl:
> By rationale, I meant reasons / logic.

The reasons for my comment was to inform you that there is a bug in some code you wrote.

> This does imply that all users rely on nix environments

Not at all. Before the addition of the `risc-v` checks, the test suite would rely on users of the test-suite having `llvm-mc` in PATH.

(see e.g. https://github.com/jasmin-lang/jasmin/blob/main/compiler/scripts/check-arm-m4#L26).

After the addition of risc-v checks, there are two new constraints:

  1. `uname` should be `Linux` or `Darwin`.
  2. if `uname` is `Darwin`,  then `riscv64-unknown-elf-as` must be in PATH.

Both are wrong. The first one because there are many kernels around. The second one because it is unnecessary (as `llvm-mc` still must be in PATH for the ARM test suite) and because it is not documented anywhere.

PR #1042, opened 2025-02-05
Add support for the -I parameter to jasmin2ec, jasmin-ct and jasmin2tex.

cassiersg:
Behavior for these tools is now identical to jasminc: `-I` parameter and JASMINPATH are merged.


cassiersg:
cc @mbbarbosa

(Post-mojito & not tested... use at your own risk.)

PR #1043, opened 2025-02-05
Add ext_eq for words in JByte_array

cassiersg:


cassiersg:
cc @mbbarbosa

vbgl (approved):

eclib/JByte_array.ec
```diff
@@ -80,32 +80,35 @@ abstract theory ByteArray.
       by rewrite _nth_of_list 1:size_map 1:size_iota 1:/# (nth_map 0) 1:size_iota 1:/# nth_iota.
     qed.
 
-    lemma get_set'SE t i j w:
+    lemma get_set'SE_eq t i j w:
       0 <= i => r*(i + 1) <= ByteArray.size =>
-      get'S (set'S t i w) j = if i = j then w else get'S t j.
+      i = j =>
```
vbgl, line 85: Question: why are there two variables `i` and `j` if the lemma is only to be applied when we have a proof that there are equal?

cassiersg: I have no answer. Maybe @bgregoir knows ?

bgregoir: I think because we did cut and past. I don't see a good reason. It is simpler to apply but harder to rewrite.

PR #1045, opened 2025-02-06
Gimli proof: fix for latest EasyCrypt

vbgl:
Recent changes (since yesterday, maybe https://github.com/EasyCrypt/easycrypt/pull/708) in EasyCrypt development branch made this proof fail. Bugfix or regression?


eponier:
During a presentation done by Manuel in Porto, a strange interaction between memory and rewrite was discovered, and PY implemented a fix soon afterwards. But I don't know whether it legitimately fails in our case, or if the fix is too strict. Let's try to ping @strub .

vbgl:
I think it’s fine here: introducing two “memory” variables to enable a rewrite is not a big restriction.

eponier (approved):
I agree. I would have liked to hear from PY, but let us fix the CI.

PR #1047, opened 2025-02-07
ToEC: deprecate mode CTG

vbgl:


cassiersg (dismissed):
Looks good to me.

compiler/examples/gimli/proofs/Makefile
```diff
@@ -2,33 +2,27 @@ ECARGS ?= -I Jasmin:../../../../eclib
 
 JASMIN2EC := ../../../jasmin2ec
 
-CT_MODE ?= CT
-GIMLI_CT = gimli_$(shell echo $(CT_MODE) | tr A-Z a-z).ec
-
 .SUFFIXES: .jazz .ec
 
-all: ct gimli_arm.ec gimli_x86.ec gimli_x86_ct.ec gimli_arm_ct.ec gimli_avx.ec gimli_avx_ct.ec
+all: gimli_ct.ec gimli_arm.ec gimli_x86.ec gimli_x86_ct.ec gimli_arm_ct.ec gimli_avx.ec gimli_avx_ct.ec
```
eponier, line 7: gimli_ct.ec is not a target. Could you also use the same order for the dependencies of `all` and for the declaration of the targets. It helps checking that everything is fine.

compiler/entry/jasmin2ec.ml
```diff
@@ -59,7 +59,7 @@ let model =
     $(b,CT): Functions additionally return timing-observable leakage for
     'cryptographic constant time' (if/while conditions, memory access
     addresses, array indices, for loop bounds).
-    $(b,CTG): Cryptographic constant time leakage is added to a
+    (Deprecated) $(b,CTG): Cryptographic constant time leakage is added to a
```
eponier, line 62: As far as I understand, `doc` is for one-line description of an argument. You should use `man` to give a detailed explanation (and start a new paragraph for each possible argument to have a nice-looking man page). But this orthogonal to this PR.

vbgl: The [`info`](https://erratique.ch/software/cmdliner/doc/Cmdliner/Arg/index.html#val-info) value does not accept any `~man` argument, does it?

compiler/examples/gimli/proofs/Makefile
vbgl: I think I’ve addressed your comment: removed `gimli_ct.ec` from the list & sort it according to the order below.

compiler/entry/jasmin2ec.ml
eponier: Indeed, sorry, I was confused by the `info` function in `Cmd`.

eponier (approved):

PR #1052, opened 2025-02-14
Use elpi.derive instead of "Scheme Equality"

eponier:
elpi.derive is much faster on inductives with many constructors. This is particularly interesting for type x86_op which has more than 100 constructors.

### DIFF
--- a/proofs/_CoqProject
+++ b/proofs/_CoqProject
@@ -15,6 +15,7 @@
 -arg "-w -deprecated-since-mathcomp-2.3.0"
 -arg "-w -deprecated-from-Coq"
 -arg "-w -deprecated-dirpath-Coq"
+-arg "-async-proofs off"
 
 -R 3rdparty Jasmin
 -R arch Jasmin
--- a/proofs/compiler/arm_lowering_proof.v
+++ b/proofs/compiler/arm_lowering_proof.v
@@ -550,13 +550,16 @@ Proof.
       by rewrite /exec_sopn /= truncate_word_le // /= zero_extend_u.
     done.
 
-  t_xrbindP=> wbase' vbase hgetx hbase woff' voff hseme hoff wres hread ? hw;
-    subst ws''.
+  apply: rbindP => wbase'.
+  apply: rbindP => vbase hgetx hbase.
+  apply: rbindP => woff'.
+  apply: rbindP => voff hseme hoff.
+  apply: rbindP => wres hread /= /ok_word_inj [? hw]; subst ws''.
   move: hbase => /to_wordI [ws0 [wbase [? /truncate_wordP [hws0 ?]]]];
     subst wbase' vbase.
   move: hoff => /to_wordI [ws1 [woff [? /truncate_wordP [hws1 ?]]]];
     subst woff' voff.
-  move: hw => [?]; subst wres.
+  move: hw => /= ?; subst wres.
 
   split; last done.
   clear hfve.
@@ -1121,13 +1124,17 @@ Proof.
   case: b hw hwrite12' => hw hwrite12'.
   - move: hexeces.
     rewrite /exec_sopn /=.
-    case: ves => [// | ? []]; t_xrbindP=> //= v _ -> /= [->] ?; subst ws''.
-    move=> [?]; subst v.
+    case: ves => [// | ? []]; last by t_xrbindP.
+    apply: rbindP => v /=.
+    apply: rbindP => _ -> [->] /ok_inj /(f_equal (head (Vbool true))) /= /Vword_inj [?]; subst ws''.
+    move=> /= ?; subst v.
     by rewrite truncate_word_le.
   move: hexeces.
   rewrite /exec_sopn /=.
-  case: ves => [// | ? []]; t_xrbindP=> //= v _ -> /= [->] ?; subst ws''.
-  move=> [?]; subst v.
+  case: ves => [// | ? []]; last by t_xrbindP.
+  apply: rbindP => v /=.
+  apply: rbindP => _ -> [->] /ok_inj /(f_equal (head (Vbool true))) /= /Vword_inj [?]; subst ws''.
+  move=> /= ?; subst v.
   by rewrite truncate_word_le.
 Qed.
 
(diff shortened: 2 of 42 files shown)


proofs/_CoqProject
```diff
@@ -15,6 +15,7 @@
 -arg "-w -deprecated-since-mathcomp-2.3.0"
 -arg "-w -deprecated-from-Coq"
 -arg "-w -deprecated-dirpath-Coq"
+-arg "-async-proofs off"
```
eponier, line 18: Oops, should not be there.

proofs/compiler/stack_alloc.v
```diff
@@ -105,19 +107,15 @@ End CmpR.
 Module Mr := Mmake CmpR.
 
 (* ------------------------------------------------------------------ *)
+#[only(eqbOK)] derive Z.
```
eponier, line 110: We may want to move that elsewhere. In utils?

proofs/compiler/x86_lowering_proof.v
```diff
@@ -176,6 +176,11 @@ Section PROOF.
     end.
   Qed.
 
+  Lemma cons_inj {A} {x1 x2 : A} (s1 s2 : seq A) :
```
eponier, line 179: We may want to move this to utils.

proofs/lang/type.v
```diff
@@ -147,7 +143,7 @@ Module CEDecStype.
     | sword w1 =>
       match t2 as t0 return {sword w1 = t0} + {True} with
       | sword w2 =>
-        match wsize_eq_dec w1 w2 with
+        match Bool.reflect_dec _ _ (wsize_eqb_OK w1 w2) with
```
eponier, line 146: In the end, I reintroduced `wsize_eq_dec` so this change is not needed.

proofs/lang/type.v
```diff
@@ -169,7 +167,7 @@ Module CEDecStype.
     + case: pos_dec (@pos_dec_r n n' I) => [Heq _ | [] neq ] //=.
       move => _; apply/eqP => -[].
       by move/eqP: (neq erefl).
-    case: wsize_eq_dec => // eqw.
+    case: Bool.reflect_dec => // eqw.
```
eponier, line 170: idem

proofs/compiler/arm_lowering_proof.v
```diff
@@ -1121,13 +1124,17 @@ Proof.
   case: b hw hwrite12' => hw hwrite12'.
   - move: hexeces.
     rewrite /exec_sopn /=.
-    case: ves => [// | ? []]; t_xrbindP=> //= v _ -> /= [->] ?; subst ws''.
-    move=> [?]; subst v.
+    case: ves => [// | ? []]; last by t_xrbindP.
+    apply: rbindP => v /=.
+    apply: rbindP => _ -> [->] /ok_inj /(f_equal (head (Vbool true))) /= /Vword_inj [?]; subst ws''.
```
eponier, line 1129: TODO: use `cons_inv` instead

proofs/compiler/arm_lowering_proof.v
```diff
@@ -1121,13 +1124,17 @@ Proof.
...
+    move=> /= ?; subst v.
     by rewrite truncate_word_le.
   move: hexeces.
   rewrite /exec_sopn /=.
-  case: ves => [// | ? []]; t_xrbindP=> //= v _ -> /= [->] ?; subst ws''.
-  move=> [?]; subst v.
+  case: ves => [// | ? []]; last by t_xrbindP.
+  apply: rbindP => v /=.
+  apply: rbindP => _ -> [->] /ok_inj /(f_equal (head (Vbool true))) /= /Vword_inj [?]; subst ws''.
```
eponier, line 1136: idem

eponier:
Note to self: it seems we can remove `eq_axiom_of_scheme`.

proofs/_CoqProject
eponier: removed

proofs/compiler/x86_lowering_proof.v
eponier: removed

proofs/lang/type.v
eponier: reverted

proofs/lang/type.v
eponier: reverted

proofs/compiler/arm_lowering_proof.v
eponier: removed

proofs/compiler/arm_lowering_proof.v
eponier: removed

eponier:
I rebased on top of main. For some reason, I need to add some `clear` before `lia` to prevent the tactic from capturing some section variables that it did not capture before.

eponier:
I have reached a reasonable state. But I'm stuck on one thing. It works on my machine, but not in the CI because of the version of coq-elpi (CI uses 2.2.0). There are some errors related to scopes that I fixed [upstream](https://github.com/LPCIC/coq-elpi/pull/647) but the fix is only from 2.2.1. It is easy to write a backward compatible fix. But there is another problem that I mentioned above. For some reason, coq-elpi.2.2.x is bugged.

So the only sensible fix I see is to require a more recent version of coq-elpi. But maybe for that we have to wait the next release of nixos in the coming weeks (this is what I get from the name 25.05). @vbgl, what do you think?

vbgl:
What versionr of coq-elpi are known to work? I’ve just tried with 2.2.3 and got:
~~~
File "./compiler/arm_instr_decl.v", line 78, characters 0-3921:
Error: The elpi tactic/command derive failed without giving a specific error
message. Please report this inconvenience to the authors of the program.
~~~

eponier:
From 2.3.

vbgl:
There is no need for a major update of nixkpgs: a minor one should be enough.

eponier:
And thanks for taking care of updating the deps, once more!

eponier:
Indeed, give me 5 minutes.

vbgl (approved):
Thanks.

PR #1061, opened 2025-02-21
Deprecate the -I command-line option of jasminc

cassiersg:


vbgl:
What timing do you recommend for removal of this feature? deprecate in 2024.07.3 & remove in 2025.02.0? In that case you can do the removal in the same PR (only the first commit shall be integrated into the release-2024.07 branch).

cassiersg:
I wasn't aware that there was a plan to release both 2024.07.3 & 2025.02.0. I don't think that a long deprecation timeline is warranted since the fixes required are not invasive an pretty easy to apply. If you think removing in 2025.02.0 is ok, I'll push a commit to do that.

In that case, how do you want the CHANGELOG to be handled ? Adding a  "removal" bullet point before the "deprecation" ?

vbgl:
Are you going to take care of fixing libjade (which currently is in a frozen state)? Otherwise let’s delay the removal.

Two entries in the CHANGELOG sound good.

cassiersg:
No plan to touch libjade from my side.

vbgl:
So in your opinion who shall do it?

cassiersg:
I'm not involved into libjade (& I have no time to get involved), so I don't know. I touched this `-I` flag following a request by Manuel (and this PR implements the decision agreed at the last bi-weekly meeting).

@mbbarbosa perhaps you know who maintains libjade and how it's going to evolve, to answer @vbgl 's question ?

tfaoliveira:
I can do the libjade patch (today or tomorrow). 

As a non-essential note, I will miss the -I option. 
Farewell `-I`. May your paths always be found elsewhere.

vbgl:
> As a non-essential note, I will miss the -I option.

Can you please elaborate? So far, arguments I’ve heard in favor of `-I`  are:

1. it exists in `jasminc` (which does not tell whether it is useful)
2. it is used in some existing makefiles (which does not tell whether it is well designed)

arguments against:

1. its syntax is inconsistent with JASMINPATH
2. its functionality is superseeded by the environment variable (and setting an env. variable is usually as easy a setting an command-life argument).

tfaoliveira:
Since you kindly asked for an elaboration, I will share some thoughts. 

From my point of view, `JASMINPATH` and `-I` serve different purposes. `JASMINPATH` seems well-suited for stable environments—I have no objections to it and believe it should be maintained. 

I see it as analogous to `C_INCLUDE_PATH` and `-I` from "C"—`-I` is sometimes preferred because it is explicitly part of the command and is shorter/simpler to type.

Setting an environment variable is not compatible with parts of my workflow—it increases the risk of errors (e.g., pointing to the wrong directory; there are several), and manually typing a long variable for debugging is impractical if done frequently. I can easily script something that fixes this for me.

About the path to libjade: there is already a `Makefile` using `JASMINPATH`, which can be found here:
https://github.com/formosa-crypto/libjade/blame/13a094f740a37e18e8f8cb32630adaa6a3097af6/oldsrc-should-delete/Makefile.checksct#L14
The patch is quite straightforward, but I realized I need to rethink this and elaborate on a new libjade refactoring plan because the current one is stalled.

vbgl:
Thanks Tiago for the elaborate answer. I do not understand the reasons (e.g, why is it more error-prone to type `JASMINPATH=P=Q jasminc …` than to type `jasminc -I P:Q …`) but I do understand that you believe that the ability to set include paths through a command-line argument is necessary and preferably in a uniform way for the various command-line tools. I hope you’ll find PR #1068 to your liking.

cassiersg:
While I don't have a strong opinion on whether we should have `-I` or not, I think that consistency is important and we should have `P=Q` or `P:Q` on both `-I` and `JASMINPATH`.

Would the breakage be manageable if we were to change to `P=Q` in `-I` ?

eponier:
In Coq (well, I'm not sure this is a good model to use, but this is the outside world), the environment variable `COQPATH` is a simplified version of `-R` or `-Q`. It is comma-separated, but you can specify just a path and the basename of the path is used for the logical name. Which makes it nearly impossible to relocate...

vbgl:
No. `COQPATH` is colon-separated. As is `OCAMLPATH`.

eponier:
Sorry this is what I meant, but my English level is bad...

eponier:
And how does it work in other languages? It seems they do not see the need of having this concept of logical names?

cassiersg:
`gcc -DVAR=value` might be a relevant precedent.

cassiersg:
Follow-up comment prompted by @vbgl 's message on zulip: I don't think `-I` should accept a list of include (so there is no need for two separators). Instead, we should repeat `-I` as many times as needed.

eponier:
> And how does it work in other languages? It seems they do not see the need of having this concept of logical names?

My question was more general than the idea of the syntax. In other languages, when you want to import code from another file/module, do you write something like `from name import name`? And when yes, what is `name`? Is it a logical name that you can map to whatever you like on the command line?

cassiersg:
C/C++/assembly use direct textual inclusion, without the possibility of having logical names: you `#include "path/to/file"` where the possible root paths to search are given by `-I` on the command-line. (OCaml is somewhat similar with `.mli`.)

Many other (more modern) languages (for those I know at least a bit: python, go, rust, swift) have a two-tiered system of code inclusion/dependency:
1. at a global level, some kind of package system, where packages are fairly independent units of code (e.g. can be hosted on public repositories such as https://pypi.org or https://crates.io ). They have logical names, and can be found by explicit path, by well-known online repository, link to a git repo, global OS path, etc. This is generally used for loosely coupled code, which we technically can't have in Jasmin at the moment. 
2. Each package can be implemented using multiple files. There, code inclusion is generally based on relative paths (for example `from .foo import bar as baz`, `mod foo; use foo::bar as baz;`), and the compiler only gets the path to a "top-level" file, then discovers by itself the other files - it doesn't nay any more path information.

As far as I understand, Jasmin is currently closer to the C-style model of code inclusion both in terms of how it is configured and in terms of semantics: the inclusion happens very early in the compilation process. Is it AST node replacement ? That would be sligthly later (hence more robust) than the C-style textual inclusion, but it's much less structured than modern systems where code inclusion is tied to namespacing. E.g., modern systems generally avoid depending on the order in which items/requires are given in a files: instead, a tree of names gets build (code inclusion creating new nodes in the namespace tree, items such as functions/types/constants being leaves), then names are resolved.

So overall, for the idea, I don't know languages that are very similar to Jasmin in how composition of files is handled. It might be worth to analyze the state of the art and try to improve jasmin based on that, but I don't think that this is a very pressing concern: it is not the most painful issue when trying to write big programs at the moment.

tfaoliveira:
Is there a design decision about precedence? For example, what happens when something is specified in `JASMINPATH` and `-I`?

vbgl:
Currently, if a name is defined twice, there is a compilation error.

eponier:
> So overall, for the idea, I don't know languages that are very similar to Jasmin in how composition of files is handled. It might be worth to analyze the state of the art and try to improve jasmin based on that, but I don't think that this is a very pressing concern: it is not the most painful issue when trying to write big programs at the moment.

Thanks @cassiersg for your detailed reply. But indeed, I don't want to block the current work on option `-I`.

PR #1063, opened 2025-02-21
Improve printing of problematic characters in jasmin2tex

### LINKED ISSUE #1060: Printing annotations in LaTeX
This code
```
#sct="{ptr: public, val: public}"
fn f(reg ptr u64[1] x) {}
```
generates
```
\#\jasminannotation{sct} = \jasminannotation{{ptr: public, val: public}}\\
\jasminkw{fn} \jasmindname{f}(\jasminstorageclass{reg ptr} \jasmintype{u64}[1] x) \{\\
\\
\}
```
which has no quotes or braces (they should be `\{` and `\}` instead of simple braces I guess?). Is this on purpose? I can create a PR if not.

sarranz:
Closes #1060. I believe I'm handling the `GEstring` case correctly, but I would like some confirmation.

### DIFF
--- a/compiler/src/latex_printer.ml
+++ b/compiler/src/latex_printer.ml
@@ -26,7 +26,6 @@ let arrow = symbol "arrow"
 let sharp fmt () = F.fprintf fmt "\\#"
 let openbrace fmt () = F.fprintf fmt "\\{"
 let closebrace fmt () = F.fprintf fmt "\\}"
-let ampersand fmt () = F.fprintf fmt "\\&"
 let percent fmt () = F.fprintf fmt "\\%%"
 let dollar fmt () = F.fprintf fmt "\\$"
 let underscore fmt () = F.fprintf fmt "\\_"
@@ -55,7 +54,6 @@ let pp_string fmt s =
   | '#' -> sharp fmt ()
   | '{' -> openbrace fmt ()
   | '}' -> closebrace fmt ()
-  | '&' -> ampersand fmt ()
   | '%' -> percent fmt ()
   | '$' -> dollar fmt ()
   | '_' -> underscore fmt ()
@@ -168,7 +166,8 @@ let pp_aligned =
 let rec pp_simple_attribute fmt a =
   match L.unloc a with
   | Aint i -> Z.pp_print fmt i
-  | Aid s | Astring s -> pannot fmt (Format.asprintf "%a" pp_string s)
+  | Aid s -> pannot fmt s
+  | Astring s -> pannot fmt (Format.asprintf "%a" pp_string s)
   | Aws ws -> Format.fprintf fmt "%a" ptype (string_of_wsize ws)
   | Astruct struct_ -> Format.fprintf fmt "(%a)" pp_struct_attribute struct_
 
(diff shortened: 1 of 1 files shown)


compiler/src/latex_printer.ml
```diff
@@ -22,9 +22,19 @@ let dname = latex "dname"
 let pannot = latex "annotation"
 let pprim = latex "primitive"
 let arrow = symbol "arrow"
+
 let sharp fmt () = F.fprintf fmt "\\#"
 let openbrace fmt () = F.fprintf fmt "\\{"
 let closebrace fmt () = F.fprintf fmt "\\}"
+let ampersand fmt () = F.fprintf fmt "\\&"
```
vbgl, line 29: This should not be needed.

compiler/src/latex_printer.ml
```diff
@@ -139,7 +168,7 @@ let pp_aligned =
 let rec pp_simple_attribute fmt a =
   match L.unloc a with
   | Aint i -> Z.pp_print fmt i
-  | Aid s | Astring s -> pannot fmt s
+  | Aid s | Astring s -> pannot fmt (Format.asprintf "%a" pp_string s)
```
vbgl, line 171: The old code should be correct for the `Aid` case.

compiler/src/latex_printer.ml
sarranz: My intention was that in the future we might have annotations that take strings and we want to print them. Should I restrict this to cover only the characters we allow inside annotations we support? (but I'm not sure how to check exhaustively all the annotations we support...)

compiler/src/latex_printer.ml
sarranz: Don't we allow underscores in identifiers?

vbgl: Underscores are letters (catcode 11).

compiler/src/latex_printer.ml
vbgl: Escaping ampersand is not needed, look at the printing of bitwise and operators below.

compiler/src/latex_printer.ml
sarranz: done

compiler/src/latex_printer.ml
sarranz: done

compiler/src/latex_printer.ml
```diff
@@ -36,6 +45,24 @@ let pp_opt p fmt =
...
+  String.iter @@ function
+  | '\\' -> backslash fmt ()
+  | '\'' -> quotesingle fmt ()
+  | '"' -> quotedouble fmt ()
+  | '#' -> sharp fmt ()
+  | '{' -> openbrace fmt ()
+  | '}' -> closebrace fmt ()
+  | '%' -> percent fmt ()
+  | '$' -> dollar fmt ()
+  | '_' -> underscore fmt ()
```
vbgl, line 59: Escaping underscores is not necessary. I trust you that you’ve check it does not harm.

vbgl (approved):

vbgl:
Of course it does harm:

~~~
u8[9] two_words = "two_words";
~~~

~~~
Undefined control sequence.
l.6 ...name{two_words} = \textquotedbl{}two\_words
                                                  \textquotedbl{};
~~~

PR #1064, opened 2025-02-24
LATEX: do not escape underscores

vbgl:
Follow-up on #1063.


sarranz:
Sorry, the way I was checking these was wrong. Thanks for fixing it.

sarranz (approved):

PR #1051, opened 2025-02-14
CHANGELOG: improve formatting

eponier:

PR #1066, opened 2025-02-24
Update CHANGELOG after release 2024.07.3

vbgl:
This should incorporate all changes proposed in #1051 (please complain if it is not the case).

Closes #1051.


eponier (approved):
Looks correct

Issue #993, opened 2024-12-13
Printing double semicolons clashes with double underscores


sarranz:
Jean-Christophe noticed that this program
```
export fn ble::e() -> reg u64 {
    reg u64 x = 0;
    return x;
}

export fn ble__e() -> reg u64 {
    reg u64 x = 1;
    return x;
}
```
generates
```
_ble__e:
ble__e:
	movq	$1, %rax
	ret
_ble__e:
ble__e:
	movq	$0, %rax
	ret
```
which is not good.
The compiler does check that we don't have a namespace `ble` with a function `e`, so we probably want to extend that?.

vbgl:
Fixed in #1067.

PR #1067, opened 2025-02-25
Adding a check to verify that functions names do not conflict when translated to assembly labels

MrDaiki:
# Issue 

Assembly generation use `PrintCommon.escape` function to build labels which can creates names conflicts because of namespaces. see (#993).

# Solution 

We introduce a map in `pretyping.ml` environment that check if a label is already used by another function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@
 - Fix EC extraction in case on nested loops
   ([PR #971](https://github.com/jasmin-lang/jasmin/pull/971)).
 
+- Program with conflicting assembly labels print warning when compiled
+  ([PR #1067](https://github.com/jasmin-lang/jasmin/pull/1067);
+  fixes [#993](https://github.com/jasmin-lang/jasmin/issues/993)).
+
 ## Other changes
 
 - The deprecated legacy interface to the LATEX pretty-printer has been removed
--- a/compiler/src/label_check.ml
+++ b/compiler/src/label_check.ml
@@ -1,0 +1,42 @@
+open Prog 
+open Utils
+
+
+type function_label = {
+  loc : L.t;
+  fname : string;
+}
+
+type duplicate_label_warn = {
+  first_decl : function_label;
+  conflict_decl : function_label;
+}
+
+let pp_duplicate_label_warn fmt {first_decl={loc=first_loc; fname=first_fname}; conflict_decl={loc=_;fname=second_fname}} =
+  Format.fprintf fmt
+    "Function '%s' and function '%s' (declared at %s) have conflicting assembly label names."
+    second_fname
+    first_fname
+    (Location.tostring first_loc)
+
+
+let warn_duplicate_label warn = 
+  warning Always (L.i_loc0 warn.conflict_decl.loc) "%a" pp_duplicate_label_warn warn
+
+let check_labels_fun (f: ('len,'info,'asm) gfunc) ((errors, label_map): (((duplicate_label_warn list) * function_label Ms.t ))) : (duplicate_label_warn list) * function_label Ms.t = 
+  let fn_label = PrintCommon.escape f.f_name.fn_name in
+  if f.f_cc != Internal then 
+    match Ms.find_opt fn_label label_map with
+    | None -> errors, Ms.add fn_label {loc=f.f_loc;fname=f.f_name.fn_name} label_map
+    | Some (label) -> 
+      let conflict = {first_decl=label; conflict_decl={loc=f.f_loc; fname=f.f_name.fn_name}} in
+      conflict::errors, label_map
+  else errors,label_map
+
+let get_labels_errors (prog: ('len,'info,'asm) gprog) = 
+    fst (List.fold_left (
+      fun label_map f -> 
+        match f with 
+        | MIfun f -> check_labels_fun f label_map
+        | _ -> label_map
+    ) ([],Ms.empty) prog)
--- a/compiler/src/label_check.mli
+++ b/compiler/src/label_check.mli
@@ -1,0 +1,33 @@
+open Prog
+
+
+(**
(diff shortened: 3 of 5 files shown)


vbgl:
The issue only arises for export function. Is it necessary to also fail when one of the functions is internal or inline?

MrDaiki:
> The issue only arises for export function. Is it necessary to also fail when one of the functions is internal or inline?

With this variant of original issue example : 
```
namespace ble {
    fn e() -> reg u64 {
        reg u64 x = 0;
        return x;
    }
}

fn ble__e() -> reg u64 {
    reg u64 x = 1;
    return x;
}

export fn sad() -> reg u64 {
    reg u64 x = ble::e();
    reg u64 y = ble__e();
    return y;
}
``` 

generated assembly code is : 
```
        .att_syntax
        .text
        .p2align        5
        .globl  _sad
        .globl  sad
_sad:
sad:
        movq    %rsp, %rsi
        andq    $-8, %rsp
        call    Lble__e$1
Lsad$2:
        call    Lble__e$1
Lsad$1:
        movq    %rsi, %rsp
        ret
Lble__e$1:
        movq    $1, %rax
        ret
Lble__e$1:
        ret
``` 

We can see that `ble::e` assembly code isn't generated, and that the two function calls are done on `ble__e`. So this issue exist even for internal functions.

I didn't found any bug using inline functions, and I don't expect one since inline dissapears before printasm. I will update the pull request to raise an error only if the function is not inline.

bgregoir (dismissed):
I think it will be nicer to not modify the type Env.env.
Write a checker that given a list gfunc ensure that this clash of name do not occurs.
We can call it when we want, maybe at the end of pretyping.

MrDaiki:
> I think it will be nicer to not modify the type Env.env. Write a checker that given a list gfunc ensure that this clash of name do not occurs. We can call it when we want, maybe at the end of pretyping.

I agree with your proposition, but this will require deeper work. For the moment, I will qualify this PR as draft so we can find a suitable (and reusable) solution.

MrDaiki:
I released a new version applying your suggestion @bgregoir. I think it is ready for review now.

compiler/src/label_check.ml
```diff
@@ -0,0 +1,43 @@
...
+    match Ms.find_opt fn_label label_map with
+    | None -> Ms.add fn_label (f.f_loc ,f.f_name.fn_name) label_map
+    | Some (loc, name) -> rs_duplicate_label_error ~loc:f.f_loc ((loc, name), f.f_name.fn_name)
+  else label_map
+
+(**
+Check if functions have conflicting names when translated to their assembly label names.
+Raise `DuplicateLabelError` if a conflict is found between two non inline functions (because inline function are substituted before assembly traduction).
+*)
+let check_labels (prog: ('len,'info,'asm) gprog) =
```
vbgl, line 35: This should return the list of conflicts instead of failing early.

compiler/src/main_compiler.ml
```diff
@@ -134,6 +134,13 @@ let main () =
       exit 0
     end;
 
+    ignore (
+      try Label_check.check_labels pprog with 
+      | Label_check.DuplicateLabelError (loc, code) ->
+        hierror ~loc:(Lone loc) ~kind:"label error"
```
vbgl, line 140: Also, a `warning` might be more appropriate.

MrDaiki:
I did  changes as suggested by @vbgl . Since It only print warning now, I didn't saw the interested or returning error list (at least in the current version of error handling). If you think it is better, I can add it.

CHANGELOG.md
```diff
@@ -27,6 +27,10 @@
 - Fix EC extraction in case on nested loops
   ([PR #971](https://github.com/jasmin-lang/jasmin/pull/971)).
 
+- Program with conflicting assembly labels print warning when compiled
+  ([PR #1067](https://github.com/jasmin-lang/jasmin/pull/1067));
+  fixes [#993](https://github.com/jasmin-lang/jasmin/issues/993).
```
vbgl, line 32:
```suggestion
  ([PR #1067](https://github.com/jasmin-lang/jasmin/pull/1067);
  fixes [#993](https://github.com/jasmin-lang/jasmin/issues/993)).
```

compiler/src/label_check.ml
```diff
@@ -0,0 +1,40 @@
...
+  if f.f_cc != Internal then 
+    match Ms.find_opt fn_label label_map with
+    | None -> errors, (Ms.add fn_label (f.f_loc ,f.f_name.fn_name) label_map)
+    | Some (loc, name) -> 
+      ((f.f_loc, ((loc, name), fn_label))::errors), label_map
+  else (errors,label_map)
+
+(**
+Check if functions have conflicting names when translated to their assembly label names.
+Raise `DuplicateLabelError` if a conflict is found between two non inline functions (because inline function are substituted before assembly traduction).
```
vbgl, line 32: The comment seems wrong.

vbgl (commented):
- Please add a `label_check.mli`
- What is the purpose of the test case?

compiler/src/label_check.ml
MrDaiki: It is (I didn't changed it in last version). It is corrected now

CHANGELOG.md
MrDaiki: Corrected

vbgl (approved):

MrDaiki:
I rewrote error type to be simpler to understand and change the error message to be more clear. I think it is ready to merge now.

compiler/src/label_check.mli
```diff
@@ -0,0 +1,33 @@
+open Prog
+
+
+(**
+Record types describing data used to detect and print informations about function labels. Used for error (warning) printing
+*)
+type function_label = {
+  loc : L.t;
+  fname : string;
+}
```
vbgl, line 10: What about `type function_label = string L.located`

PR #1068, opened 2025-02-26
Add support for the -I parameter to jasmin2ec, jasmin-ct and jasmin2tex

vbgl:
This reverts #1061 and revives #1042.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@
   ([PR #976](https://github.com/jasmin-lang/jasmin/pull/976) and
   [PR #986](https://github.com/jasmin-lang/jasmin/pull/986)).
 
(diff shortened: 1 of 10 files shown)


vbgl:
Marking as draft while bikeshedding on the possible separators.

eponier:
There is one thing that bothers me. Currently, strange things happen if `=` or `:` appear in paths. I acknowledge it is not common, but I don't think it is that hard to support, and it is better not to put constraints on the filesystem of the user. Just splitting on the first delimiter should be enough (this is what you did with `String.split` instead of `String.split_on_char` by the way).

vbgl:
Future work:

 - deal with repeated names
 - clarify whether delimiters can appear in paths

I think there is a workaround for paths with unusual spellings: use a symlink with only letters.

bgregoir (approved):

PR #1072, opened 2025-02-28
Update authors

eponier:
Some names were added in branch release-2024.07 but never in main. And add Parmida Javadian.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,6 +27,7 @@ Lucas Tabary-Maujean
 Manuel Barbosa
 Maxime Dénès
 Miguel Quaresma
+Parmida Javadian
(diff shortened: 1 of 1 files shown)


eponier:
Seems I forgot one more name

eponier:
I added Benjamin Salling Hvass whose name should have been there for a few releases already.

eponier:
I should really make a script at some point that checks that the AUTHORS file contains the name of the git contributors, but there are exceptions (I did not add the Coq maintainers who pushed here when something break on the Coq side, or one-line contributors in general). And people use sometimes different names in their commits (just the nickname, or their name without accents...)

vbgl (approved):

PR #1071, opened 2025-02-27
Wint2

bgregoir:
This PR introduce new types in jasmin source: siXX and uiXX.
For the compiler they are equal to wXX.
It also introduce the operator word_to_sint and word_to_uint 
    (int) w  : is word_to_uint      (for compatibility)
    (uint) w : is word_to_uint   
    (sint) w : is word_to_sint   ( NEW)
    (XXsi) w : cast from wXX to siXX 
    (XXsi) si : cast from siYY to siXX
    (XXsi) i : cast from int to sint  
    Same with XXui
    (int) si | (sint) si : cast from siXX to int
    (int) ui | (uint) si : cast from uiXX to int

New notation for uXX : wXX is not accepted and used for printing
Compilation starts by replacing all operations on siXX uiXX by corresponding wXX operations.
Extraction starts by replacing all operations on siXX uiXX by corresponding int operations.

PR #1090, opened 2025-03-10
Fix EC extraction of signed division

vbgl:
Extracted from https://github.com/jasmin-lang/jasmin/pull/1071 and added a non-regression test case.


compiler/examples/extraction-unit-tests/sdiv.jazz
```diff
@@ -0,0 +1,5 @@
+/* Extraction bug */
+export fn main() -> reg u64, reg u64 {
+reg u64 a = 1 b = -1 c = a /64s b d = -4 e = 3 f = d %64s e;
```
bgregoir, line 3: This syntax is really horrible ... I was not aware of it.

bgregoir (approved):

Issue #777, opened 2024-04-02
Debug info should have a location for return


eponier:
I played with the new flag `-g` on the following example
```
fn breakpoint () {}

export fn foo(reg u64 a_ptr) -> reg u64 {
    stack u8[2] a_s; reg u8 tmp;

    tmp  = (u8)[a_ptr + 0];
    tmp += 1;
    breakpoint ();
    a_s[0]  = tmp;
    tmp = (u8)[a_ptr + 8*1];
    tmp += 2;
    a_s[1] = tmp;

    reg u64 res;
    res = (64u)a_s[0];
    [a_ptr] = res;
    return res;
}
```
I observed two sources of wrong location:
- no location for returns, this is in particular problematic for breakpoint which has no instruction except an (implicit) return
- wrong location in the AST, but this is just https://github.com/jasmin-lang/jasmin/issues/22 striking again

PR #1094, opened 2025-03-13
Attach proper location to function prologue & epilogue

vbgl:
Partially addresses #777.

The main change is that linearization no longer introduces any `dummy_instr_info`.

### DIFF
--- a/compiler/src/prog.ml
+++ b/compiler/src/prog.ml
@@ -122,7 +122,7 @@ type ('len,'info,'asm) gfunc = {
     f_args : 'len gvar list;
     f_body : ('len,'info,'asm) gstmt;
     f_tyout : 'len gty list;
-    f_outannot : Annotations.annotations list; (* annotation attach to return type *)
+    f_ret_info : FInfo.return_info;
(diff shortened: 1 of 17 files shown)


compiler/src/prog.ml
```diff
@@ -123,7 +123,7 @@ type ('len,'info,'asm) gfunc = {
     f_body : ('len,'info,'asm) gstmt;
     f_tyout : 'len gty list;
     f_outannot : Annotations.annotations list; (* annotation attach to return type *)
-    f_ret  : 'len gvar_i list
+    f_ret  : 'len gvar_i list L.located
```
bgregoir, line 126: I have the impression that the change will be much smaller if you add the location in the field f_outannot.
Something like f_outannot : L.t * Annotations.annotations list;
This will allow to not have to patch all occurence of f_ret.
Then maybe f_outannot is not a good name anymore and can renamed.

vbgl: Thanks for this good advice. I’ve just implemented it.

bgregoir (approved):

PR #1096, opened 2025-03-17
Simplify proofs for SMTs

vbgl:
Example failure: https://gitlab.com/jasmin-lang/jasmin/-/jobs/9455675186#L1495


eponier (approved):
One other instance I guess.

PR #1099, opened 2025-03-18
Add support for abstract types

lyonel2017:


proofs/lang/sem_type.v
```diff
@@ -15,78 +15,84 @@ Definition nosubword   := {| sw_allowed := false |}.
...
-  + by case: ty => // ws [ws' [*]]; eauto.
-  by case: ty'; eauto.
+  + by case: ty => // ws [ws' [*]] ; eauto.
+    by case: ty'; eauto  => s; case s ; eauto.
 Qed.
 
 (* ----------------------------------------------------------- *)
-Definition sem_t (t : stype) : Type :=
+
+Class Tabstract := { iabstract : string -> Type }.
```
vbgl, line 56:
```suggestion
Class Tabstract := iabstract : string -> Type.
```

lyonel2017: Rocq's unifier occasionally fails with a singleton class defintion, where a definition with a record works. The Lemma `val_uincl_alt` in `values.v` is an example.

vbgl: You may report this issue to the Rock developers.

I’ve changed the statement of `val_uincl_alt` to something less involved.

I don’t think I can accept all these new `Build_Tabstract` in the OCaml code.

lyonel2017: I agree that the `Build_Tastract` are not very nice. Do you have a good idea of how to avoid them?

vbgl: Yes. First comment in this thread.

lyonel2017: That's right. At first it didn't work, and I was surprised. But it does work, I just made a mistake when recompiling.

compiler/src/lexer.mll
```diff
@@ -36,6 +36,8 @@
     "bool"  , T_BOOL ;
     "int"   , T_INT  ;
 
+    "abstract", ABSTRACT;
+
```
vbgl, line 40: Please do not add new unnecessary keywords. Concrete syntax suggestions:

~~~
type t;
~~~

or

~~~
param type t;
~~~

vbgl:
Please do not forget the latex printer.

compiler/src/lexer.mll
lyonel2017: Should be solved.

lyonel2017:
> Please do not forget the latex printer.

Should be solved.

eponier:
I'm reviewing this. I see lots of small things to fix, I'll just add a commit doing the changes, this will be more efficient for everyone.

vbgl:
Don’t waste your precious time Jean-Christophe. Recent discussion has clarified that this PR as no use-case, neither real nor perceived. It shall be closed.

eponier:
Oops, I spent the last two hours on it. But thanks, I'll stop there then.

vbgl:
I’ve been fooled too and spend some non-trivial effort fixing wild uses of JMeq.

PR #1100, opened 2025-03-19
Tests: disable a few redundant checks

vbgl:
It is wrong to verify that negative tests fail under any possible configuration (when said configuration is unrelated to the test).


vbgl:
Мерси.

PR #1101, opened 2025-03-24
Stack alloc: subarrays with non-constant indices

eponier:
Lots of TODOs:

- [x] fix proof
- [x] properly write the OCaml code in alias.ml/stackAlloc.ml (with help from @bgregoir)
  - certainly better than what I wrote before, but still worth a good review
- [x] remove byteset.v
- [ ] should we keep -print-stack-alloc-checker?
  - it is an option that prints the state of the Coq checker after each instruction; I used it a lot to debug while writing the pass; do we want it to be in main?
  - it is removed in last version
- [x] CHANGELOG
- [x] clean tests and add more
- [ ] better commit message
  - actually, the current one is concise but not that bad
- [ ] do we want to split in more files or just differently? the 1st file is huge
  - for instance, we could put everything related to sexpr in a separate file, this is just one idea
- [ ] do we want to keep stack_alloc_params?
  - since stack_alloc_params are defined in stack_alloc.v, each time this file is edited, it triggers the recompilation of x86_params, arm_params and riscv_params, and this was painful while developing the pass, so I moved the params to a dedicated file called stack_alloc_params. To some extent, I find the separate file to expose the params in a clearer way than before, but I won't fight to keep the split if it is considered not worth.
  - in the last version, we went further in this direction since we now also have stack_alloc_params_proof


eponier:
I cleaned the tests, and added many more.

I need some help to properly fix the OCaml oracle.

eponier:
Apart from cosmetic changes, this is ready, so I'm interested by getting reviews.

bgregoir:
do we want to keep stack_alloc_params?
I think it is always good if we can reduce dependencies.

bgregoir:
should we keep -print-stack-alloc-checker?
I have the impression that it is more debug info to develop the compiler that something that will help the developper,
so I can be wrong.

eponier:
Another question: what should we do about pass riscv_lower_addr, which allows to write complex array accesses in a transparent way on RISC-V (they are compiled into 3 instructions)? The pass was necessary as long as we could not use subarrays with variable index, but when this PR gets approved, should we deprecate it? It makes some code really simpler, but it also somehow hides the real cost of these accesses on RISC-V.

eponier:
After discussion with Benjamin, here are two features that we may want to add in the future (written here just for future reference):
- identify disjointness of symbolic subarrays in simple cases: for now, only subarrays with constant offsets and lengths can be proved disjoint, we could also manage simple cases like `a[i:N]` and `a[i+N:M]`. The case `j = i + N; a[i:N] and a[j:M] are disjoint` is much more complex, since stack alloc is after lowering and thus at this point `j = i + N` is a low-level operation that is hard to recognize
- be more precise when reading and writing constant indices: compared to the previous version, I deliberately made the choice of reasoning at (sub-)array level instead at cell level. Meaning that when we write to an array, we systematically invalidate the full zone of the array instead of just the cell; when we read from an array, we want the full array to be valid instead of just the cell. If we need more precision, we could change this.

compiler/tests/fail/stack_allocation/x86-64/range_overflow.jazz
```diff
@@ -1,9 +1,7 @@
-export
-fn f () -> reg u64 {
-  stack u64[2] a, b;
-  reg u64 res;
-  a[0] = 1;
-  b[1:1] = a[0:1];
-  res = b[1];
-  return res;
+export fn main (reg u64 x) -> reg u64 {
```
bgregoir, line 1: The error message it not clear (range/slice) this is internal code.

compiler/tests/fail/stack_allocation/x86-64/reg_var_expected.jazz
```diff
@@ -0,0 +1,9 @@
+export fn main () -> reg u64 {
```
bgregoir, line 1: Why knowing where the variable is defined is relevant for the error

compiler/tests/fail/stack_allocation/x86-64/regions_not_equal_array.jazz
```diff
@@ -0,0 +1,12 @@
+export fn main (reg u64 i) -> reg u64 {
+  stack u64[2] s;
+  reg ptr u64[2] r;
+  reg u64 res;
+
+  s[0] = 0;
+  r = s[i:2];
```
bgregoir, line 7: zone = [0:16][(8 * ((int /* of u64 */) i.222)):16] }
I think that "/* of u64 */" make the message hard to read.

bgregoir: Also the fact that we use long name

compiler/tests/fail/stack_allocation/x86-64/stack_ptr_in_expr.jazz
```diff
@@ -0,0 +1,9 @@
+export fn main (reg ptr u64[1] r) -> reg u64 {
+  stack ptr u64[1] sp;
+  reg u64 res;
+
+  sp = r;
+  res = sp[0] + 1;
```
bgregoir, line 6: stack allocation: sp (defined at "tests/fail/stack_allocation/x86-64/stack_ptr_in_expr.jazz", line 5 (2-4)) is a stack pointer, it should not appear in an expression

sp can appear in an expression like in r = sp.

compiler/tests/fail/stack_allocation/x86-64/unaligned_sub_offset.jazz
```diff
@@ -0,0 +1,10 @@
+// FIXME: double check
+export fn main (reg u64 i) -> reg u64 {
+  stack u32[4] s;
+  reg ptr u64[2] r;
+
+  r = s[i:4];
```
bgregoir, line 6: I think it is ok but it should not be an internal error.
  stack allocation: unaligned sub offset (r (defined at "tests/fail/stack_allocation/x86-64/unaligned_sub_offset.jazz", line 6 (2-3)))

compiler/tests/fail/stack_allocation/x86-64/unscaled_access.jazz
```diff
@@ -0,0 +1,10 @@
+// FIXME: double check
```
bgregoir, line 1: Same remark, not an internal error

compiler/tests/fail/stack_allocation/x86-64/writage_reg_ptr_args_not_disjoint.jazz
```diff
@@ -0,0 +1,16 @@
+fn f (reg ptr u64[1] r1, reg ptr u64[1] r2) -> reg ptr u64[1] {
+  r1[0] = r2[0];
```
bgregoir, line 2: Can you introduce a intermediate variable to be sure that the error will not come from the body of f.

proofs/compiler/arm_params.v
```diff
@@ -33,34 +35,53 @@ Context {atoI : arch_toIdent}.
 Definition is_load e :=
   if e is Pload _ _ _ _ then true else false.
 
+(* copied from riscv_lower_addressing, TODO: factor *)
```
bgregoir, line 38: Maybe this can be done (factor)

bgregoir:
This really represent a lot of work, thank!

compiler/tests/fail/stack_allocation/x86-64/range_overflow.jazz
bgregoir: proposed message "Cannot access to the subarray [16;32[ of  a.201[0; 24[, the access overflow, your program is probably unsafe"

proofs/compiler/stack_alloc.v
```diff
@@ -104,97 +110,220 @@ End CmpR.
...
+
+Definition is_const e :=
+  if e is Sconst n then Some n else None.
+
+(* A slice represents a contiguous portion of memory.
+  In [symbolic_slice], the components are symbolic expressions; symbolic
+  slices are used in the analysis, since we do not necessarily know the
+  offsets statically.
+*)
+(* We reuse [pexpr], but only the arithmetic part is of interest here. *)
```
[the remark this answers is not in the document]
eponier: Indeed

compiler/tests/fail/asmgen/x86-64/incompatible_args.jazz
```diff
@@ -1,7 +1,7 @@
 export fn main () -> reg u64 {
   reg u64 res;
   stack u128 s;
-  s = 3000000;
+  s = 300;
```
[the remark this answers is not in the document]
eponier: I don't know, I'll revert

compiler/tests/fail/stack_allocation/x86-64/expression_too_complex.jazz
```diff
@@ -0,0 +1,10 @@
+export fn main (reg u64 i) -> reg u64 {
+  stack u64[2] s;
+  reg ptr u64[1] r;
+
+  s[0] = 0;
+  r = s[i/(i+1):1];
```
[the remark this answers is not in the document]
eponier: done

compiler/tests/fail/stack_allocation/x86-64/merge_invalid_accesses.jazz
```diff
@@ -0,0 +1,10 @@
+// This is just a copy of bug_54.jazz, but easier to find
```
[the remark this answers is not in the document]
eponier: I added a comment explaining that this *currently* fails

proofs/compiler/stack_alloc.v
```diff
@@ -104,97 +110,220 @@ End CmpR.
...
+(* true : Valid, false : Unknown *)
+(* With just one-level precision, a sub-status is either fully Valid
+   or fully Unknown (actually, we could be a bit more precise, especially
+   with constants, but currently we stay simple). *)
+Fixpoint get_sub_interval (i : intervals) s : bool :=
+  match i with
+  | [::] => true (* valid *)
+  | s' :: i =>
+    if s == s' then false (* fully borrowed: unknown *)
+    else if odflt false (symbolic_slice_ble s s') then
```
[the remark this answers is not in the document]
eponier: I introduced a dedicated function

compiler/tests/fail/stack_allocation/x86-64/bad_alignment.jazz
```diff
@@ -1,16 +1,10 @@
-fn f (reg ptr u64[2] r) -> reg u32 {
-  reg u32 res;
-  stack u64[2] a;
-  stack u32[10] b;
-  a[0]=0;
-  b[1:2] = a[u32 0:2];
-  res = b[1];
-  return res;
-}
+// FIXME: double check
```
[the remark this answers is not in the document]
eponier: I renamed it into `bad_range_alignment.jazz` indeed. I change `bad_alignment.jazz` to be just a version of `bad_range_alignment` with a variable instead of a constant. The error messages for both files is bad for users.

compiler/tests/fail/stack_allocation/x86-64/expression_too_complex.jazz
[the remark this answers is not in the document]
eponier: And I renamed the file `array_access_too_complex.jazz`.

Precisely, here is the new error message
```
"tests/fail/stack_allocation/x86-64/array_access_too_complex.jazz", line 6 (2-19):
stack allocation: do not know how to compile this array access into a memory access,
                  the expression
                  ((int /* of u64 */) (i / 64u (i +64u ((64u) 1))))
                  must contain only the operators “+” and “*”
zsh: exit 1     ./jasminc tests/fail/stack_allocation/x86-64/array_access_too_complex.jazz

```

compiler/tests/fail/stack_allocation/x86-64/merge_same.jazz
```diff
@@ -0,0 +1,13 @@
+export fn foo () -> reg u64 {
+  reg u64 r;
+  stack u64[4] s;
+  reg ptr u64[4] rt;
+  stack ptr u64[4] st;
```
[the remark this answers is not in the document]
eponier: I made the test simpler.

compiler/tests/fail/stack_allocation/x86-64/merge_same3.jazz
```diff
@@ -0,0 +1,20 @@
+// FIXME: is the error/the error message correct?
```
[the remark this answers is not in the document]
eponier: I renamed the test `merge_distinct_regions.jazz` and added comments giving details.

compiler/tests/fail/stack_allocation/x86-64/merge_too_large.jazz
```diff
@@ -0,0 +1,6 @@
+export fn f (reg ptr u64[1] r) -> reg ptr u64[1] {
```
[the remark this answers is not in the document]
eponier: Neither do I ;)

compiler/tests/fail/stack_allocation/x86-64/partial_region.jazz
```diff
@@ -0,0 +1,14 @@
+export fn main () -> reg u64 {
+  stack u64[2] s;
+  reg ptr u64[1] r;
+  reg u64 tmp res;
+
+  s[0] = 0; s[1] = 1;
+  r = s[0:1];
+  r[0] = 1; // after writing to r, s is partial
```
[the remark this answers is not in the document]
eponier: I added a comment.

compiler/tests/fail/stack_allocation/x86-64/merge_too_large.jazz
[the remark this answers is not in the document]
eponier: The error message appears when trying to merge a global or a param with a local. I made this clear in the error message. Not sure this is really clearer for users, though.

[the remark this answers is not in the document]
eponier:
```
"tests/fail/stack_allocation/x86-64/merge_too_large_param.jazz", line 4 (2-18):
stack allocation: cannot merge a param and a local that is larger (r.185 of size 8, and a.195 of size 16)
```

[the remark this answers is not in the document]
eponier: (I made two tests, one testing param and one testing global)

compiler/tests/fail/stack_allocation/x86-64/range_overflow.jazz
eponier: I followed your suggestion.
```
"tests/fail/stack_allocation/x86-64/range_overflow.jazz", line 4 (2-13):
stack allocation: cannot access the subarray [16; 32[ of a.200[0; 24[, the access overflows, your program is probably unsafe
```

compiler/tests/fail/stack_allocation/x86-64/reg_var_expected.jazz
eponier: cf. https://github.com/jasmin-lang/jasmin/pull/1124

compiler/tests/fail/stack_allocation/x86-64/regions_not_equal_array.jazz
eponier: I fixed the long name (not shown if `-debug` is not passed). For the cast, cf. https://github.com/jasmin-lang/jasmin/pull/1124.

compiler/tests/fail/stack_allocation/x86-64/stack_ptr_in_expr.jazz
eponier:
```
"tests/fail/stack_allocation/x86-64/stack_ptr_in_expr.jazz", line 6 (8-10):
stack allocation: sp (defined at "tests/fail/stack_allocation/x86-64/stack_ptr_in_expr.jazz", line 5 (2-4)) is a stack ptr, it should not appear in a sub-expression
```

compiler/tests/fail/stack_allocation/x86-64/unaligned_sub_offset.jazz
eponier: New error message
```
"tests/fail/stack_allocation/x86-64/unaligned_sub_offset.jazz", line 9 (8-9):
stack allocation: the access to array
                  r (defined at "tests/fail/stack_allocation/x86-64/unaligned_sub_offset.jazz", line 8 (2-3))
                  could not be proved to be aligned;
                  if you know what you are doing or want to perform an unaligned access,
                  you can use “#unaligned”
```

compiler/tests/fail/stack_allocation/x86-64/unscaled_access.jazz
eponier: See previous comment

compiler/tests/fail/stack_allocation/x86-64/writage_reg_ptr_args_not_disjoint.jazz
eponier: I did it.

compiler/tests/fail/stack_allocation/x86-64/merge_same3.jazz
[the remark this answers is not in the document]
eponier: In the end, I just deleted it, because there are merge_intersection.jazz and merge_intersection2.jazz.

eponier:
I rebased on latest main.

eponier:
> Another question: what should we do about pass riscv_lower_addr, which allows to write complex array accesses in a transparent way on RISC-V (they are compiled into 3 instructions)? The pass was necessary as long as we could not use subarrays with variable index, but when this PR gets approved, should we deprecate it? It makes some code really simpler, but it also somehow hides the real cost of these accesses on RISC-V.

Still a question

bgregoir:
For your last point, I think we can keep it as it is for the moment.
Did someone has an objection, or I can merge ?

vbgl (dismissed):
Don’t forget to clean the history before or while merging.

bgregoir:
@eponier do you want to clean the history or I squash and merge ?

eponier:
I take care of it.

eponier:
Done

bgregoir:
I wait for the checks, we can certainly merge tomorrow.

bgregoir (approved):

eponier:
:tada:

PR #1107, opened 2025-03-31
Use relative paths in -print-dependencies

vbgl:
This makes the output of `-print-dependencies` reusable in different contexts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@
 
 ## Other changes
 
+- Option `-print-dependencies` now uses relative paths
+  ([PR #1107](https://github.com/jasmin-lang/jasmin/pull/1107)).
+
 - Add warning to signal deprecated intrinsic operators 
   ([PR #1092](https://github.com/jasmin-lang/jasmin/pull/1092)).
 
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -428,7 +428,8 @@ end  = struct
   let decls env = env.e_decls 
 
   let dependencies env =
-    Map.fold ( @ ) env.e_loader.loaded []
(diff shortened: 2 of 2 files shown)


vbgl:
Suggestion: use relative or absolute depending on how the compiler gets aware of a given file.

eponier:
I wanted to experiment with the feature in main and in this PR to have a better idea of what the corresponding behaviors are.

Here is my setup:
`/tmp`: working directory
`/tmp/src/file.jazz`: source file

- without logical name
  - contents of `file.jazz`: `require "libfile.jazz"`
  - output of `-print-dependencies`:
    - main: `/tmp/src/libfile.jazz`
    - this PR: `src/file.jazz`
    -  if the file not there, the error is relative `typing: error reading file "src/libfile.jazz" (src/libfile.jazz: No such file or directory)`

- with a logical name
  - contents of `file.jazz`: `from Lib require "libfile.jazz"`
  - with a relative path `-I Lib=./lib`
    - main: `/tmp/lib/libfile.jazz`
    - this PR: `src/file.jazz`

  - with an absolute path `-I Lib=/tmp/lib`
    - main: `/tmp/lib/libfile.jazz`
    - this PR: `src/file.jazz`

- if the file is not there, the error is always absolute `typing: error reading file "/tmp/lib/libfile.jazz" (/tmp/lib/libfile.jazz: No such file or directory)`

eponier:
Regardless of what the relative/absolute aspect, there is something wrong the current state of the PR, because it prints the name of the source file rather than the name of the dependency.

eponier:
I'll see what I manage to do about this.

vbgl:
The change proposed in this PR is wrong. Let’s close it and continue the discussion about `print-dependencies` somewhere else.

PR #1110, opened 2025-04-02
changed pseudo-monadic notation (>>= -> >>r= | >>o=, similarly for >>…

ptorrx:
Minor modification of the pseudo-monadic notation (>>= and >>) for 'result' and 'option', proposed in order to avoid conflicts with the actual monadic notation, which is widely used in the Interaction Trees development, and prevent confusion.


vbgl:
Aren’t scopes powerful enough to disambiguate such an overloaded notation?

ptorrx:
There are going to be files where we use both, so that would require to locally scope each occurrence with %

bgregoir:
indeed

Issue #1112, opened 2025-04-03
ARM: invalid assembly when using MULS


vbgl:
Consider the following program.

~~~
param int N = 9;

export fn main(reg u32 x) -> reg u32 {
  reg u32[N] r;
  inline int i;
  for i = 0 to N { r[i] = x; }
  for i = 0 to N {
    ?{}, r[i] = #MULS(r[i], x);
  }
  return x;
}
~~~

The assembly generated by `jasminc` (2024.07.3) is not valid. The LLVM assembly raises three different errors:

- invalid instruction (`MULs lr, lr, r0`)
- operand must be a register in range [r0, r7] (`MULs lr, r4, r0`)
- destination register must match source register (`MULs r1, r2, r0`)

vbgl:
Fixed in #1159.

Issue #1040, opened 2025-02-04
Instruction request: SHA256RNDS2, SHA256MSG1 and SHA256MSG2


ruipedro16:

vbgl:
Are there EasyCrypt or Coq specification of these functions that could be reused?

ruipedro16:
Im not aware of any

PR #1115, opened 2025-04-08
Generalize forced explicit registers in instr. descriptors

vbgl:
Some x86 instructions have a mandatory explicit operand XMM0 (e.g. SHA256RNDS2)

Some ARM instructions only accept registers in the [r0; r7] range (e.g., MULS)


vbgl:
This is preparatory work to address #1112 and #1040 (left as future work).

bgregoir (approved):

PR #1118, opened 2025-04-08
Asm target rewriting

MrDaiki:
This pull request objective is to rewrite assembly printers of the jasmin compiler. So far, each printer was using it's own functions for assembly printing. In particular, X86 printer used a different behaviour than the other architectures.

This pull request implement the following changes : 
* factorisation of all duplicate code in assembly pretty printers
* Rewriting of PrintASM module as a standard assembly printer
* Make all assembly printers use interface of PrintASM module
* Some refactoring of the code

It has been tested on a lot of jasmin sources using this [repo](https://gitlab.inria.fr/vlaporte/get-jasmin-assembly). On all tested files, the new assembly printer produce the same output as the old one.


compiler/src/ppasm.ml
bgregoir, line 1: Can we use this PR, to rename this file into pp_x86.ml (same for mli)

bgregoir (commented):
Thank for the PR.

compiler/src/risc_utils.ml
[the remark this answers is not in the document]
MrDaiki: I decided to factor logic of riscv and arm because they had duplicate functions that are not present in x86 target

compiler/src/ppasm.ml
MrDaiki: Indeed that's a good idea

compiler/src/risc_utils.ml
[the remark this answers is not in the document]
bgregoir: I am more in favour of a single file, but it is not a strong requirement.

compiler/src/ppasm.ml
bgregoir: I let you do that and we can certainly merge

MrDaiki: It is done

compiler/src/risc_utils.ml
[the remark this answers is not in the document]
vbgl: I’m going to merge both files.

PR #1122, opened 2025-04-11
changing .globl to .global in pp_x86 for simplification

MrDaiki:
As stated by [GAS doc](https://sourceware.org/binutils/docs/as/Global.html), `.globl` and `.global` are equivalent. Yet we use `.globl` for x86 and `.global` for arm and riscv. This PR make all headers use `.global`. This will help to factor assembly printing


vbgl (approved):
Thanks.

PR #1123, opened 2025-04-11
Add option to print export information

LeoJuguet:
Add option `-print-export-info` to print interesting information of export functions.
For the moment it only prints the alignment of the arguments for each exported function.

### DIFF
--- a/compiler/src/printer.ml
+++ b/compiler/src/printer.ml
@@ -432,8 +432,9 @@ let pp_sprog ~debug pd asmOp fmt ((funcs, p_extra):('info, 'asm) Prog.sprog) =
   let pp_opn = pp_opn pd asmOp in
   let pp_var = pp_var ~debug in
   let pp_f_extra fmt f_extra =
-    Format.fprintf fmt "(* @[<v>alignment = %s; stack size = %a + %a; max stack size = %a;@ max call depth = %a;@ saved register = @[%a@];@ saved stack = %a;@ return_addr = %a@] *)"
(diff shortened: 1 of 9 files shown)


compiler/src/printer.ml
```diff
@@ -423,8 +423,9 @@ let pp_sprog ~debug pd asmOp fmt ((funcs, p_extra):('info, 'asm) Prog.sprog) =
   let pp_opn = pp_opn pd asmOp in
   let pp_var = pp_var ~debug in
   let pp_f_extra fmt f_extra =
-    Format.fprintf fmt "(* @[<v>alignment = %s; stack size = %a + %a; max stack size = %a;@ max call depth = %a;@ saved register = @[%a@];@ saved stack = %a;@ return_addr = %a@] *)"
+    Format.fprintf fmt "(* @[<v>alignment = %s; argument alignment = [%a];@ stack size = %a + %a; max stack size = %a;@ max call depth = %a;@ saved register = @[%a@];@ saved stack = %a;@ return_addr = %a@] *)"
```
MrDaiki, line 426: Wouldn't it be interesting to output data as a standard format (for example json) ?

MrDaiki (changes requested):
PR look good to me. You need to edit `CHANGELOG.md` and add your name to AUTHOR file before we can merge this PR

compiler/src/printer.ml
eponier: We have [this PR](https://github.com/jasmin-lang/jasmin/pull/656) that has been open for long, but we could not find a principled way to implement it.

LeoJuguet: I've added an option to print the export information in JSON.

This new export prints alignment information for all arguments. Where the previous printer was limited to the relevant information (alignment <> 8) to make it easier to read.

vbgl (approved):

PR #1124, opened 2025-04-11
Be less verbose if not debug

eponier:
- do not print ` /* of uXX */` for casts if not debug
- do not print ` (defined at loc)` for vars if not debug

The goal is to hopefully get more readable error messages by default, and a more verbose one if `-debug` is given to the command line.

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -82,6 +82,8 @@ let pp_suffix fmt =
   | PVvv (ve, sz, ve', sz') -> F.fprintf fmt "_%s_%s" (string_of_velem Unsigned sz ve) (string_of_velem Unsigned sz' ve')
 
 let pp_tyerror fmt (code : tyerror) =
+  (* We do not need the extra verbosity of [debug] in this context *)
+  let pp_eptype = Printer.pp_eptype ~debug:false in
   match code with
   | UnknownVar x ->
       F.fprintf fmt "unknown variable: `%s'" x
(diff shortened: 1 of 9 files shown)


eponier:
@bgregoir finds that the ` /* of UXX */` part in the messages makes them sometimes hard to read. So I made this patch to print it only if `Glob_options.debug` is set. But I hadn't thought it would imply so many changes.

eponier:
Moreover, at some point I had to provide a value for the `debug` param I added to some functions. I was not sure what to put there in some cases. It could be `true` if we think we always want a precise message, or `false` if we consider that the additional details are never interesting, or `!Glob_options.debug` if we want to give to the user the option. I tried to make sensible choices, but I'm not sure I succeded.

eponier:
Also it feels like `Glob_options.debug` is overloaded: it is used both to make existing messages more verbose, and to print additional messages. Maybe we should decouple the two things.

eponier:
I added another commit impacting printing of vars.

compiler/src/pretyping.ml
```diff
@@ -81,7 +81,8 @@ let pp_suffix fmt =
   | PVx (szo, szi) -> F.fprintf fmt "_u%au%a" pp_wsize szo pp_wsize szi
   | PVvv (ve, sz, ve', sz') -> F.fprintf fmt "_%s_%s" (string_of_velem Unsigned sz ve) (string_of_velem Unsigned sz' ve')
 
-let pp_tyerror fmt (code : tyerror) =
+let pp_tyerror ~debug fmt (code : tyerror) =
```
vbgl, line 84: Is it ever useful to call this function with `debug:true`? IMHO these error messages hardly need the extra verbosity.

compiler/src/printFexpr.mli
```diff
@@ -1,3 +1,3 @@
-val pp_fexpr : Format.formatter -> Fexpr.fexpr -> unit
-val pp_rexpr : Format.formatter -> Fexpr.rexpr -> unit
-val pp_lexpr : Format.formatter -> Fexpr.lexpr -> unit
+val pp_fexpr : debug:bool -> Format.formatter -> Fexpr.fexpr -> unit
+val pp_rexpr : debug:bool -> Format.formatter -> Fexpr.rexpr -> unit
+val pp_lexpr : debug:bool -> Format.formatter -> Fexpr.lexpr -> unit
```
vbgl, line 3: Does this new argument have any effect? When reaching `linear`, are there any casts to `int` left?

vbgl (commented):
I’m not convinced that the added complexity (many functions now take a `~debug` argument) is worth it.

compiler/src/printFexpr.mli
eponier: No effect on the tests at least, so I removed the argument. Thanks for the remark!

eponier:
> I’m not convinced that the added complexity (many functions now take a `~debug` argument) is worth it.

I agree, I didn't thought that the change would be so pervasive. Thanks to the two comments you made, I reduced the number of changes. Maybe this is now acceptable?

vbgl (approved):

PR #1125, opened 2025-04-11
Putting .global (export) directive at top of generated assembly (for coherence)

MrDaiki:
Currently, we have two different behaviour regarding the position of `.global` directive in generated assembly sources. In x86, we put them at the top of the file. In arm and riscv, we put them just before the function label.

This PR modify the behaviour of arm and riscv assembly printer to put `.global` directives after file header like it is done in x86. This will increase readability of generated assembly and improve coherence between generated assembly for all architectures.

The doc only specify that export directive should be placed before function label, so this change shouldn't break anything (documentation of this feature is however lacking).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@
   ([PR #1114](https://github.com/jasmin-lang/jasmin/pull/1114)).
 
 - Rewriting of assembly printers
-  ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118)).
+  ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118),
+  [PR #1125](https://github.com/jasmin-lang/jasmin/pull/1125)).
 
 # Jasmin 2025.02.1 — Nancy, 2025-04-10
 
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -262,12 +262,6 @@ let pp_body fn =
     (List.map (fun x -> Dwarf x) (DebugInfo.source_positions ii.base_loc))
     i
 
-let pp_fn_head fn fd =
-  let fn = escape fn in
-  if fd.asm_fd_export then
-    [ Instr (".global", [ mangle fn ]); Instr (".global", [ fn ]) ]
-  else []
-
 let pp_fn_prefix fn fd =
   let fn = escape fn in
   if fd.asm_fd_export then
@@ -286,11 +280,10 @@ let pp_fn_pos fn fd =
 
 let pp_fun (fn, fd) =
   let fn = fn.fn_name in
-  let head = pp_fn_head fn fd in
   let pre = pp_fn_prefix fn fd in
   let body = pp_body fn fd.asm_fd_body in
   let pos = pp_fn_pos fn fd in
-  head @ pre @ body @ pos
+  pre @ body @ pos
 
 let pp_funcs funs = List.concat_map pp_fun funs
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@
 - Rewriting of assembly printers
   ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118)).
 
+- Moving export directive (`.global`) after file headers in ARM and Risc-V assembly printing
```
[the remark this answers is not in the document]
MrDaiki: To me it changes the behavior of the compiler so it should appears, but I understand why you do not agree.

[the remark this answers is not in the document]
vbgl: It can be merged with the previous entry (PR 1118).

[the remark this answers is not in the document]
MrDaiki: I did the change suggested by @vbgl

CHANGELOG.md
```diff
@@ -18,7 +18,8 @@
   ([PR #1114](https://github.com/jasmin-lang/jasmin/pull/1114)).
 
 - Rewriting of assembly printers
-  ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118)).
+  ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118))
+  ([PR #1125](https://github.com/jasmin-lang/jasmin/pull/1125)).
```
eponier, line 22:
```suggestion
  ([PR #1118](https://github.com/jasmin-lang/jasmin/pull/1118),
  [PR #1125](https://github.com/jasmin-lang/jasmin/pull/1125)).
```

MrDaiki: Thanks for the suggestion, I applied it

MrDaiki:
I think it is ready for a new review

compiler/src/pp_arm_m4.ml
```diff
@@ -301,9 +294,19 @@ let pp_data globs names =
     format_glob_data globs names
   else []
 
+let pp_fn_decl (fn,fd) =
+  let fn = escape fn.fn_name in
```
bgregoir, line 298: Can you share this code with the one of pp_riscv.ml

compiler/src/pp_arm_m4.ml
```diff
@@ -301,9 +294,19 @@ let pp_data globs names =
...
+  if fd.asm_fd_export then
+    [ Instr (".global", [ mangle fn ]); Instr (".global", [ fn ]) ]
+  else []
+  
+let pp_decls funcs =
```
bgregoir, line 303: Same comment

bgregoir (commented):
After that we can merge.

compiler/src/pp_arm_m4.ml
MrDaiki: this will be done in a later PR

eponier:
Actually, thinking of it, I wonder what behavior is clearer. On one side, arm and risc-v's behavior mixes the bodies of functions with these annotations, which is not super nice, but at least if you wonder whether the function is exported, you just have to go to the start of the function.

MrDaiki:
> Actually, thinking of it, I wonder what behavior is clearer. On one side, arm and risc-v's behavior mixes the bodies of functions with these annotations, which is not super nice, but at least if you wonder whether the function is exported, you just have to go to the start of the function.

Okay, I understand your point of view. @vbgl @bgregoir could we have your opinion on this matter ? 

My objective with this PR is to make the writing of assembly code as standard as possible so we can factor out a common interface for it.

bgregoir:
> Actually, thinking of it, I wonder what behavior is clearer. On one side, arm and risc-v's behavior mixes the bodies of functions with these annotations, which is not super nice, but at least if you wonder whether the function is exported, you just have to go to the start of the function.
I think it is not very hard to go to the beginning of the file to see if the function is exported.
The solution also allows to easily know which function are exported.

bgregoir (approved):

PR #1127, opened 2025-04-23
safety: use dune to run tests

vbgl:
The outputs of the checker on the test suite are now saved. This makes it easier to spot regression, in particular to be sure that failing tests actually fail for the expected reason (and no due to e.g., a syntactic error).


vbgl:
Small update: only the output of negative tests is recorded.

bgregoir (approved):

PR #1130, opened 2025-04-23
Factoring behaviour of assembly printer in a functor module

MrDaiki:
This PR introduce a new module `AsmTarget` which abstract traversal of assembly program during the printing pass. It has been tested for all supported architectures on libjade and common test file without provoking any change to generated assembly.

### DIFF
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -43,6 +43,8 @@ let pp_reg_address_aux base disp off scal =
       ~kind:"assembly printing"
       "the address computation is too complex: an intermediate variable might be needed"
 
+let pp_brace s = Format.asprintf "{%s}" s
+
 let pp_imm = pp_imm imm_pre
 
 let pp_register = pp_register arch
@@ -105,7 +107,6 @@ let get_IT i =
       | Some c -> [ Instr ("it", [ pp_condt c ]) ]
     end
   | _ -> []
-
 
 module ArgChecker : sig
   (* Return the (possibly empty) suffix for the mnemonic according to its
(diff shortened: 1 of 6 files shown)


compiler/src/AsmTarget.mli
```diff
@@ -0,0 +1,87 @@
...
+          type reg = Target.reg
+          type regx = Target.regx
+          type xreg = Target.xreg
+          type rflag = Target.rflag
+          type cond = Target.cond
+          type asm_op = Target.asm_op
+          val asm_of_prog :
+            (reg, regx, xreg, rflag, cond, asm_op) Arch_decl.asm_prog ->
+            PrintASM.asm_element list
+        end
```
vbgl, line 86: Why don’t you use `S` here?

compiler/src/pp_arm_m4.ml
```diff
@@ -194,119 +195,127 @@ let pp_ADR pp opts args =
...
+  type reg = Arm_decl.register
+  type regx = Arch_utils.empty
+  type xreg = Arch_utils.empty
+  type rflag = Arm_decl.rflag
+  type cond = Arm_decl.condt
+  type asm_op = arm_op
+  let headers = [ Instr (".thumb", []); Instr (".syntax unified", []) ]
+
+  let pp_data_segment_header globs name =   
+    if not (List.is_empty globs) then
```
vbgl, line 219: This condition is always true, isn’t it?

compiler/src/pp_riscv.ml
```diff
@@ -87,111 +78,102 @@ let pp_ext = function
...
+  type xreg  = Arch_utils.empty
+  type rflag = Arch_utils.empty
+  type cond  = Riscv_decl.condt
+  type asm_op = Riscv_instr_decl.riscv_op
+
+
+  (* TODO_RISCV: Review. *)
+  let headers = []
+
+  let pp_data_segment_header _ globs =
```
vbgl, line 101: Are you sure that you want to name `globs` the second argument?

compiler/src/pp_arm_m4.ml
MrDaiki: It is, I forgot to remove the redundant condition

compiler/src/AsmTarget.mli
MrDaiki: It is corrected

compiler/src/pp_riscv.ml
MrDaiki: Corrected

compiler/src/pp_arm_m4.ml
MrDaiki: I also did the change in pp_riscv, I made the same mistake

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,119 @@
...
+
+    type reg
+    type regx
+    type xreg
+    type rflag
+    type cond
+    type asm_op
+
+    val headers : asm_element list
+    val pp_data_segment_header : Obj.t list -> ((Var0.Var.var * Wsize.wsize) * BinNums.coq_Z) list -> asm_element list
```
vbgl, line 18: Why is this a function?

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,119 @@
...
+
+        let pp_function (name,decl) =
+            let headers = Target.pp_function_header name decl in
+            let body = pp_body name decl in
+            let tail = Target.pp_function_tail name decl in 
+            headers @ body @ tail
+
+    
+        let pp_functions funcs =
+            List.fold_left (fun acc (func) ->acc @ (pp_function func)) [] funcs
```
vbgl, line 81: This looks like an instance of `concat_map`.

(i.e., `let pp_functions = List.concat_map pp_function`).

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,119 @@
...
+            if decl.asm_fd_export then
+                let fn = escape name.fn_name in
+                [
+                    Instr (".global", [mangle fn]);
+                    Instr (".global", [fn])
+                ]
+            else []
+        
+        let pp_functions_decl (funcs) =
+            List.fold_left (fun acc (func) -> acc @ (pp_function_decl func)) [] funcs
```
vbgl, line 94: `concat_map`?

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,119 @@
...
+        
+        
+        let pp_data_segment_body globs names =
+        let datas = Asm_utils.format_glob_data globs names in
+        List.fold_left (fun acc data -> acc @ [(data)]) [] datas
```
vbgl, line 99: Is this the identity?

compiler/src/pp_arm_m4.ml
```diff
@@ -194,119 +195,125 @@ let pp_ADR pp opts args =
...
+        in
+        let args = pp_shift op args in
+        get_IT instr @ [ Instr (name, args) ]
+
+  let pp_function_header fn fd =
+    let fn = escape fn.fn_name in
+    if fd.asm_fd_export then
+      [
+        Label (mangle fn);
+        Label fn;
```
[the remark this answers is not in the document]
vbgl: Note that after factoring the common parts, the “export-function-header” is a constant (for each target).

compiler/src/AsmTarget.ml
MrDaiki: That is totally the case, I don't even know why I did wrote that ...

compiler/src/AsmTarget.ml
MrDaiki: It didn't needed to, I corrected it

compiler/src/pp_arm_m4.ml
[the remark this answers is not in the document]
MrDaiki: It has been factored

compiler/src/AsmTarget.ml
MrDaiki: corrected

compiler/src/AsmTarget.ml
MrDaiki: corrected

MrDaiki:
I applied your @vbgl suggestions on `concat_map`. I also factor the behaviour of function tail. The test to know if we should write a function tail is done in the functor, which simplify the signature of the module.

compiler/src/pp_arm_m4.ml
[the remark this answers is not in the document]
MrDaiki: Also done

compiler/src/AsmTarget.mli
```diff
@@ -0,0 +1,78 @@
...
+    (* Data segment header*)
+    val data_segment_header : PrintASM.asm_element list
+    
+    (* Start of the function*)
+    val function_header : PrintASM.asm_element list
+
+    (* Function return instructions 
+    It is called only on exported functions.
+    *)
+    val pp_function_tail : CoreIdent.funname -> PrintASM.asm_element list
```
vbgl, line 29:
```suggestion
    val pp_function_tail : PrintASM.asm_element list
```

This is a constant.

compiler/src/AsmTarget.mli
```diff
@@ -0,0 +1,78 @@
...
+
+    (* Pretty print of instruction*)
+    val pp_instr_r :
+      CoreIdent.funname ->
```
vbgl, line 33:
```suggestion
      Name.t ->
```

Only the field `fn_name` is used.

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,127 @@
...
+    val headers : asm_element list
+    val data_segment_header :  asm_element list
+    val function_header  : asm_element list
+    val pp_function_tail    : funname -> asm_element list
+    val pp_instr_r : funname -> (reg, regx, xreg, rflag, cond, asm_op) Arch_decl.asm_i_r -> asm_element list
+
+end
+
+
+module AsmTargetBuilder = struct
```
vbgl, line 26: What’s the purpose of this module (too much nesting)?

compiler/src/AsmTarget.ml
```diff
@@ -0,0 +1,127 @@
+open PrintASM
```
vbgl, line 1: Should all the contents of this file be moved inside `printASM.ml`?

vbgl: Same question about the contents of `asm_utils.ml` (to be moved into `printASM.ml`).

compiler/src/AsmTarget.mli
MrDaiki: I corrected it. I wasn't sure it would not cause problem with name escaping

compiler/src/AsmTarget.mli
MrDaiki: I corrected it. I wasn't sure it would not cause problem with name escaping

compiler/src/AsmTarget.mli
MrDaiki: You are right. I changed implementation in arm and it is now constant

compiler/src/AsmTarget.ml
MrDaiki: I am not sure to understand what you mean. Is it the `S` and `Make` sub modules or the fact that `AsmTarget` and `AsmTargetBuilder` are located in the same file ?

compiler/src/AsmTarget.ml
MrDaiki: I prefer it this way because each file handle different mechanism in my opinion : 
* `PrintAsm` -> final pass of compilation that produce assembly code 
* `asm_utils` -> utilitary functions used while translating `asm_prog` to `asm_element`
* `AsmTarget` -> abstract mechanism to translate `asm_prog` to `asm_element`

I can gather it in the same file if you want

compiler/src/AsmTarget.ml
vbgl: The problem is that there exists an `AsmTargetBuilder` module: its contents could be moved up (i.e., defined at the top-level of the `AsmTarget.ml` file.

compiler/src/AsmTarget.ml
vbgl: I’ve no strong opinion about this. My feeling is that when functions are scattered in many files, it is harder to find them (and to remember what is where, and to decide where to introduce new functions when the need arises, etc.).

MrDaiki: maybe we can get other feadbacks about this (@bgregoir @eponier)

compiler/src/AsmTarget.ml
MrDaiki: You are right, the nesting doesn't serve anything, I will remove it.

MrDaiki: It is done

vbgl (approved):

compiler/src/AsmTarget.ml
eponier: I was on vacation, sorry. I agree with @vbgl in general. In the future, if we feel that the split in 3 files is unneeded/unnatural, we should not hesitate to merge some of them.

PR #1133, opened 2025-04-25
Changing syntax for variable initialisation at declaration

MrDaiki:
# Issue

Currently, the syntax of jasmin authorize initialisation of the form : 
```
reg u64 x=1 y; // another variable declaration after initialisation
reg u64 x y = 1; // another variable declaration before initialisation of a variable

```
This syntax doesn't look well and could be confusing. When we write :
``` 
reg u64 x y = 1;
``` 

Does it mean we want to assign to `y` or both `x` and `y` ?

# Proposition 

This PR change the way syntax of initialisation is handled : 
``` 
reg u64 x y = 1;
``` 
is now equivalent to assigning 1 to both `x` and `y`. To assign only to `y`, we now write : 
```
reg u64 x, y = 1;
```
We also enforce the usage of a comma after an initialisation : 
``` 
reg u64 x=1 y; // This is now invalid
reg u64 x=1, y; // This is still valid
```


eponier:
In my opinion, this PR goes in the right direction, but I would be even stricter and reject
```
reg u64 x y = 1;
```
because it is too ambiguous.

bgregoir:
@eponier, I think the idea is to first deprecate the syntax, and remove it in the following release.

bgregoir:
Some files need to be patched (they use the syntax).

eponier:
From what I read, the syntax I quote is accepted as "initialize both variables to 1". Whereas the first message contains

> This syntax doesn't look well and could be confusing. When we write :
> 
> ```
> reg u64 x y = 1;
> ```
> 
> Does it mean we want to assign to `y` or both `x` and `y` ?
I completely agree with the first message, so I suggest to forbid this "multiple initializations at once" syntax.

MrDaiki:
> From what I read, the syntax I quote is accepted as "initialize both variables to 1". Whereas the first message contains
> 
> > This syntax doesn't look well and could be confusing. When we write :
> > ```
> > reg u64 x y = 1;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does it mean we want to assign to `y` or both `x` and `y` ?
> 
> I completely agree with the first message, so I suggest to forbid this "multiple initializations at once" syntax.

okay, so each variable initialisation should look like one of the following : 
```
reg u64 x=1, y z; // still valid
reg u64 x y=1, z; // invalid
```

which mean that if you have an initialisation, it must concern one variable at most. I do agree with your point, but I wasn't sure of what to do after we discuss it with @bgregoir.

If we valid this design, I can remove this PR and we will resolve it in next version. I will the reopen the PR (or a new one)

eponier:
This is my point of view, maybe others have different views.

MrDaiki:
> This is my point of view, maybe others have different views.

Let's ask for them : @vbgl @bgregoir what do you think about this issue ?

vbgl (changes requested):
I think this change goes in the wrong direction.

I do not deny that the current syntax can be abused to write obfuscated code. However, this fact is not enough to make the syntax more complex.

There are actual issue with the current syntax. Please tackle them instead. To me, the main limitation of this feature is that it is not possible to use it with a RHS that produces multiple values (e.g., `#set0()`).

vbgl:
I still fail to see how this proposal solve any actual issue.

vbgl:
What about the following: declaration with initialization is restricted to a single variable (i.e., one cannot declare many variables when initializing some of them).

vbgl:
> Maybe we could have two simple rules:
> 
>     * declarations with spaces as separators: no definition, just declaration
> 
>     * declarations with commas as separators: definitions are allowed

If some people see this as an improvement, I would be OK with that change.

I think there is no need for a deprecation because it is already possible to write compatible code (just put commas everywhere).

eponier:
> What about the following: declaration with initialization is restricted to a single variable (i.e., one cannot declare many variables when initializing some of them).

That would be fine too.

vbgl:
I’ve tried something there: https://github.com/jasmin-lang/jasmin/pull/1185. Feedback welcome.

compiler/src/parser.mly
```diff
@@ -448,12 +448,21 @@ storage:
...
+%inline noninitdecl:
+| v=var { v,None }
+
+%inline initdecl:
+| v=var EQ e=pexpr { (v, Some e) }
+
 %inline decl:
-| v=var { v, None }
-| v=var EQ e=pexpr { v, Some e }
+| vs=separated_nonempty_list(empty,loc(noninitdecl)) { vs }
```
vbgl, line 458: FWIW, there is a `nonempty_list` combinator…

PR #1142, opened 2025-05-04
Itrees4

bgregoir:


bgregoir:
Hi, I will squash all commit in one, but excepted that I think it is ready for review.
In this PR, we mostly do two things: 
- split psem into psem_core (all lemma on sem_pexpr and write_lval) and psem only the lemma on instruction. 
- Add the proof of some transformation based on itree semantics

proofs/lang/psem_core.v
```diff
@@ -90,3 +99,1356 @@ Instance sCP_stack : semCallParams (pT := progStack) :=
...
+
+Lemma use_memP_eq_on wdb gd s1 s2 e:
+  ~~use_mem e ->
+  evm s1 =[read_e e] evm s2 ->
+  sem_pexpr wdb gd s1 e = sem_pexpr wdb gd s2 e.
+Proof.
+  by move=> h1 h2; rewrite (use_memP wdb (s2:= with_vm s2 (evm s1)) _ gd h1) //; apply: eq_on_sem_pexpr.
+Qed.
+
+(* FIXME this is close to write_var_spec but less specified *)
```
vbgl, line 749: Is this a TODO for later or to be fixed in this PR?

proofs/lang/psem_core.v
```diff
@@ -90,3 +99,1356 @@ Instance sCP_stack : semCallParams (pT := progStack) :=
...
+  case: l => /=.
+  + by move => _ ? /write_noneP [].
+  + by rewrite /write_var; t_xrbindP => > /set_varP [].
+  + by t_xrbindP.
+  + by move => al aa ws [] [vt vn] /= _ e; apply: on_arr_varP => n t hty /= ?; t_xrbindP.
+  by move => aa ws len [] [vt vn] /= _ e; apply: on_arr_varP => n t hty /= ?; t_xrbindP.
+Qed.
+
+(* MOVE THIS *)
+Section Expr.
```
vbgl, line 1320: When will this be moved? And to where?

bgregoir: psem_core is a split of the original psem. The comment was there in psem. 
I think it can be remove. I don't see why and where can move this.

proofs/lang/psem_core.v
bgregoir: it is not a comment introduced by this PR. it was there in psem.

bgregoir: so it will not be fixed by this PR.

vbgl (approved):

PR #1143, opened 2025-05-05
Minor cleaning

vbgl:
- Typo & formatting
- Ensure wint annotations can be printed (in a syntactically valid way)
- Less warnings in the test suite


compiler/tests/success/x86-64/test_set0.jazz
bgregoir, line 1: Why do you delete this file ?

bgregoir (commented):
Excepted that I think it is ok

compiler/tests/success/x86-64/test_set0.jazz
vbgl: Because I’ve move its contents to the `set0.jazz` file next to it.

bgregoir (approved):

PR #1146, opened 2025-05-06
Remove indirection

vbgl:
Does anyone remember the use for exposing these functions from `Ssa` as part of the `Regalloc` module?

### DIFF
--- a/compiler/src/compile.ml
+++ b/compiler/src/compile.ml
@@ -285,9 +285,9 @@ let compile (type reg regx xreg rflag cond asm_op extra_op)
     Prog.V.clone x
   in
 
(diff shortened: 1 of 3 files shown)


eponier:
Digging into the history, apparently initially there was only `Regalloc.split_live_ranges` doing several things at the same time, and exported. Then in c5929b8880da0c129b474f2704a25047c387c98b, I split it into 3 and probably decided to export the three of them because the original one was exported.

vbgl:
Thanks for digging up history. What do you then think of the changes of this PR?

eponier:
No strong feeling, probably better for comprehension if we remove functions with the same name and which do the same thing.

eponier (approved):

PR #1150, opened 2025-05-07
Regalloc: do not blindly remove renaming assignments

### LINKED ISSUE #483: Regalloc conflict produces internal error when function is inlined

vbgl:
Fixes #483.

### DIFF
--- a/compiler/src/regalloc.ml
+++ b/compiler/src/regalloc.ml
@@ -212,13 +212,24 @@ let collect_equality_constraints_in_func
       is_move_op
       ~(with_call_sites: (funname -> ('info, 'asm) func) option)
       (msg: string)
-      (int_of_var: var_i -> int option)
-      (check_safe_copy_elision: L.i_loc -> var_i -> var_i -> L.i_loc list)
+      (tbl: int Hv.t)
+      (nv: int)
+      (get_live_out: 'info -> Sv.t)
       copn_constraints
       (s: ('info, 'asm) collect_equality_constraints_state)
       (f: ('info, 'asm) func)
     : unit
   =
+  (* This proceeds in two passes over the instructions of the function f
+     The first pass:
+       - collects constraints from opn (architecture-specific)
+       - marks as equal variables that are φ-congruent
+       - remembers the set of “renaming assignments” introduced by inlining
+       - marks as friends variables involved in other renaming-like instructions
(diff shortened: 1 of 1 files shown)


compiler/src/prog.ml
```diff
@@ -503,6 +503,12 @@ and spilled_c s c =  List.fold_left spilled_i s c
 
 let spilled fc = spilled_c Sv.empty fc.f_body
 
+let assigns = function
+  | Cassgn (x, _, _, _) -> written_lv Sv.empty x
+  | Copn (xs, _, _, _) | Csyscall (xs, _, _) | Ccall (xs, _, _) ->
+      List.fold_left written_lv Sv.empty xs
+  | Cif _ | Cfor _ | Cwhile _ -> Sv.empty
```
bgregoir, line 510: where the assign of the for loop variable is taken into account ?

bgregoir (commented):
Still have to understand the second commit

compiler/src/prog.ml
vbgl: The assignments to the for loop variable is not taken into account. Should this be changed? documented? raise an error?

The only use of this function is in register allocation, after for loops have been removed, so it does not matter much.

bgregoir: Ok. Maybe in that case, an assert false for the case of Cfor is better. It will prevent bad usage in the future.

PR #1151, opened 2025-05-07
Load store expr

bgregoir:
The idea of this PR is to use an single expression instead of a pair var * expr to represent memory access.
It allows to use type uiXX for pointer and the extraction can extract pointers to int.

### DIFF
--- a/proofs/arch/asm_gen.v
+++ b/proofs/arch/asm_gen.v
@@ -225,9 +225,6 @@ Definition addr_of_fexpr (rip: var) ii sz (e: fexpr) :=
   | None => Error (E.error ii (pp_box [::pp_s "not able to assemble address :"; pp_fe e]))
   end.
 
-Definition addr_of_xpexpr rip ii sz v e :=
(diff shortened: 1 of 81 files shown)


proofs/arch/asm_gen.v
```diff
@@ -225,8 +225,8 @@ Definition addr_of_fexpr (rip: var) ii sz (e: fexpr) :=
   | None => Error (E.error ii (pp_box [::pp_s "not able to assemble address :"; pp_fe e]))
   end.
 
-Definition addr_of_xpexpr rip ii sz v e :=
-  addr_of_fexpr rip ii sz (Fapp2 (Oadd (Op_w sz)) (Fvar v) e).
+Definition addr_of_xpexpr rip ii sz e :=
+  addr_of_fexpr rip ii sz e.
```
vbgl, line 229: This function now looks like the identity.

vbgl (approved):
Well done. Thanks.

PR #1154, opened 2025-05-12
CI: disable coq-master

vbgl:
The first commit restores the ability to run the `coq-master` job (on CI or at home) and witness it failing during extraction (with the very kind suggestion: try writing your program in Haskell instead).

The second commit just disable the job as it is probably pointless as-is.


eponier:
Why don't we use master for `elpi` version like we do for any other package?

eponier:
I was fool when I thought I started understanding how nix works...

eponier (approved):

eponier:
For the record, the failure of coq-master was expected after https://github.com/math-comp/math-comp/pull/1256 (see https://github.com/jasmin-lang/jasmin/pull/1093 for details).

vbgl:
Thanks.

To be pedantic, your questions are not related to nix (the tool, nor the language) but to the nice program that build the derivation for coq-elpi (crafted mostly by Pierre & Cyril).

eponier:
Oh, so it is a choice of the derivation not to pick a version by default?

vbgl:
I’d rather say it’s a choice of the programmers.

eponier:
> I’d rather say it’s a choice of the programmers.

Thanks, I still have some hope then!

PR #1156, opened 2025-05-15
Changing dune version to 3.7

MrDaiki:
This PR replace version of dune used by the compiler from 3.0 to 3.7. This will allow the usage of new features such as qualify subdirs. I also add to make dune ignore warning 67 because it was detected with this new version.


MrDaiki:
A question before review : do I need to update changelog (wasn't sure for this change)

vbgl (approved):
I see no need for an entry in the CHANGELOG.

PR #1158, opened 2025-05-16
Typing: check length non-negative

### LINKED ISSUE #1106: Assertion fails on subarray of length 0
Trying to compile

```
export fn main () -> reg u32 {
  stack u32[1] a b;
  a[0] = 1;
  b[0] = 2;
  a[0:0] = b[0:0];
  reg u32 res = b[0];
  return res;
}
```
returns
```
Fatal error: exception File "src/coreConv.ml", line 3, characters 2-8: Assertion failed
```

This may be fixed when we allow something else than integers as array length.

vbgl:
Fixes #1106.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@
   ([PR #1147](https://github.com/jasmin-lang/jasmin/pull/1147);
   fixes [#1134](https://github.com/jasmin-lang/jasmin/issues/1134)).
 
+- Check that array lengths are strictly positive
+  ([PR #1158](https://github.com/jasmin-lang/jasmin/pull/1158);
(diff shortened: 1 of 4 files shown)


vbgl:
This also fixes crashes in the following examples:

~~~
param int T = 0;
export fn test(reg ptr u8[10] t) {
  reg u8[0] r = #copy(t[0:T]);
}
~~~

~~~
param int T = 0;
inline
fn test(reg u8[T] t) {
  reg u8[T] r = t;
}
~~~

~~~
param int N = 0;
fn test(reg ptr u8[10] p, reg u64 msf) -> reg ptr u8[10] {
  reg ptr u8[N] r = #protect_ptr(p[0:N], msf);
  p[0:N] = r;
  return p;
}
~~~

bgregoir (approved):

PR #1159, opened 2025-05-16
Fix TODO: prevent register allocation from ACR_subset by adding confl…

zhengjieyu:
Modify register allocation to avoid assigning registers from ACR_subset

### DIFF
--- a/proofs/arch/arch_decl.v
+++ b/proofs/arch/arch_decl.v
@@ -258,7 +258,7 @@ Variant addr_kind : Type :=
 (* -------------------------------------------------------------------- *)
 (* Argument description.
  * An argument may be either implicit or explicit.
- * Registers may be constrained to a specific one or a subset
+ * Registers may be constrained to a specific one or to avoid a given subset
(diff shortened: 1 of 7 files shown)


proofs/arch/arch_decl.v
```diff
@@ -288,7 +288,7 @@ Definition check_oreg or ai :=
   | ACR_exact _, _       => false
   | ACR_vector x, XReg r => x == r ::>
   | ACR_vector _, _      => false
-  | ACR_subset s, Reg r  => r \in (s : seq ceqT_eqType)
+  | ACR_subset s, Reg r  => negb (r \in (s : seq ceqT_eqType))
```
vbgl, line 291: This change looks wrong. In particular the current use of ACR_subset does not assume this meaning.

vbgl:
This does not seem to fix #1112, which is unexpected.

vbgl (approved):

vbgl:
Fixes #1112.

PR #1170, opened 2025-05-27
Safety assert2

lyonel2017:


vbgl:
Why has the extraction for CT been changed in an incompatible way?

PR #1175, opened 2025-06-03
Support mathcomp 2.4

eponier:
With mathcomp 2.4, we get a lot of warnings of the form
```
Notation ringType is deprecated since mathcomp 2.4.0.
Try pzRingType (the potentially-zero counterpart) first, or use nzRingType instead.
```
This is not possible to solve without requiring 2.4.0, so for the moment, warnings are silenced.


eponier:
Oops, I hadn't checked extraction. This should be ok now.

PR #1176, opened 2025-06-03
Safety: reduction congr → num

vbgl:


vbgl:
I’ve just added an example (test case) and slightly improved the readability of the `get_constr` function in the congruence domain.

bgregoir (approved):
This is nice.

PR #1177, opened 2025-06-03
sarr of positive -> sarr of wsize & positive

eponier:


bgregoir:
Hi Jean-Christophe,
Thank for this. I will take a look to this more carefully. 
> Interrogations that I have:
> 
> * I used one polymorphic type `stype` instantiated into `atype` and `ctype`. Maybe two distinct types would be ok and simpler?
My personal intuition is that two types will be better, maybe in the future we will want to change the definition of atype
to make it more expressive/flexible and it seems simpler if this will not impact ctype. 

> * arr_init: I replace positive with wsize * positive, do we want that? Do we need in the long run to have arr_init polymorphic in the length of arrays? If it is used only on stack arrays whose lengths should be known otherwise we don't know how to allocate the stack, maybe not.
I think arr_init are added to all array (reg ptr/ stack ptr/stack) so it is cleary better to have wsize * positive, or any other datatype that will abstractly encode the length of array.

> * randomebytes: same question, but there I left positive
   Same answer here.
> * same question for copy? actually, copy already takes a wsize, but it is more the size to use for the copy
  Same
> * for architecture instructions, I used `ctype` because we have only simple types there and no arrays, but in pseudo-ops we have `copy` where I think we want `atype` because we want it to be polymorphic in the length of arrays in the future. 
    Maybe we can introduce core_type (bool + word) for machine instructions, they can be easily injected intoa atype (and ctype).
This means that in arch_extra.v, I have to inject instructions described with `ctype` in instructions described with `atype`, so I had to fight a bit with dependent types. It works, but I am not sure this is the right way to do it.

> * linear: should I use `atype` or `ctype` for the types of function? At that time, no array anymore, so the choice does not matter much, but maybe `atype` is more consistent with the idea `atype` = syntax?
  Or core_type at this point :-)

proofs/arch/arch_extra.v
```diff
@@ -296,23 +296,70 @@ Qed.
 
 HB.instance Definition _ := hasDecEq.Build extended_op extended_op_eq_axiom.
 
+Lemma to_atypeK : cancel to_atype to_ctype.
+Proof. by case. Qed.
+
+(* We don't use the fact that we know is_not_sarr on tin and tout.
+   And it probably cannot compute due to opaque terms, is this a problem? *)
```
bgregoir, line 303: I suspect that it will be a problem during the proof, but we will see.

proofs/lang/type.v
```diff
@@ -187,96 +223,248 @@ Arguments Mt.set P m%_mtype_scope k v.
...
+
+Lemma subtypeaEl ty ty' :
+  subatype ty ty' →
+  match ty return Prop with
+  | sword sz => ∃ sz', ty' = sword sz' ∧ (sz ≤ sz')%CMP
+  | _        => convertible ty ty'
+  end.
+Proof. exact: subtype_genEl. Qed.
+
+Lemma toto (ty1 ty2:ctype) : convertible_gen eq_op ty1 ty2 -> ty1 = ty2.
```
bgregoir, line 418: really this name ?

proofs/lang/type.v
```diff
@@ -187,96 +223,248 @@ Arguments Mt.set P m%_mtype_scope k v.
...
+
+Lemma convertible_trans t1 t2 t3 :
+  convertible t1 t2 -> convertible t2 t3 -> convertible t1 t3.
+Proof.
+  apply convertible_gen_trans.
+  move=> [??] [??] [??].
+  by move=> /eqP -> /eqP ->.
+Qed.
+
+Lemma test ty1 ty2 : convertible ty1 ty2 -> to_ctype ty1 = to_ctype ty2.
```
bgregoir, line 363: Not a better name ?

proofs/lang/type.v
eponier: Just a draft, I didn't take the time to clean anything, sorry. But of course, the name will change.

proofs/lang/type.v
eponier: Idem.

PR #1178, opened 2025-06-03
Various improvements to the printer

vbgl:
The purpose of this PR is to improve the readability of the output of the printer, without introducing any ambiguity.


compiler/src/printCommon.ml
```diff
@@ -17,6 +17,11 @@ let pp_aligned fmt =
   function
   | Memory_model.Aligned ->
      Format.fprintf fmt "#aligned "
+  | Unaligned -> ()
```
bgregoir, line 20: I think the function pp_aligned is not used anymore.
So it can be removed.

bgregoir (approved):
Excepted the small comment all seems ok, and the result looks nicer

compiler/src/printCommon.ml
vbgl: I think you are wrong.

vbgl: `pp_mem_access`

PR #1190, opened 2025-06-06
SCT: only consider live variables at the end of for loops

### LINKED ISSUE #1189: SCT checker reaches an assert false with an empty for loop
I'm getting an assert false from the SCT checker from `main` with this program
```
inline fn foo() {
  inline int j;
  for j = 1 to 1 {}
}
```
The one from the 2024.07.2 release doesn't.

vbgl:
Fixes #1189.


sarranz (approved):
Looks good to me. Thanks a lot!

bgregoir (approved):

PR #1193, opened 2025-06-11
readthedocs: add graphviz as a dependency

eponier:


eponier:
I killed the CI, since the change I did impacts only readthedocs.

PR #1194, opened 2025-06-12
Minor cleaning

vbgl:
Côme, can you please remind us the purpose of the RISC-V test case? (ping @clebreto)

### DIFF
--- a/compiler/entry/commonCLI.ml
+++ b/compiler/entry/commonCLI.ml
@@ -81,14 +81,12 @@ let parse_and_compile (type reg regx xreg rflag cond asm_op extra_op)
   let prog =
     if pass <= Compiler.ParamsExpansion then prog
     else
-      let module E = struct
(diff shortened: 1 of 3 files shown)


clebreto:
Hi @vbgl, could you provide more context ?
We implemented the RISC-V test case in the same fashion as the test case for armv7m.
We hoped to ensure the availability of the expected instructions (intrinsics) within the compiler, and the availability of the expressions introduced by the lowering part within the compiler.

That said, I don't feel like it provides any additional information to you.

clebreto:
If the test is not triggering a warning anymore, I assume it is fair to move it to success indeed.

A final word from @eponier would likely close the discussion.

vbgl:
I see. Thank you very much. This change is thus wrong. Let me fix that.

eponier (approved):
Nice comments!

PR #1199, opened 2025-06-13
fixing edge case of copy in dead variables checks

MrDaiki:
Fixing a bug in linting pass to account the case of `#copy` opperator

### DIFF
--- a/compiler/linter/Analysis/Liveness/LivenessAnalyser.ml
+++ b/compiler/linter/Analysis/Liveness/LivenessAnalyser.ml
@@ -67,10 +67,16 @@ module LivenessDomain : BackwardAnalyser.Logic with type domain = Sv.t = struct
     (_ : Location.i_loc)
     (lvs : lvals)
     (_ : E.assgn_tag)
-    (_ : 'asm Sopn.sopn)
+    (op : 'asm Sopn.sopn)
     (exprs : exprs)
     (domain : domain) =
-    Annotation (live_assigns domain lvs exprs)
+    match op with
+    | Opseudo_op (Ocopy _) ->
+      (*We do not kill copied variables in that case since copy is substituted with a loop, and thus not a real assignment*)
+      Annotation (Sv.union  (List.fold_left vars_lv domain lvs) (Prog.vars_es exprs))
+    | Oslh _ | Oasm _ | Opseudo_op _ ->
+      Annotation (live_assigns domain lvs exprs)
+
 end
 
 include BackwardAnalyser.Make (LivenessDomain)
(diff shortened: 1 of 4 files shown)


compiler/linter/Analysis/Liveness/LivenessAnalyser.ml
```diff
@@ -68,10 +68,17 @@ module LivenessDomain : BackwardAnalyser.Logic with type domain = Sv.t = struct
...
     (exprs : exprs)
     (domain : domain) =
-    Annotation (live_assigns domain lvs exprs)
+    match op with
+    | Oslh _
+    | Oasm _ -> Annotation (live_assigns domain lvs exprs)
+    |   Opseudo_op po ->
+      match po with
+      | Pseudo_operator.Ocopy _ -> Annotation (Sv.union domain (Prog.vars_es exprs)) (* edge case : as copy is replace by a loop, assigned variable should not be killed, we just make right variables live*)
+      | _ -> Annotation (live_assigns domain lvs exprs)
```
vbgl: Please do not repeat the common case.

vbgl (commented):
Can you please add non-regression test-cases and make sure they are part of the `tests/warnings.t` suite. You can use `dune promote` to fix the expected output.

I think that you forgot to mark as alive the dependencies of the l-values in the `#copy` case.

vbgl:
In the following example, the linter reports two issues, one of them being wrong:
~~~
export fn load(reg ptr u32[1] a) {
  stack u32[1] q;
  reg ptr u32[1] p = q;
  p = #copy_32(a);
}
~~~

compiler/src/prog.ml
```diff
@@ -320,6 +320,17 @@ let written_lv s =
...
+    Lnone (_, _)
+  | Lvar _ -> None
+  | Lmem (_, _, _, e)
+  | Laset (_, _, _, _, e)
+  | Lasub(_, _, _, _, e) -> Some e
+
+let lv_exprs : lvals -> exprs =
+  List.filter_map lv_expr
+
+
```
vbgl: I think this function is not necessary. Please remove it.

compiler/tests/fail/linter/dead_variables.jazz
```diff
@@ -87,3 +87,22 @@ export fn bug_1214(reg ptr u8[4] p) -> reg ptr u8[4] {
   q[0] = v;
   return p;
 }
+
+export fn explicit_copy() -> reg u8 {
+    stack u8[8] pk;
+    stack u8[8] tpk;
+    tpk=#copy(pk);
```
vbgl: The `pk` array is not initialized (so this program has no behavior). You can take the same example as below and make the copy operator explicit.

PR #1206, opened 2025-06-17
Prepare release 2025.06.0

vbgl:


eponier:
I don't want to delay this PR, but I have to go. I'll push something later tonight.

eponier:
I found typos in old parts of the Changelog, I pushed them in a separate commit.

eponier:
I removed `zquot` and `zrem` which are jargon. I used "signed division and modulo". Maybe it would be more precise to write that it is division with rounding towards zero, and that it corresponds to the signed division...

PR #1210, opened 2025-06-19
Improvements to the opam files

eponier:
My initial goal was to call `dune install` in `compiler/jasmin.opam` to install both the executables and the library.

I realized that this file was broken syntactically (and also the `cp ../eclib/*.ec` cannot work because the build is done in a sandbox where only `compiler` and its descendants can be seen). So I had two questions: why does it work when we publish it on opam-repository? And why was it not detected by the CI?

For the first question, this `compiler/jasmin.opam` file and the one we distribute for each Jasmin release in opam-repository were quite different. And also the one we distribute in opam-repository is considered at the same level as `compiler` and `eclib`, not inside `compiler`, so here the `cp eclib/*` is valid. Since it is not possible to make `compiler/jasmin.opam` work, I decided to make it nearly identical to the one distributed in opam-repository.

For the second question, in the CI, we just checked that the dependencies were ok, and then we called the `build` section by hand. I changed the job to really try an opam installation. This is hacky in two ways:
- to reproduce the opam installation, I need the file to be at the root instead of inside `compiler`, so I copy it (actually, I tried to have it at the root of the repo, but then dune explains that it does not find the package definition...)
- it messes with the opam switch that is cached, so I call `opam remove` just after the installation, but in some unlikely situation, this may break

I made several small commits, so that we can drop some of them if they are debatable.

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -76,6 +76,7 @@ $(DISTDIR).tgz: MANIFEST
 	../scripts/distribution $(DISTDIR)/compiler $<
 	cp -riv ../eclib $(DISTDIR)/
 	cp ../Makefile.compiler $(DISTDIR)/Makefile
+	cp ../jasmin-compiler.opam $(DISTDIR)/
 	tar -czvf $(DISTDIR).tgz --owner=0 --group=0 $(DISTDIR)
 	$(RM) -r $(DISTDIR)
 
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -71,6 +71,7 @@ install:
 uninstall:
 	$(RM) $(DESTDIR)$(BINDIR)/jasminc
 	$(RM) $(DESTDIR)$(BINDIR)/jasmin2tex
+	$(RM) $(DESTDIR)$(BINDIR)/jasmin-ct
 	$(RM) $(DESTDIR)$(BINDIR)/jasmin2ec
 
 # --------------------------------------------------------------------
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -22,6 +22,7 @@ DESTDIR  ?=
 PREFIX   ?= /usr/local
 BINDIR   := $(PREFIX)/bin
 INSTALL  ?= install
+DUNE_OPTS ?= --prefix $(PREFIX)
 
 # --------------------------------------------------------------------
 DISTDIR ?= jasmin-compiler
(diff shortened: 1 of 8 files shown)


compiler/Makefile
```diff
@@ -22,6 +22,7 @@ DESTDIR  ?=
 PREFIX   ?= /usr/local
 BINDIR   := $(PREFIX)/bin
 INSTALL  ?= ../scripts/install-sh
+DUNE_OPTS ?= --prefix $(DESTDIR)$(BINDIR)
```
vbgl: This looks wrong. I’d expect `PREFIX` instead of `BINDIR`

compiler/jasmin.opam
```diff
@@ -1,24 +1,24 @@
...
+license: "MIT"
 homepage: "https://github.com/jasmin-lang/jasmin"
 bug-reports: "https://github.com/jasmin-lang/jasmin/issues"
-synopsis: "High-Assurance and High-Speed Cryptography"
-license: "MIT"
+dev-repo: "git+https://gitlab.com/jasmin-lang/jasmin-compiler.git"
 
 build: [
-  make "build"
+  [make "-C" "compiler"]
```
vbgl, line 16: How is this supposed to work? Users shall move this file on directory up before using it?

compiler/Makefile
eponier: Indeed

compiler/jasmin.opam
eponier: I'd say there are two reasons to have an opam file in a repository:
- to have some installation procedure of the dev version (either `opam install --deps-only` if we just want a working state, or `opam install` if we want to try out the dev version of Jasmin)
- to store the opam file that we'll publish on opam-repository when we make a release

Of course, it is better when it fills both purposes at once. But the file `compiler/jasmin.opam` has two constraints:
- it cannot refer to `eclib` that is outside its scope, as mentioned above
- it does not make sense if we don't run `make CIL` before

At the current location, it fits only the second reason, and this is sad. But I tried to move it up and have `jasmin.opam` and `jasmin-compiler.opam` at the root of the repository, and I think it would be a better place indeed. But then dune complains that we don't define any package. I tried to move `dune-project` up also, but then it complains that it does not know what package the dune targets belong to. And in some sense, they belong to both, so I did not know what to pick, and reverted these changes.

I also wanted to push the opam file to the gitlab repo where we store extraction. I think this is where it makes more sense, because there it fulfills the two reasons I mentioned at the top of this comment (except that it does not make really sense to develop on the extraction, but it makes sense if what we want is to try Jasmin's main branch).

eponier: I may have an idea, let me try something.

compiler/Makefile
eponier: Fixed

compiler/Makefile
```diff
@@ -71,8 +72,15 @@ install:
...
 	$(RM) $(DESTDIR)$(BINDIR)/jasmin2tex
+	$(RM) $(DESTDIR)$(BINDIR)/jasmin-ct
 	$(RM) $(DESTDIR)$(BINDIR)/jasmin2ec
 
+install-full:
+	dune install $(DUNE_OPTS)
+
+uninstall-full:
+	dune uninstall $(DUNE_OPTS)
+
```
vbgl: Why doesn’t this just replace the current `install` and `uninstall` targets?

eponier: I didn't know whether we still wanted a lightweight target with just the binaries. If we agree this is not needed, I can do the change inplace indeed.

eponier:
The alternative implementation in https://github.com/jasmin-lang/jasmin/pull/1220 was merged, let's close this one.

PR #1213, opened 2025-06-23
Linter: include subdirs unqualified

vbgl:
This patch has been applied for the Debian packages of release 2025.06.0 so as to be able to build the Jasmin compiler using dune 2 (latest version available on Debian stable), so why not upstreaming it?


eponier (approved):
The readability is not worse.

NB: Debian 13 is about to be released, but it will be easy to revert this commit at some point if we feel this is the right move

vbgl:
You seem a bit optimistic about Debian release schedule, but you are right. Thanks.

PR #1219, opened 2025-06-25
SCT: fix use of liveness information in for loops

### LINKED ISSUE #1218: SCT & for loop: failed assertion
The following program makes the SCT checker to crash:

~~~
export fn f(reg u32 x) -> reg u32 {
  x = x;
  inline int a b;
  a = 0;
  for b = a to 1 {
    x += 1;
  }
  return x;
}
~~~

When looking for a fixed point for the for loop, `a` is bound in the initial environment (`venv1`) but not in the environment after a loop iteration (`venv'`).

vbgl:
Fixes #1218.


eponier (approved):
I don't understand the fix, but I trust you.

vbgl:
You should not: I have also introduced the bug…

PR #1220, opened 2025-06-25
Improvements to the opam files, second try

eponier:
Alternative to https://github.com/jasmin-lang/jasmin/pull/1210

In this PR, `compiler/jasmin.opam` is distributed with the extracted OCaml and only makes sense there, in the same spirit as `compiler/Makefile.compiler`.

### DIFF
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -76,6 +76,7 @@ $(DISTDIR).tgz: MANIFEST
 	../scripts/distribution $(DISTDIR)/compiler $<
 	cp -riv ../eclib $(DISTDIR)/
 	cp ../Makefile.compiler $(DISTDIR)/Makefile
+	cp jasmin.opam $(DISTDIR)/
 	tar -czvf $(DISTDIR).tgz --owner=0 --group=0 $(DISTDIR)
 	$(RM) -r $(DISTDIR)
 
--- a/compiler/Makefile
+++ b/compiler/Makefile
@@ -22,6 +22,7 @@ DESTDIR  ?=
 PREFIX   ?= /usr/local
 BINDIR   := $(PREFIX)/bin
 INSTALL  ?= install
+DUNE_OPTS ?= --prefix $(PREFIX)
 
 # --------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


compiler/Makefile
```diff
@@ -80,7 +74,9 @@ $(DISTDIR).tgz: MANIFEST
 	$(RM) -r $(DISTDIR)
 	../scripts/distribution $(DISTDIR)/compiler $<
 	cp -riv ../eclib $(DISTDIR)/
+	mkdir $(DISTDIR)/scripts && cp ../scripts/install-sh $(DISTDIR)/scripts/
```
vbgl: I think the correct fix is to stop relying on `install-sh` to create a directory & copy a few files.

compiler/Makefile
```diff
@@ -21,7 +21,7 @@ CHECKCATS ?= \
 DESTDIR  ?=
 PREFIX   ?= /usr/local
 BINDIR   := $(PREFIX)/bin
-INSTALL  ?= ../scripts/install-sh
+DUNE_OPTS ?= --prefix $(PREFIX)
```
vbgl, line 25: What about using `DUNE_INSTALL_PREFIX` instead?

vbgl: (this might not work with dune 2, so maybe for later)

compiler/Makefile
eponier: Use `cp` then? Or `install`?

compiler/Makefile
eponier: The idea was to give full flexibility. For instance, in nixpkgs, both `--prefix` and `--libdir` are used, so I wanted to give a way to override the full set of options. But I'm not sure I understood what you meant, so this not be incompatible.

compiler/Makefile
eponier: I rebased on latest main (and thus on top of https://github.com/jasmin-lang/jasmin/pull/1222).

eponier:
So be it!

vbgl (approved):

PR #1222, opened 2025-06-27
Exit script/install-sh

vbgl:


scripts/distribution
```diff
@@ -5,8 +5,7 @@ import sys, os, re, itertools as it, subprocess as sp
 import glob, fnmatch
 
 # --------------------------------------------------------------------
-MYROOT  = os.path.dirname(__file__)
-INSTALL = os.path.join(MYROOT, 'install-sh')
+INSTALL = os.path.join('install')
```
eponier: If I'm not mistaken, the call to `os.path.join` is not needed, this just returns `'install'`.

vbgl: Right. Fixed.

Issue #1223, opened 2025-07-02
Too many dead assignment warnings


vbgl:
The following example triggers a lot of warnings (from the dead-assignment linter check).

~~~
export fn flags() -> reg u64 {
  reg u64 r;
  ?{}, r = #set0();
  return r;
}

export fn only_one_live_result(reg u64 x) -> reg u64 {
  reg u64 z = 0;
  x, z = #swap(z, x);
  return x;
}
~~~

In the `flags` function, there are warnings about flags variables (`_of_` etc.) being needlessly assigned.

In the other function, there is a warning about the assignment to `z` being dead which is only partially true, and I can imagine programmers willing to write `z` rather than `_` there.

PR #1224, opened 2025-07-02
Changing behaviour of dead variables checker for multiple lvalues assignments + removing implicit architecture variable from dead variables check

MrDaiki:
This PR address issues raised in [#1223](https://github.com/jasmin-lang/jasmin/issues/1223).

* We change the behaviour of dead variables checker so that it only raised errors when no left values are living after the assignment.
* We do not raise error anymore if found dead variable is a known implicit variable of the architecture.


compiler/linter/Checker/DeadVariables.ml
```diff
@@ -14,27 +14,46 @@ let create_dv_error err_payload loc =
...
+        fun fmt ->
+          Format.fprintf fmt "Instruction contains at least one unused variables. %a " 
+          (Format.pp_print_list ~pp_sep:(fun fmt () -> Format.fprintf fmt ",") (Printer.pp_var ~debug:false))variables
+        );
+  }
 
 let check_func func =
-  let errors = ref [] in
+  let dv_errors = ref [] in
+  let dv_errors_instr = ref [] in
```
vbgl: Why are there two lists? The two kind of errors can be easily distinguished by their code.

compiler/linter/Checker/DeadVariables.ml
```diff
@@ -14,27 +14,46 @@ let create_dv_error err_payload loc =
...
-      let err_payload= create_dv_error v (i_loc.base_loc) in
-      errors := err_payload :: !errors
-    ) dead_variables
+    let assigns = Jasmin.Prog.assigns i_desc in
+    let dead_variables = Sv.diff assigns domain in
+    if (Sv.is_empty (Sv.inter assigns domain)) then (* We check if at least one lvalue is alive. If then, we do not create an error*)
+      Sv.iter (fun v ->
+        let err_payload= create_dv_error v (i_loc.base_loc) in
+        dv_errors := err_payload :: !dv_errors
+      ) dead_variables;
```
vbgl: Why do you produce many errors in this case? The whole instruction is dead.

compiler/tests/fail/linter/bug_1223.jazz
```diff
@@ -0,0 +1,5 @@
+export fn only_one_live_result(reg u64 x) -> reg u64 {
```
vbgl, line 1: Please make sure that the linter is run on this file when running the test suite.

compiler/linter/Checker/DeadVariables.ml
MrDaiki: Because it is not the same error. For the moment, the second kins of error is not raised, I am not sure on how to integrate it

compiler/linter/Checker/DeadVariables.ml
MrDaiki: So you would prefer that the program only output one error for the whole line, can you tell me if this construct is good for you : 
```
[Error] <loc> : instruction <whole_instruction> only affect dead variables (<list of variables>)
```

vbgl: Looks good.

compiler/linter/Checker/DeadVariables.ml
vbgl: Please put them in the same list and print them all.

PR #1225, opened 2025-07-02
adding developper nix file that contains developpers packages

MrDaiki:
Little PR to add an utilitary nix file to create a shell that contains developers tools (ocaml-lsp and ocamlformat for development, llvm packages to run make check)

### DIFF
--- a/compiler/shell-dev.nix
+++ b/compiler/shell-dev.nix
@@ -1,0 +1,9 @@
+with import <nixpkgs> {};
+
+let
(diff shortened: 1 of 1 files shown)


vbgl:
Can you please elaborate on the scenario in which this might be useful?

MrDaiki:
It is useful for environment that need to have ocaml-lsp to provide code insights. The part with llvm is to be able to run the check script from compiler directory.

PR #1228, opened 2025-07-04
opam: fix bounds and install command

eponier:
Publishing the latest version of jasmin to opam (https://github.com/ocaml/opam-repository/pull/28116) and playing with it revealed a few bugs.


eponier:
I should have updated the root `opam` file too.

eponier:
done

vbgl (approved):

PR #1230, opened 2025-07-10
use Rnd events

bgregoir:
use Rnd event in the ITree semantic.


sarranz:
Just so that we don't duplicate work, I have rebased this on top of main as part of the submission artifact, I can update this PR later.

PR #1231, opened 2025-07-11
Improve errors 2025 07 11

eponier:
Hopefully clearer messages when too complex addresses appear


eponier:
I hesitated returning `assert false` in `pp_reg_address_aux` (`src/pp_arm_m4.ml` and `src/pp_riscv.ml`) after this change, because now it should systematically be caught earlier.

eponier:
Also, for the moment, what happens when there is no base in an address is a bit strange. It passes asm_gen, but then crashes at printing in `parse_reg_address`. I don't know what to do. Does it make sense to have addresses without base?

eponier:
> Also, for the moment, what happens when there is no base in an address is a bit strange. It passes asm_gen, but then crashes at printing in `parse_reg_address`. I don't know what to do. Does it make sense to have addresses without base?

After some investigation with @vbgl, it passes on x86 and generates valid code. On arm and risc-v, it crashes for the moment, but doing better than that is left as future work.

PR #1237, opened 2025-07-16
POC: updating the changelog with a script

eponier:
Alternative to https://github.com/jasmin-lang/jasmin/pull/1236

This is a POC of a design to avoid stupid merge conflicts involving the changelog. It relies on a simple script taken from https://github.com/math-comp/math-comp.

Instead of editing CHANGELOG.md, PRs have to create a file `<PR number>-some-text.md` in a sub-directory of `changes`, depending on the kind of changes done by the PR, where <PR number> is the PR number. I create the same categories as in https://github.com/jasmin-lang/jasmin/pull/1236.

Before doing a release, one should type `./changes/generate-release-changelog.sh`. It will generate a file called CHANGELOG_UNREALEASED.md that the release manager should insert inside CHANGELOG.md.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+# Description
+
+Please include a summary of the changes and the related issue (if applicable). Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Fixes # (issue) <!-- if applicable -->
+
+# Checklist
+
+- [ ] Add a changelog entry in `changes` if the PR is a user-visible change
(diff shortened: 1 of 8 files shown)


vbgl:
I suggest to give a try with this solution. Practice will tell if we need something more elaborate.

eponier:
I added a PR template based on https://axolo.co/blog/p/part-3-github-pull-request-template

vbgl:
I’ve just rebased to address the merge conflicts.

PR #1239, opened 2025-07-17
More succinct error messages

vbgl:


eponier:
I improved one error message.

PR #1240, opened 2025-07-17
Checksafety: handle MSB_CLEAR intrinsics

vbgl:


eponier:
For the record, the change is to lose precision instead of crashing.

eponier (approved):

PR #1242, opened 2025-07-17
Compile #set0_64 into a 32-bit xor zero-extended to 64 bits on x86

eponier:
which, according to Vincent, might be useful


eponier:
The proof is nearly identical to the other cases, but sharing it was not obvious, so I gave up.

vbgl (approved):

PR #1246, opened 2025-07-18
Changes after removing most of the wiki

eponier:
- move installation instructions to the doc
- move jasmin.sty to the doc
- extend the README


eponier:
Well spotted!

README.md
```diff
@@ -1,19 +1,40 @@
...
+  ```
+
+- with opam
+  ```
+  opam install jasmin
+  ```
+
+- with APT (debian, ubuntu), a package is available in a dedicated repository, see the [documentation](https://jasmin-lang.readthedocs.io/en/latest/misc/installation_instructions.html#debian-packages)
+
+- from source, see the [documentation](https://jasmin-lang.readthedocs.io/en/latest/misc/installation_instructions.html)
```
vbgl: Why do you repeat what is in the manual? I think that a link to the manual is as useful and less prone to errors.

compiler/src/latex_printer.ml
```diff
@@ -461,7 +461,7 @@ let pp_info fmt =
   F.fprintf fmt "@[<v>@[%% The produced LATEX snippet is meant to be included in a@]@ ";
   F.fprintf fmt "@[%% jasmincode environment provided by the jasmin package@]@ ";
   F.fprintf fmt "@[%% defined in file: @]@ ";
-  F.fprintf fmt "@[%% https://github.com/jasmin-lang/jasmin/wiki/resources/jasmin.sty@]@ ";
+  F.fprintf fmt "@[%% https://jasmin-lang.readthedocs.io/en/latest/tools/jasmin.sty@]@ ";
```
vbgl: In fact, I think that it is too early for this change. The way our deployment of documentation is coupled to our release management means that we must first wait for the `jasmin.sty` file to be deployed at the target address (i.e., wait for next tag to be set in the repo) before we can change the link in the source code.

docs/source/tools/jasmin2tex.md
```diff
@@ -5,4 +5,4 @@ The `jasmin2tex` tool (distributed with the Jasmin compiler; named `jazz2tex` be
 
     jasmin2tex -o output.tex input.jazz
 
-The produced LATEX snippet is meant to be included in a `jasmincode` environment provided by the `jasmin` package defined in the following file: [jasmin.sty](./resources/jasmin.sty).
+The produced LATEX snippet is meant to be included in a `jasmincode` environment provided by the `jasmin` package defined in the following file: [jasmin.sty](jasmin.sty).
```
vbgl: This will not work: you need to tell sphinx to install the `jasmin.sty` file.

README.md
eponier: It's just that I'm used to READMEs that are much more self-contained than ours, and in particular include a section about installation. I don't want to duplicate the content of the doc, just to give some quick pointers. Of course, the complete version is in the doc.

compiler/src/latex_printer.ml
eponier: It refers to the `latest` branch of the doc, so this is not true. But I wonder whether giving pointers to the `latest` instead of `stable` is a good practice.

docs/source/tools/jasmin2tex.md
eponier: Actually this works, but we can do better.

README.md
eponier: I tried to improve a bit.

docs/source/tools/jasmin2tex.md
eponier: I replaced with `[](jasmin.sty)`, this shows a nice download button.

compiler/src/latex_printer.ml
vbgl: I think it is OK to use `stable`. It will be wrong until the next release. Which is not too bad…

eponier: done

compiler/src/latex_printer.ml
```diff
@@ -452,7 +452,7 @@ let pp_info fmt =
   F.fprintf fmt "@[<v>@[%% The produced LATEX snippet is meant to be included in a@]@ ";
   F.fprintf fmt "@[%% jasmincode environment provided by the jasmin package@]@ ";
   F.fprintf fmt "@[%% defined in file: @]@ ";
-  F.fprintf fmt "@[%% https://github.com/jasmin-lang/jasmin/wiki/resources/jasmin.sty@]@ ";
+  F.fprintf fmt "@[%% https://jasmin-lang.readthedocs.io/en/stable/tools/jasmin.sty@]@ ";
```
vbgl, line 455: https://jasmin-lang.readthedocs.io/en/latest/tools/jasmin.sty leads to a 404. So I guess the hardcoded URL will be wrong forever.

Possible fix: revert this change and continue pointing to the wiki (for which we know a valid URL). @eponier Is there an other workaround?

PR #1247, opened 2025-07-18
Linearization generates correct code

eponier:
with a large stack frame and the return address passed on the stack

First reported by @xvzcf

### DIFF
--- a/compiler/tests/success/arm-m4/large_stack/large_stack_return_address.jazz
+++ b/compiler/tests/success/arm-m4/large_stack/large_stack_return_address.jazz
@@ -1,0 +1,27 @@
+// This checks that we properly deal with large stack frames
+// when the return address is passed on the stack
+
+fn g () -> reg u32 {
+  reg u32 res = 0;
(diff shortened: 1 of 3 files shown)


compiler/tests/success/arm-m4/large_stack/large_stack_return_address.jazz
```diff
@@ -0,0 +1,27 @@
...
+fn f () -> reg u32 {
+  // we can add the immediate 7584 in arm-m4, but not 7580
+  stack u8[7580] s;
+  reg u32 res;
+
+  // we make a call to be sure the return address in on the stack
+  res = g();
+
+  s[u32 0] = res;
+  res = s[u32 0];
```
vbgl:
```suggestion
  s[:u32 0] = res;
  res = s[:u32 0];
```

The other syntax has been deprecated.

eponier: indeed.

PR #1248, opened 2025-07-19
Refactor negative tests

vbgl:


compiler/tests/negative.expected
MrDaiki, line 1: Are you sure that we want to have all diagnostics expected results in the same file : 
Wouldn't it be better if each negative test file had a dedicated expected file that we compare against compiler output ?

For example, let say we want to test `fail/annotation/x86-64/calldepth.jazz`. What I suggest is having a file called `fail/annotation/x86-64/calldepth.expected` containing the following text : 

```
"fail/annotation/x86-64/calldepth.jazz", line 3 (0) to line 8 (1):
stack allocation: the maximum call depth is 1 (expected: 2)
```

It might require some reworking but would probably be better to manage in the future. We could automatically split the current file on "fail/*:" to generate the initial expected files.

__EDIT__ : With some reflection, a dedicated PR would probably be needed to integrate this change.

compiler/tests/negative.expected
```diff
@@ -0,0 +1,1520 @@
...
+asmgen: instruction MOVZX_u64u8 is given incompatible args.
+Allowed args are:
+  [[reg]; [reg; mem (glob allowed)]]
+
+Statistics:
+	Annots:      0
+	Pretyping:  44
+	Parsing:     4
+	Typing:      1
+	Compile:   189
```
MrDaiki, line 1520: I don't understand why statistics should be there

compiler/tests/common.mli
```diff
@@ -0,0 +1,10 @@
+open Jasmin.Utils
+
+val check_dir :
```
MrDaiki, line 3: It might be useful to specify in which order files are traversed with this function

compiler/tests/common.ml
```diff
@@ -0,0 +1,29 @@
...
+          [
+            ("x86-64", X86_64 :: archs);
+            ("arm-m4", ARM_M4 :: archs);
+            ("risc-v", RISCV :: archs);
+            ("common", [ X86_64; ARM_M4; RISCV ]);
+          ]
+      with Not_found -> archs
+    in
+    check_dir check_file archs path errors
+  else if String.ends_with filename ".jazz" then check_file archs path errors
```
MrDaiki, line 24: Don't we need to also check jinc files ?

compiler/tests/negative.ml
```diff
@@ -0,0 +1,121 @@
...
+  mutable pretyping : int;
+  mutable parsing : int;
+  mutable typing : int;
+  mutable hi : int;
+}
+
+let counters : counters =
+  { annot = 0; pretyping = 0; parsing = 0; typing = 0; hi = 0 }
+
+let doit path errors arch =
```
MrDaiki: I don't think this function name is really expressive. `run_test_on_file` might bet better (but maybe too long)

compiler/tests/common.mli
vbgl: What do you mean by “specify”?

compiler/tests/common.ml
vbgl: I think the only reason for naming them `*.jinc` is to avoid running tests on them.

compiler/tests/negative.expected
vbgl: Sounds like a very bad idea to me. More code, information scattered all over the file system, no benefit.

compiler/tests/negative.expected
vbgl: Statistics have at least two interest:
1. know what component we test
2. spot if something change unexpectedly (say a parsing error that becomes a compilation error)

compiler/tests/negative.ml
vbgl: Please find a more compelling argument. The suggested name does not appear immediately better.

compiler/tests/common.mli
MrDaiki: write a comment to explain it

compiler/tests/common.ml
MrDaiki: Ok I didn't know

compiler/tests/negative.ml
MrDaiki: My argument is that the `doit` name isn't very clear for a person that doesn't already know the code. Replacing it with an expressive name seems a better idea for me if someone needs to modify this code in the future

compiler/tests/negative.expected
MrDaiki: Anyway it is just a suggestion and should receive a dedicated PR if implemented

compiler/tests/negative.expected
MrDaiki: I understand the interest of having test statistics. My question was more about why putting them in the expected file

vbgl: Where else could we put them? Point 2/ above argues in favor of putting them in the expectation file.

compiler/tests/common.mli
vbgl: Done.

compiler/tests/negative.ml
vbgl: What about `f`? Or better: `aux`. ;-)

vbgl: Done.

compiler/tests/negative.expected
MrDaiki: Ok, I didn't saw it like that

MrDaiki (approved):
LGTM

PR #1249, opened 2025-07-19
Rework Loop.nb

vbgl:
This is an experiment.

- The `Loop` module from `compiler-util` is removed.
- Every compiler pass that needs fuels now takes an explicit `fuel` argument
- Some correctness proofs expect that said fuel is non-zero, so there are a few `.+1` here and there
- Therefore some care in taken in the proofs to prevent unwanted simplification (this is a bit ugly).

The OCaml part calls the compiler with infinite fuel, but this could easily be made configurable.

Comments welcome: is this a step in the right direction?


bgregoir:
I have nothing against this. Did we frequently reach the limit ?

bgregoir:
I missed the last line of the message, I don't know what to think about this Utils.fuel ... I like it but

vbgl:
> Did we frequently reach the limit ?

Never, as far as I know.

PR #1250, opened 2025-07-19
Regalloc: small refactoring

vbgl:
Extracts two subroutines from the register allocation pass:
- a pre-processing step that creates fresh variables to compute and store the return addresses;
- a post-processing step that builds the “oracle” for stack allocation.

This makes it easier to reuse these steps and slightly decreases complexity (smaller functions, less arguments).

### DIFF
--- a/compiler/src/regalloc.mli
+++ b/compiler/src/regalloc.mli
@@ -16,10 +16,13 @@ type reg_oracle_t = {
 module type Regalloc = sig
   type extended_op
 
+  val create_return_addresses :
+    (('info, 'asm) sfundef -> Z.t) -> ('info, 'asm) sfundef list -> retaddr Hf.t
+  (** Compute where the return address will be stored *)
+
   val renaming : (unit, extended_op) func -> (unit, extended_op) func
 
   val subroutine_ra_by_stack : (unit, extended_op) func -> bool
-
 
   (** Returns:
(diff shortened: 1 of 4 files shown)


compiler/src/regalloc.mli
```diff
@@ -16,26 +16,35 @@ type reg_oracle_t = {
...
-  (** Returns:
-    - the input function with variables turned into registers
-    - the set of killed registers (see note below)
-    - a free register in which the stack pointer can be saved; only when asked
-    - a free register for the return address; only for subroutines
-
-  Note: Export functions can freely use caller-saved registers: they are not
-  reported as killed. Subroutines report ALL killed registers.
-
-   *)
```
MrDaiki, line 33: Why changing comment location ?

MrDaiki (approved):
LGTM. just one question about a comment move that I don't understand (probably a standard I don't know about)

compiler/src/regalloc.mli
vbgl: To make it consistent with the other comments in this file that occur below the declaration they refer to.

vbgl: This is also standard practice in OCaml to put the comments below. See for instance: https://github.com/ocaml/ocaml/blob/5.3.0/stdlib/stdlib.mli#L44-L47

PR #1253, opened 2025-08-01
Make pp_arm_m4 check exhaustive

sarranz:
I would like to get a warning when we add a new instruction and we forget to check that its arguments are valid.


vbgl (approved):
Looks like a good practice.

PR #1260, opened 2025-08-19
eclib: lightweight proof for to_uintBb

namasikanam:
Hi, the CI check of the latest EC commit [failed](https://github.com/EasyCrypt/easycrypt/actions/runs/17068200471/job/48403958178) for the new lemma `to_uintBb`. Also, I ran JWord.ec on my laptop, and I failed at `to_uintBb` too. The reason is probably the instability of SMT solving. So, I rewrote the proof to unburden the SMT solver.

I found that the CI check failed after I contributed to EasyCrypt for the first time. I felt a bit embarrassed. I hope this PR can make the CI of EasyCrypt work again. Thanks :)

### DIFF
--- a/eclib/JWord.ec
+++ b/eclib/JWord.ec
@@ -811,7 +811,11 @@ have /= x_range := to_uint_cmp x.
 have /= y_range := to_uint_cmp y.
 rewrite /b2i to_uintD.
 case: c => _; last first.
-+ case: (to_uint y = 0) => /= hy; rewrite to_uintN ?hy /#.
++ rewrite to_uintN !addr0 modzDmr.
+  case: (to_uint x < to_uint y) => Hxy /=.
+  + rewrite (addzC (to_uint x - to_uint y) modulus).
+    by rewrite &(modz_sub_carry) // subz_lt0 //.
+  + by rewrite pmod_small /#.
 rewrite to_uintN.
 case: (to_uint y = max_uint) => /= hy.
 + rewrite to_uintD hy /#.


vbgl:
What version of Z3 are you using (on your laptop and in EC CI)? If we aligned it with Jasmin CI, that could prevent similar cases to occur again. Jasmin CI currently uses 4.15.1.

vbgl:
> let EC CI call the eclib job from Jasmin CI

Sounds like a maintenance nightmare to me…

> Z3 4.8.17, 4.12.6, 4.13.4

Does really easycrypt spawns three instances of the Z3 prover?

fdupress:
Yes to trying to synchronise SMT prover versions, but not to pulling a new image into the EasyCrypt CI loop a priori. (Also, the Jasmin image and job pull EC from main; the point of the external CI is that it uses the current branch.)

The job passed CI in the PR, but failed the next **identical** run post-merge. We want to catch things like this because they are an issue with our proofs, not with the SMT solvers or their configuration. So the failure in itself is a symptom, and putting effort into hiding it is counter-productive.

But, keeping the SMT versions out of sync is silly. So how do we solve this?

EasyCrypt's adoption of new solvers is slow by design: our libraries and external jobs are not "just program verification things", and tend to torture SMT solvers in ways that most of their regression tests and benchmarks don't measure. We catch regressions in SMT behaviours all the time, _and_ we have a lot of proofs to maintain with very few people willing and able. Those things are not things we need to stay on top of, and we choose to be deliberate about upgrading SMT solvers, so we know we have the time to deal with the regressions.

On the other hand, Jasmin is more or less tied to nixpkgs and the versions of provers it contains. I don't want our deliberate approach to stop you from keeping your things simple.

One way to force us to consider an upgrade is to keep your easycrypt.project file tight to the prover versions you carry. (So if nixpkgs gives you z3 4.15, specify that you want z3 4.15.) This will cause more external CI failures on our side, and might create a burden, but we don't know what that burden is until we try it.

Talk through this on Friday?

fdupress:
> Does really easycrypt spawns three instances of the Z3 prover?

No, we include all 3 in our image, to give flexibility to the external CI job. (We pick the lowest version that is above that specified in the easycrypt.project.)

We could do the same for providing prover versions that are *ahead* of the one we use in our main CI job.

strub:
I won’t be available on Friday. Regarding the CI image: I think it’s perfectly reasonable for the EasyCrypt CI to include more versions of the provers. For instance, we could commit to maintaining the last 3–4 minor versions of the most commonly used provers, even if some go beyond the versions currently used in the EasyCrypt standard library. This way, each project can choose which provers to use without forcing EasyCrypt itself to stay in lockstep with them.

As a side note, since EasyCrypt still requires Z3 4.12 while the current release is already at 4.15, we’ll need to move forward anyway. (Though that’s separate from my main point.)

PR #1263, opened 2025-08-20
Do not introduce missing L-values

vbgl:


sarranz (approved):
Thanks a lot! I'll recall this in the next Formosa meeting, since some people will need to patch their code.

vbgl:
The compiler have been issuing a warning about that for ages. I think people had plenty of time to patch their code. But indeed, that may deserve an announcement.

PR #1269, opened 2025-08-28
eclib: install a configuration file

vbgl:
This make it possible for EasyCrypt (next version) to automatically detect the installed Jasmin library (without extra configuration step by the user).

### DIFF
--- a/eclib/Makefile
+++ b/eclib/Makefile
@@ -8,25 +8,35 @@ CHECKS   ?= jasmin
 # --------------------------------------------------------------------
 DESTDIR  ?=
 PREFIX   ?= /usr/local
-BINDIR   := $(PREFIX)/bin
 LIBDIR   := $(PREFIX)/lib
-SHRDIR   := $(PREFIX)/share
 INSTALL  ?= install
 
 # --------------------------------------------------------------------
-.PHONY: default usage check install uninstall
+.PHONY: default usage check install uninstall clean
 
 default: check
 
 usage:
-	@echo "Usage: make <target> where <target> in [check|install|uninstall]" >&2
+	@echo "Usage: make <target> where <target> in [check|clean|install|uninstall]" >&2
 
 check:
 	easycrypt runtest $(ECARGS) $(ECCONF) $(CHECKS)
 
-install:
-	$(INSTALL) -m 0755 -d $(DESTDIR)$(LIBDIR)/jasmin/easycrypt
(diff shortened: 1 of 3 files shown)


eclib/Makefile
```diff
@@ -8,25 +8,32 @@ CHECKS   ?= jasmin
...
 
 check:
 	easycrypt runtest $(ECARGS) $(ECCONF) $(CHECKS)
 
-install:
-	$(INSTALL) -m 0755 -d $(DESTDIR)$(LIBDIR)/jasmin/easycrypt
-	$(INSTALL) -m 0644 -t $(DESTDIR)$(LIBDIR)/jasmin/easycrypt *.ec
+jasmin.conf:
+	echo "[general]" | tee $@
+	echo "idirs=Jasmin:$(LIBDIR)/jasmin/easycrypt" | tee -a $@
```
eponier: You swapped `jasmin` and `easycrypt` in the paths of `install:`, but not here. And there is `DESTDIR` in `install:` and not here. Are these differences intended?

eponier:
And is there a risk of overwriting the config file of the user if they already have one?

eclib/Makefile
vbgl: well spotted. The omission of DESTDIR is intended.

vbgl: Fixed.

vbgl:
Done.

eponier:
Adding this comment was a good thing. But my question was more: does a normal EC user already have a conf file at that path?

eclib/Makefile
```diff
@@ -8,25 +8,34 @@ CHECKS   ?= jasmin
...
+	@echo "idirs=Jasmin:$(LIBDIR)/easycrypt/jasmin" | tee -a $@
+
+install: jasmin.conf
+	$(INSTALL) -m 0755 -d $(DESTDIR)$(LIBDIR)/easycrypt/jasmin
+	$(INSTALL) -m 0644 -t $(DESTDIR)$(LIBDIR)/easycrypt/jasmin *.ec
+	$(INSTALL) -m 0755 -d $(DESTDIR)$(LIBDIR)/easycrypt/config
+	$(INSTALL) -m 0644 -t $(DESTDIR)$(LIBDIR)/easycrypt/config *.conf
 
 uninstall:
 	$(RM) -r $(DESTDIR)$(LIBDIR)/jasmin
```
eponier: This should be updated also I guess. Apart from that, I can merge.

vbgl: Done. Thanks for the careful review.

PR #1275, opened 2025-09-08
Linter: refactor & add “levels”

### LINKED ISSUE #1223: Too many dead assignment warnings

vbgl:
# Description

See the commit messages.

Fixes #1223.

The design of the “linting levels” is experimental and _basic_: we could easily contemplate something more fine-grained such as a lattice of sets of sets of warnings, with names, etc. However the purpose of this PR is to start, and to provide to users something usable.

# Checklist

### DIFF
--- a/compiler/linter/Checker/DeadVariables.ml
+++ b/compiler/linter/Checker/DeadVariables.ml
@@ -27,13 +27,19 @@ let create_dv_error_instr loc =
       );
   }
 
+(* Instructions with an AT_keep tag are not treated as dead *)
+let has_keep_tag =
+  function
+  | Cassgn (_, tag, _, _) | Copn (_, tag, _, _) -> tag = AT_keep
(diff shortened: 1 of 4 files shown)


vbgl:
I’ve just added the documentation.

vbgl:
There still are known bugs…

PR #1277, opened 2025-09-09
docs: fix some internal links

vbgl:
~~Add jasmin.sty to the docs~~


eponier:
If we put the file next to `jasmin2tex.md` and use `[](jasmin.sty)`, we get more or less the same result, plus an additional sign showing that it is a file to download. I don't understand why it does not appear in variants of this syntax (including the one you use in the PR) though, it seems a bug to me.

vbgl:
Alright. I’ve simplified this PR to just fix two dead links. I’ll leave the fixing of jasmin.sty to you as you seem to better understand what is going on.

eponier (approved):

eponier:
I can't say I understand better, it's more that I lose more time on experimenting...

PR #1283, opened 2025-09-18
Ignores useless zero-extensions of primitives

eponier:
# Description

We detect trivially useless zero-extensions of primitives (the size of the extension is the same as the size of the output of the primitive), drop them and produce a warning.

# Checklist


vbgl:
What about a changelog entry, as more programs are accepted?

PR #1284, opened 2025-09-18
Fixes the typechecking of #swap

eponier:
# Checklist

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -1998,13 +1998,13 @@ let rec tt_instr arch_info (env : 'asm Env.env) ((annot,pi) : S.pinstr) : 'asm E
         match xs with
         | [x; y] ->
           let loc, x, oxty = tt_lvalue arch_info.pd env_lhs x in
-          let yloc, y, _oytu = tt_lvalue arch_info.pd env_lhs y in
+          let yloc, y, oyty = tt_lvalue arch_info.pd env_lhs y in
           let ty =
             match oxty with
             | None -> rs_tyerror ~loc (string_error "_ lvalue not accepted here")
             | Some ty -> ty in
           let _ =
-             match oxty with
+             match oyty with
             | None -> rs_tyerror ~loc (string_error "_ lvalue not accepted here")
             | Some yty -> check_ty_eq ~loc:yloc ~from:yty ~to_:ty in
           [x ty; y ty], ty
--- a/compiler/tests/fail/typing/x86-64/swap.jazz
+++ b/compiler/tests/fail/typing/x86-64/swap.jazz
@@ -1,0 +1,4 @@
+export fn main (reg u64 x, reg u32 y) -> reg u64 {
+  x, y = #swap(x, x);
+  return x;
+}
--- a/changes/02-bugfix/1284-swap-typechecking.md
+++ b/changes/02-bugfix/1284-swap-typechecking.md
@@ -1,0 +1,2 @@
+- Fixes the typechecking of the `#swap` primitive
+  ([PR #1284](https://github.com/jasmin-lang/jasmin/pull/1284)).


compiler/src/pretyping.ml
```diff
@@ -1998,13 +1998,13 @@ let rec tt_instr arch_info (env : 'asm Env.env) ((annot,pi) : S.pinstr) : 'asm E
...
-          let yloc, y, _oytu = tt_lvalue arch_info.pd env_lhs y in
+          let yloc, y, oyty = tt_lvalue arch_info.pd env_lhs y in
           let ty =
             match oxty with
             | None -> rs_tyerror ~loc (string_error "_ lvalue not accepted here")
             | Some ty -> ty in
           let _ =
-             match oxty with
+             match oyty with
             | None -> rs_tyerror ~loc (string_error "_ lvalue not accepted here")
```
vbgl, line 2008: I don’t get the rationale for failing here.

I agree that it is a pain to deal with the case when both L-values are `_` (although the type could be inferred from the RHS) so it is fair to reject it. But accepting one `_` seems doable.

compiler/tests/fail/typing/x86-64/swap.jazz
```diff
@@ -0,0 +1,4 @@
+export fn main (reg u64 x, reg u32 y) -> reg u64 {
+  x, y = #swap(x, x);
+  return x;
+}
```
vbgl, line 4: Why do you consider that this program is ill-typed? This is very similar to:
~~~
export fn main (reg u64 x, reg u32 y) -> reg u64 {
  y = x;
  return x;
}
~~~

compiler/src/pretyping.ml
eponier: I just wanted to make the pretyping code do what it was supposed to do initially, because there was clearly a typo in the code.

I thought about being less strict, and allowing underscores, but what is the motivation? `x, _ = #swap (x, y);` is just `x = y`. Maybe it makes sense for debugging?

compiler/tests/fail/typing/x86-64/swap.jazz
eponier: I can write
```
export fn main (reg u32 x, reg u64 y) -> reg u32 {
  x, y = #swap(x, x);
  return x;
}
```
if you prefer

eponier: (it passes pretyping but fails at typing; with the PR, it is caught in pretyping)

vbgl: So your PR is wrong: type errors should be caught by type-checking…

Really, there is no reason to treat `swap` differently than any (binary) operator: check the arguments, compute the greatest common type, add casts if needed, and use that type to fill any hole in the LHS.

compiler/src/pretyping.ml
vbgl: > but what is the motivation?

The purpose of pre-typing is to produce abstract-syntax, filling holes (e.g., missing type annotations). The fact that it makes explicit a few implicit conversions is already a bit borderline. Crashing to enforce silly style rules is not a correct behavior of pre-typing: the motivation is to fix that bug.

compiler/tests/fail/typing/x86-64/swap.jazz
eponier: Hum, I see. But, once again, my goal was a quick fix to the current code that treats `#swap` specially (and that is clearly doing something weird, since it compares a type with itself on line 2009).

To me, the mere fact that we have a special case for `#swap` feels wrong, but I guess we cannot do better for now. Once we have arrays of unknown sizes, we may hopefully declare `#swap` with some kind of polymorphism and remove the specific case for it.

compiler/src/pretyping.ml
eponier: I completely agree that the current situation is far from perfect, but fixing it was not the goal of this PR. I'm not planning to spend more time on this, if you think this goes in the wrong direction, then we can just close the PR.

compiler/tests/fail/typing/x86-64/swap.jazz
vbgl: see: af833a54221902dbc644dddc0b72377018f1eb56

compiler/src/pretyping.ml
vbgl: > fixing it was not the goal of this PR

I may have over-interpreted the title of the PR.

PR #1286, opened 2025-09-22
Regalloc: simplify interaction with stack-alloc

vbgl:
Current behavior is wrong: the register holding the return address is computed during the first round of register-allocation instead of the second one.

### DIFF
--- a/compiler/src/regalloc.ml
+++ b/compiler/src/regalloc.ml
@@ -1031,13 +1031,7 @@ let renaming (f: ('info, 'asm) func) : (unit, 'asm) func =
   let subst = subst_of_allocation vars a in
   Subst.subst_func subst f
 
-(** Returns extra information (k, rsp) depending on the calling convention.
-
- - Subroutines:
-   - k: all registers overwritten by a call to f (including ra)
-   - rsp: None
-
- - Export:
+(** Returns extra information (k, rsp) for export functions
     - k: all callee-saved registers overwritten by this function (including rsp)
     - rsp: if ~stack_needed and if there is a free register, a free register to hold the stack pointer of the caller (aka environment)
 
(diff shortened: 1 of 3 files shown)


compiler/src/regalloc.ml
```diff
@@ -612,13 +612,8 @@ let allocate_one nv vars loc (cnf: conflicts) (x_:var) (x: int) (r: var) (a: A.a
          other
 
 type reg_oracle_t = {
-    (* The list of callee save registers that are modified by
-      a call to the export function *)
     ro_to_save: var list;
```
bgregoir: You don't want to have the same comment than in the mli ?

proofs/compiler/allocation.v
```diff
@@ -685,15 +685,10 @@ Definition check_f_extra_s (r: M.t) (e1 e2: extra_fun_t) p1 p2 : cexec M.t :=
...
-  (e1.(sf_save_stack) == e2.(sf_save_stack)),
-  (e1.(sf_return_address) == e2.(sf_return_address)) &
+  (is_RAnone e1.(sf_return_address) == is_RAnone e2.(sf_return_address)) &
   (e1.(sf_align_args) == e2.(sf_align_args))]
       (E.error "extra not equal") in
-  if e1.(sf_return_address) == RAnone then
-    check_vars p1 p2 r
-  else
-    check_vars p1 p2 r.
+  check_vars p1 p2 r.
```
bgregoir, line 691: this was interesting 👍

bgregoir (dismissed):

compiler/src/regalloc.ml
vbgl: I’ve just added the comment.

proofs/compiler/allocation.v
vbgl: Indeed ;-)

bgregoir (approved):
Sorry for the delay.

PR #1287, opened 2025-09-22
doc: tweak the version switcher

eponier:
# Description

This is a hack to make the version switcher of the doc a bit nicer (to me, at least). In particular:
- no duplication between stable and the current stable tag
- changing version preserves the current path in the doc

## Context

By default, readthedocs create two additional tags, `latest` and `stable`. `latest` points to the latest main, `stable` is just a copy of the latest release. The pain point is that when the stable page is accessed, it is not shown what version it corresponds to. And there is some kind of duplication between `stable` and the latest release. The version switcher show both, while they contain the same content.

Another problem of the version switcher is that changing version brings you back to the homepage of the documentation. If you want to check how a page evolved in time, this is painful.

The first fix I tried was to redirected `stable` to `/`, which is then expanded to the default version. But the inconvenience of this approach is that it systematically redirects to the homepage. (Another solution would to redirect to the current latest release, but this means updating the redirect rule after each release.) Also, I once observed some strange interaction with my browser: since the doc website is static, it looks like browsers cache it aggressively, meaning after a change I have to ask to drop the cache before reloading the page, but it did not work for some reason due to the redirection.

## Solution

Instead, we implement the following hack. In the version switcher, `stable` is renamed into `vXXX (stable)` and the latest release `vXXX` is removed. Note that, in this solution, both pages `stable` and `vXXX` exist, they are just not shown in the version switcher. We also tweak the implementation, so that the relative path in the doc is preserved when changing version.

The main drawback of this approach is that it duplicates a file from sphinx_rtd_theme, because I did not find another way to change it. And I removed half of it, because this half contains templates and I could not figure out how to process the template file (the code that I removed is not used for the moment, and I don't expect it to be used in the future, so it's not too bad).

Here are related issues:
- https://github.com/readthedocs/readthedocs.org/issues/4000
- https://github.com/readthedocs/readthedocs.org/issues/5318
- https://github.com/readthedocs/readthedocs.org/issues/4529
- https://github.com/readthedocs/sphinx_rtd_theme/issues/1638

Note that the theme will be updated in the future (https://github.com/readthedocs/addons/issues/574), hopefully this hack may be removed then.

### DIFF
--- a/docs/source/_static/js/versions.js
+++ b/docs/source/_static/js/versions.js
@@ -26,35 +26,31 @@ IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
-const themeVersionSelector = true;
-const themeLanguageSelector = false;
+function onSelectorSwitch(event) {
+  const option = event.target.selectedIndex;
+  const item = event.target.options[option];
+  window.location.href = item.dataset.url;
+}
 
-if (themeLanguageSelector || themeVersionSelector) {
-  function onSelectorSwitch(event) {
-    const option = event.target.selectedIndex;
-    const item = event.target.options[option];
-    window.location.href = item.dataset.url;
+document.addEventListener("readthedocs-addons-data-ready", function (event) {
+  const config = event.detail.data();
+
+  const versionSwitch = document.querySelector(
+    "div.switch-menus > div.version-switch",
+  );
+
+  let versions = config.versions.active;
+
+  // HACK: It is assumed that index 0 is latest, 1 is stable and 2 is the current stable tag.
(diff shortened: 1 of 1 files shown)


eponier:
This can be tested here: https://jasmin-lang.readthedocs.io/en/docs-experiment/

docs/source/_static/js/versions.js
```diff
@@ -0,0 +1,127 @@
...
+                  ${language.language.name}
+              </option>`,
+          )
+          .join("\n")}
+       </select>
+    `;
+
+        languageSwitch.innerHTML = languageSelect;
+        languageSwitch.firstElementChild.addEventListener("change", onSelectorSwitch);
+      }
```
vbgl: This is dead code (`themeLanguageSelector` is defined as `false` above). Are we sure we want to keep it?

eponier: My initial idea was to take the file as a template and just customize it. But since I already removed half of it, I can remove more. I don't think we'll have documentation in other languages than English soon, if ever.

PR #1288, opened 2025-09-23
Minor cleaning

vbgl:


compiler/tests/negative.ml
```diff
@@ -47,7 +47,7 @@ let config path =
         ("fail/typing/x86-64/non_inline_stack_arg.jazz", reset_warn_recoverable);
         ("fail/typing/x86-64/non_inline_stack_res.jazz", reset_warn_recoverable);
         ( "fail/typing/x86-64/write_constant_pointer_direct_array.jazz",
-          disable_warnings [ PedanticPretyping ] );
+          reset_warn_recoverable );
```
eponier, line 50: Shouldn't we do the same for the next test?

eponier (approved):
Except the comment above, this is good!

compiler/tests/negative.ml
vbgl: Of course no. These « pedandic » warnings should go away (from pretyping).

eponier: I answer late, but anyway I didn't want to block the PR. For me, it is not clear that checking `const` and `mut` is not part of typing (but I acknowledge that `write_constant_pointer_subproc_array.jazz` does not use `const`, it is inferred from the array not being returned from function `foo`, maybe this is the distinction you see with `write_constant_pointer_direct_array.jazz`). Additionally,  the file is in the `typing` subdirectory, meaning it was written to exercise typing in the broad sense (pretyping + typing), so letting it fail at this step of compilation is a sensible thing to me.

Anyway, I feel this is another occurrence of the debate about whether tests should exercise all the current behaviours of the compiler, including the parts that we are not satisfied with. For me, having a test does not necessarily mean that this is the intended behaviour, just that this is the current behaviour of the compiler. The day we find time to finally fix the compiler, we have a natural test case to check that the unsatisfying behaviour disappeared.

PR #1289, opened 2025-09-29
Infer returned-params during stack-allocation

vbgl:
# Description

The first two commits and the last one remove dead code.

The third commit allows to selectively turn a pre-typing error into a simple warning. This pre-typing check shall be turned into a proper linting analysis.

The fourth commit is the main change relevant to this PR: stack-allocation no longer uses the returned-params information that is present in the calling-convention (and inferred during parsing) but instead uses the result of the alias analysis to determine which regions amongst the function arguments are returned.

This paves the way to more flexibility in the programs that can be written, but more importantly, it simplifies the working of the compiler:

  - pre-typing no longer needs to perform the inference (it still does for backward compatibility reasons)
  - calling-convention no longer need to carry from source to assembly the returned-param information

Fixes # (issue) 

# Checklist

### DIFF
--- a/compiler/src/compile.ml
+++ b/compiler/src/compile.ml
@@ -244,40 +244,13 @@ let compile (type reg regx xreg rflag cond asm_op extra_op)
     in
     Hv.iter doarr harrs;
 
-    let f_cc =
-      match fd.f_cc with
-      | Subroutine si ->
-          (* Since some arguments/returns are expended we need to fix the info *)
-          let tbl = Hashtbl.create 17 in
-          let newpos = ref 0 in
-          let mk_n x =
-            match x.v_kind, x.v_ty with
-            | Reg (_, Direct), Arr (_, n) -> n
-            | _, _ -> 1
-          in
-          let doarg i x =
-            Hashtbl.add tbl i !newpos;
-            newpos := !newpos + mk_n x
-          in
-          List.iteri doarg fd.f_args;
-          let doret o x =
-            match o with
-            | Some i -> [Some (Hashtbl.find tbl i)]
-            | None -> List.init (mk_n (L.unloc x)) (fun _ -> None)
-          in
-          let returned_params =
-            List.flatten (List.map2 doret si.returned_params fd.f_ret) in
-          FInfo.Subroutine { returned_params }
-
-      | _ -> fd.f_cc
-    in
     let do_outannot x a =
       try
         let (_, va) = Hv.find harrs (L.unloc x) in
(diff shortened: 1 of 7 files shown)


compiler/tests/fail/stack_allocation/x86-64/return_ptr_global.jazz
```diff
@@ -0,0 +1,6 @@
+u32[1] g = { 0xabcd };
+
+export fn main() -> reg ptr u32[1] {
```
bgregoir, line 3: This is an interesting example, do we really want to reject it ?

bgregoir: This is certainly for futur work but I think that it is ok if stack alloc knows that the returned array is not writable.

compiler/tests/negative.expected
```diff
@@ -1239,11 +1239,23 @@ and destination
...
+"fail/stack_allocation/x86-64/return_ptr_global.jazz", line 5 (9-10):
 compilation error in function main:
-stack allocation: the returned pointer (r) is one of the parameters, but it does not point to the region it pointed to initially
+variable allocation: returned variable p points to a region that should not escape the local scope: #g:g[0; 4[
+
+fail/stack_allocation/x86-64/return_ptr_local.jazz:
+
+"fail/stack_allocation/x86-64/return_ptr_local.jazz", line 5 (9-10):
+compilation error in function main:
+variable allocation: returned variable r points to a region that should not escape the local scope: r[0; 4[
```
eponier, line 1252: For future work, it would be better if the local scope mentioned `s` instead.

vbgl: The variable that is picked as a representative of its class is determined by a comparison function. Now the choice between `r` and `s` is completely arbitrary:
https://github.com/jasmin-lang/jasmin/blob/51a918c012f1dde92e209b3283e57fb25e2db571/compiler/src/alias.ml#L153

I don’t know if there is a better heuristic.

eponier: I have a patch.

eponier: This is a straightforward patch: when the variables are local and have the same length, avoid picking pointers. Cf. https://github.com/jasmin-lang/jasmin/pull/1294

eponier:
> I think that there is no user-facing change, so there is no need neither for updating the documentation nor for a changelog entry. Am I right?

I considered that you were right :smile:

PR #1293, opened 2025-10-08
Split stype into 3 types

eponier:
# Description

- `atype` (for abstract type): syntactic types (e.g. the types of
  variables)
- `ctype` (for concrete type): semantical types (e.g. the types of values)
- `ltype` (for low type): type of assembly instructions

This allows to make `atype` evolve without affecting the two other types.
This prepares in particular the apparition of length variables in the
types of arrays.

`atype` and `ctype` have already diverged. In `atype`, the case of arrays now
stores both the cell size and the length, instead of just the number of
bytes. A predicate "convertible" is introduced to compare `atype`s, since
`==` might be too strict.

Supersedes https://github.com/jasmin-lang/jasmin/pull/1177

The new names are not necessarily definitive. See the discussion in https://github.com/jasmin-lang/jasmin/pull/1177 for more context.


compiler/src/conv.ml
```diff
@@ -34,16 +34,16 @@ let z_unsigned_of_word sz z = z_of_cz (Word0.wunsigned sz z)
...
 
 let cty_of_ty = function
-  | Bty Bool      -> T.Coq_sbool
-  | Bty Int       -> T.Coq_sint
-  | Bty (U sz)   -> T.Coq_sword(sz)
-  | Arr (sz, len) -> T.Coq_sarr (pos_of_int (size_of_ws sz * len))
+  | Bty Bool      -> T.Coq_abool
+  | Bty Int       -> T.Coq_aint
+  | Bty (U sz)   -> T.Coq_aword(sz)
+  | Arr (sz, len) -> T.Coq_aarr (sz, pos_of_int len)
```
vbgl, line 40: So `cty` stands for `atype`…

compiler/src/evaluator.ml
```diff
@@ -186,8 +186,18 @@ let run (type reg regx xreg rflag cond asm_op extra_op)
   exec ep spp sip scs0 p ii fn args m
 
 (* ----------------------------------------------------------- *)
-let pp_undef fmt ty = 
-  Format.fprintf fmt "undef<%a>" PrintCommon.pp_ty (Conv.ty_of_cty ty)
+(* FIXME: do we really want this conv function,
+   or directly a function printing ctype? *)
```
vbgl: I’d go for a function printing `ctype` specialized to the use case: use `pp_btype` for base types and assert false for arrays.

proofs/arch/arch_decl.v
```diff
@@ -559,8 +555,6 @@ Definition instr_desc (o:asm_op_msb_t) : instr_desc_t :=
        id_args_kinds := exclude_mem d.(id_args_kinds) d.(id_out) ;
        id_nargs      := d.(id_nargs);
        id_eq_size    := instr_desc_aux1 ws d.(id_eq_size);
-       id_tin_narr   := d.(id_tin_narr);
-       id_tout_narr  := instr_desc_tout_narr _ d.(id_tout_narr);
```
vbgl, line 563: Is this change related to this PR? This is not an objection. I’m just curious.

proofs/arch/arch_extra.v
```diff
@@ -296,23 +296,118 @@ Qed.
...
+  refine
+  (match ctype_eqb ty1 ty2 as b return _ = b -> _ with
+  | true => _
+  | false => _
+  end erefl).
+  move=> b; left. apply /ctype_eqb_OK. done.
+  move=> b; right. apply /ctype_eqb_OK/negPf. done.
+Defined.
+*)
+
```
vbgl: This might be cleaned up a bit.

Side note: could it be considered a bug in elpi/HB that the `ctype_eqb_OK` reflection lemma is opaque?

compiler/src/conv.ml
eponier: Here `c` is for `Coq`, but I agree that this is confusing.

compiler/src/evaluator.ml
eponier: ok

proofs/arch/arch_decl.v
eponier: Yes, now that `ltype` does not contain a case for arrays, there is no need to have a proof that there are no arrays. This is now enforced statically in the type.

proofs/arch/arch_extra.v
eponier: Sorry, I admit I pushed without really proofreading.

I discussed with Enrico about this a while ago. He implemented [this change](https://github.com/LPCIC/coq-elpi/commit/30fc8742e3181a265992b53407a8e99fa783142c) to make the sumbool version compute a little. Namely, it can reduce to `left _` or `right _`, covering most of the use cases where we only want to check for equality. In our case, though, I want it to reduce to `left erefl` on concrete cases, so that everything simplifies just by computation. To do this, I think it would need to change completely the way the sumbool version is created. Currently, it is just a wrapping of the reflect version. I think a version that truly computes would need to be written in parallel to the reflect version, so this is not a small bugfix. But it could be considered a bug (not that `ctype_eqb_OK` is opaque, but that the sumbool version is based on it).

eponier: Also, we could go with the default sumbool version, this would just mean that we would need to use `computational_eq_refl` in a systematic way. For now, we have it for free when comparing concrete types.

eponier: Cleaned up

compiler/src/evaluator.ml
eponier: done

proofs/arch/arch_decl.v
vbgl: Thanks for the clarification.

proofs/arch/arch_extra.v
vbgl: Thanks. Not completely satisfying, but much better (and good enough)!

proofs/compiler/allocation_proof.v
```diff
@@ -163,12 +166,13 @@ Section CHECK_EP.
...
       by eexists; split; first by reflexivity.
     move => t e He e11 He11 e12 He12 [] // t' e2 e21 e22 r re vm1.
-    t_xrbindP => r1 r' /eqP <- /He Hr' /He11 Hr1 /He12 Hr2 {He He11 He12}.
+    t_xrbindP => r1 r' /convertible_eval_atype -> /He Hr' /He11 Hr1 /He12 Hr2 {He He11 He12}.
     move=> /Hr'{Hr'}[] /Hr1{Hr1}[] /Hr2{Hr2}[] Hre Hs2 Hs1 Hs;split=>// scs m v1.
     t_xrbindP=> b > /Hs [_] /= [->] /= /[swap] /to_boolI -> /value_uinclE ->.
     move=> ?? /Hs1 [?[-> /=]] /value_uincl_truncate H/H{H} [? -> ?].
     move=> ?? /Hs2 [?[-> /=]] /value_uincl_truncate H/H{H} [? -> ?] <- /=.
     by eexists;split;eauto;case: (b).
+  Local Opaque arr_size.
```
vbgl: This line is weird. Typo?

proofs/compiler/constant_prop.v
```diff
@@ -422,8 +422,9 @@ Fixpoint const_prop_rvs globs (m:cpm) (rvs:lvals) : cpm * lvals :=
     (m, rv::rvs)
   end.
 
-Definition wsize_of_stype (ty: stype) : wsize :=
-  if ty is sword sz then sz else U64.
+(* is also defined in x86_lowering, should it be Uptr instead of U64? *)
+Definition wsize_of_atype (ty: atype) : wsize :=
+  if ty is aword sz then sz else U64.
```
vbgl: In this file, it seems that the `else` branch is dead code. So I’d rather change the comment to say that instead.

proofs/compiler/allocation_proof.v
eponier: Indeed, it should be `Transparent` instead.

proofs/lang/psem_defs.v
```diff
@@ -48,7 +46,7 @@ Definition gv2val (gd:glob_value) :=
 Definition get_global gd g : exec value :=
   if get_global_value gd g is Some ga then
     let v := gv2val ga in
-    if type_of_val v == vtype g then ok v
+    if type_of_val v == eval_atype (vtype g) then ok v
```
vbgl, line 49: Should this use `convertible`?

proofs/lang/sem_type.v
```diff
@@ -14,79 +14,137 @@ Class WithSubWord := { sw_allowed : bool }.
 Definition nosubword   := {| sw_allowed := false |}.
 Definition withsubword := {| sw_allowed := true |}.
 
-Definition compat_type (sw:bool) :=
-  if sw then subtype else eq_op.
+Definition compat_atype (sw:bool) :=
```
vbgl, line 17: Why use `bool` here rather than `WithSubWord`?

proofs/lang/sopn.v
```diff
@@ -376,93 +377,98 @@ Definition SLHmove_instr :=
...
+      [:: aword ws; ty_msf ]
       [:: E 0; E 1 ] (* this info is irrelevant *)
-      [:: sword ws ]
+      [:: aword ws ]
       [:: E 2 ]      (* this info is irrelevant *)
       (@se_protect_sem ws)
       true.
 
+(* TODO: move *)
+Notation cty_msf := (cword msf_size).
```
vbgl: Move where, and when? I guess this should be done in this PR.

proofs/lang/psem_defs.v
eponier: No. When we compare `atype`, we use `convertible`, but when we compare `ctype` we use `==`.

proofs/lang/sem_type.v
eponier: Actually, I don't know. I just adapted the previous code.

proofs/lang/sopn.v
eponier: done

proofs/compiler/allocation_proof.v
eponier: done

proofs/lang/psem_defs.v
vbgl: So it might be named `aonvertible` instead. (just kidding).

proofs/compiler/constant_prop.v
eponier: I changed the comment (but kept the part about code duplication).

PR #1294, opened 2025-10-08
alias.ml: better representatives

eponier:
We avoid picking pointers as representatives of the alias classes. In particular, we prefer local stack over pointers. This gives clearer error messages.


vbgl (approved):
Well done.

PR #1295, opened 2025-10-09
Register allocation: changes in heuristics

vbgl:
# Description

This change makes it possible to compile more programs, e.g., all of libjade legacy (that currently relies on `-lazy-regalloc` in various implementations) and makes register allocation more robust: more registers are chosen as a result of a computation (that depends on the program) rather than based on internal implementation details of the data-structures.

Hoping there are not too many regressions.

# Checklist


vbgl:
Both.

eponier (approved):

PR #1297, opened 2025-10-14
ASM printing: change formatting of raw bytes

vbgl:
# Description

Printing of global data in assembly listings puts up to sixteen bytes per line. Each byte is expressed as a non-negative decimal integer.

# Checklist

### DIFF
--- a/compiler/src/asm_utils.ml
+++ b/compiler/src/asm_utils.ml
@@ -30,14 +30,18 @@ let format_glob_data globs names =
   let names =
     List.map (fun ((x, _), p) -> (Conv.var_of_cvar x, Conv.z_of_cz p)) names
   in
-  List.flatten
-    (List.mapi
-       (fun i b ->
-         let b = Byte (Z.to_string (Conv.z_of_int8 b)) in
-         match List.find (fun (_, p) -> Z.equal (Z.of_int i) p) names with
-         | exception Not_found -> [ b ]
-         | x, _ -> [ Label (string_of_glob occurrences x); b ])
-       globs)
+  let init = [], [] in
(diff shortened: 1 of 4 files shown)


eponier (approved):
I  trust you on the fact that it means the same thing in assembly language.

compiler/src/asm_utils.ml
```diff
@@ -30,14 +30,18 @@ let format_glob_data globs names =
...
-         | exception Not_found -> [ b ]
-         | x, _ -> [ Label (string_of_glob occurrences x); b ])
-       globs)
+  let init = [], [] in
+  let close (bytes, acc) = if bytes = [] then acc else Bytes (List.rev bytes) :: acc in
+  let push b (bytes, acc) = (b :: bytes, acc) in
+  let add_label x s = ([], Label x :: close s) in
+  let finish s = s |> close |> List.rev in
+  List.fold_lefti (fun s i b ->
+      let b = Z.format "%3i" (Conv.z_unsigned_of_word U8 b) in
```
eponier, line 39: Does `Conv.z_unsigned_of_word` make a real difference? Could it be negative before?

vbgl: It could be negative before. What used to be printed as `-1` will now be shown as `255`. If there is never a need for printing a sign, each numbers needs at most three characters. So we have shorter lines.

vbgl:
> it means the same thing in assembly language.

Here is a link to some scarce documentation: https://sourceware.org/binutils/docs/as/Byte.html

compiler/src/asm_utils.ml
eponier: Fine

PR #1298, opened 2025-10-14
Add support for shifts & negation in array sizes

### LINKED ISSUE #1296: Allow `<<` between constants in array sizes
Currently this code:
```jazz
param int x = 2;
param int y = 1 << x;

export fn f(reg ptr u64[y] v) {

}
```
Gives this error:
```
param expansion: operator << not allowed in array size (only standard arithmetic operators and modulo are allowed)
```

vbgl:
# Description

Add support for left & right shifts in array sizes, as well as negation. This covers all operators that consume and produce integer values.

Fixes #1296

Done:

 - [x] add support for right shifts
 - [x] double-check what other operators are missing

# Checklist


sarranz (approved):
Thanks!

PR #1301, opened 2025-10-21
Rework printing of variable declarations

### LINKED ISSUE #1145: Printing is wrong when several variables have the same name
Even if the variables are originally in different scopes (hence no warning), the printer hoist them to a common scope.

The program:

~~~
export fn main(reg ptr u8[1] arg) {
	if (0 == 0) {
		reg u32 x = 0;
	} else {
		reg ptr u8[1] x = arg;
	}
}
~~~

is printed as:

~~~
export
fn main (reg const ptr u8[1] arg) -> () {
  reg u32 x;
  reg mut ptr u8[1] x;

  if (0 == 0) {
    x = ((32u) 0); /* u32 */
  } else {
    x = arg; /* u8[1] */
  }
  return ();
}
~~~

This last program is ill-typed.

vbgl:
# Description

Naive printing of variable declarations at the function entry is wrong when different variables have the same name. The position where to declare each variable is now computed from its occurrences.

A variable declaration is now printed right before its first use, in the deepest scope that surrounds all its occurrences.

Fixes #1145

# Checklist

### DIFF
--- a/changes/02-bugfix/1301-print-declarations.md
+++ b/changes/02-bugfix/1301-print-declarations.md
@@ -1,0 +1,5 @@
+- Printing before parameter substitution now prints variable declarations close
+  to their first use, in the deepest possible scope (they used to be printed at
+  the top of the function body).
+  ([PR #1301](https://github.com/jasmin-lang/jasmin/pull/1301);
+  fixes [#1145](https://github.com/jasmin-lang/jasmin/issues/1145)).
--- a/compiler/src/printer.ml
+++ b/compiler/src/printer.ml
@@ -225,33 +225,34 @@ let pp_return_type pp_size fmt =
   in
   F.fprintf fmt "%a" (pp_list ",@ " pp)
 
-let pp_gfun ~debug pp_info (pp_size:F.formatter -> 'size -> unit) pp_opn pp_var fmt fd =
+let pp_gfun ~debug (pp_size:F.formatter -> 'size -> unit) pp_opn pp_var fmt fd =
+  let ds = ScopeTree.get_declaration_sites fd in
   let pp_vd =  pp_var_decl pp_var pp_size in
-  let pp_locals fmt fd =
-    let seen = ref Spv.empty in
-    let mark x = seen := Spv.add x !seen in
-    let is_seen x = Spv.mem x !seen in
-    List.iter mark fd.f_args;
-    fold_vars_fc (fun x () ->
-        if x.v_kind <> Const && not (is_seen x) then (
-          mark x;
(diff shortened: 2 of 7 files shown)


eponier (commented):
This is smart, but I wonder if this is necessary to be that smart. From a high-level point of view, it seems that it would have been simpler to just print the declarations where they are in the source program. What did I miss?

Also I can see two peculiar cases in the current implementation:
- the program
```
  reg u64 res;
  return res;
```
which makes no sense is printed without the declaration, which makes even less sense

- if a variable is declared but not used, it disappears

These are non-blocking because the compiler now produces a warning anyway.

eponier:
Also I'm a bit worried about the performance aspect on really large programs, but there are probably other places in the compiler where it will make a difference first.

vbgl:
The example with an empty program that returns a fresh variable is a true regression. Should we really fix it?

eponier:
> > just print the declarations where they are in the source program
> 
> Where is this information available?

At the moment, it is not. I just wonder if this would have been a (probably larger but) more natural patch.

> > ```
> > * if a variable is declared but not used, it disappears
> > ```
> 
> I think this is already the case.

Indeed.

> > Also I'm a bit worried about the performance aspect on really large programs
> 
> Note that this code is usually not executed in common uses of the compiler. It is only executed when calling `-ptyping`.

Indeed.

eponier:
> The example with an empty program that returns a fresh variable is a true regression. Should we really fix it?

I agree the case is so strange that we could leave it broken. Or here is a short (completely ad-hoc) fix. Do as you wish.
```
diff --git i/compiler/src/printer.ml w/compiler/src/printer.ml
index 7f8eeb3f5..a5c82f6a3 100644
--- i/compiler/src/printer.ml
+++ w/compiler/src/printer.ml
@@ -235,6 +235,14 @@ let pp_gfun ~debug (pp_size:F.formatter -> 'size -> unit) pp_opn pp_var fmt fd =
   let ret = List.map L.unloc fd.f_ret in
   let set_var_type x ty = GV.mk x.v_name x.v_kind ty x.v_dloc x.v_annot in
   let pp_ret fmt () =
+    (* In the rare case where there is no instruction in the function,
+       the results that are not arguments are not get properly declared
+       by [pp_info]. We fix it here.
+       These results are not initialized, so the function has no semantics,
+       but at least printing is correct. *)
+    if Miloc.is_empty ds then
+      List.iter (fun x ->
+        if not (List.mem x fd.f_args) then F.fprintf fmt "%a;@ " pp_vd x) ret;
     F.fprintf fmt "return @[(%a)@];"
       (pp_list ",@ " pp_var) ret in
```

PR #1307, opened 2025-11-06
More assembler directives about function symbols

### LINKED ISSUE #1302: Add more assembler directives for ARMv7
A coworker found that recent versions of GCC require more directives than what Jasmin produces for ARMv7. In particular,

`.thumb_func`
> The .thumb_func directive tells the assembler that the next symbol will point to Arm THUMB code
> It should be used for all functions labels.
>  This information is necessary in order to allow the assembler and linker to generate correct code for interworking between Arm and Thumb instructions and should be used even if interworking is not going to be performed. 

`.type`
> The .type directive allows you to tell the assembler what type a symbol is. Most of the time we just use %function and %object

(Taken from https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/useful-assembler-directives-and-macros-for-the-gnu-assembler and https://sourceware.org/binutils/docs/as/ARM-Directives.html.)

I'm having some trouble finding specific documentation on the requirements, but I see that GCC and Clang always generate
```
        .thumb_func
        .type   foo, %function
foo:
```

vbgl:
# Description

Add `.thumb_func` and `.type` directives before export functions in assembly listings.

Fixes #1302

# Checklist

### DIFF
--- a/compiler/src/pp_arm_m4.ml
+++ b/compiler/src/pp_arm_m4.ml
@@ -69,11 +69,6 @@ let pp_asm_arg (arg : (register, Arch_utils.empty, Arch_utils.empty, rflag, cond
 
 (* -------------------------------------------------------------------- *)
 
-(* TODO_ARM: Review. *)
-let headers = [ Instr (".thumb", []); Instr (".syntax unified", []) ]
-
-(* -------------------------------------------------------------------- *)
-
 let pp_set_flags opts = if opts.set_flags then "s" else ""
 
 (* We assume the only condition in the argument list is the one we need to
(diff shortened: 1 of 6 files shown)


compiler/src/pp_arm_m4.ml
```diff
@@ -69,11 +69,6 @@ let pp_asm_arg (arg : (register, Arch_utils.empty, Arch_utils.empty, rflag, cond
 
 (* -------------------------------------------------------------------- *)
 
-(* TODO_ARM: Review. *)
-let headers = [ Instr (".thumb", []); Instr (".syntax unified", []) ]
```
sarranz, line 73: Are these unnecessary?

vbgl: This is dead code.

sarranz: I see, they are defined below.

PR #1316, opened 2025-11-13
Refactor safety checker to support multiple architectures through functorization

clebreto:
## Description

This PR refactors the Jasmin safety checker to support multiple architectures (X86-64, ARMv7-M, and RISC-V) through a generic architecture abstraction layer. Previously, the safety checker was tightly coupled to X86-64 specific operations. This refactoring introduces a functor-based design that allows the safety checker to work with any architecture by implementing a common `SafetyArch` module type.

## Motivation

The existing safety checker implementation was hardcoded for X86-64 architecture, making it difficult to extend Jasmin's safety analysis to other supported architectures like ARM and RISC-V. This limitation prevented users from leveraging the safety checker's capabilities (such as array bounds checking, memory safety analysis, and constant-time verification) when targeting non-x86 platforms.

### Architecture Abstraction Layer
- **New module type `SafetyArch`** (`safetyArch.mli`): Defines the interface that each architecture must implement, including:
  - `is_comparison`: Identifies comparison operations that set flags
  - `is_conditional`: Identifies conditional expressions (`CMOVcc` on x86, to be instantiated for ARM)
  - `split_asm_opn`: Breaks down operations into result expressions
  - `post_opn`: Defines post-conditions for operations that cannot be precisely expressed
  - `opn_heur`: Provides heuristics for flag-based control flow analysis
  - `flags_heur`: Type for tracking flag computations (ZF, CF)

- **Architecture implementations**:
  - `x86_safety`: X86-64 architecture (migrated from hardcoded logic)
  - `arm_safety`: ARMv7-M architecture implementation
  - `riscv_safety`: RISC-V architecture implementation


compiler/safetylib/safetyUtils.ml
```diff
@@ -92,6 +92,12 @@ let rec fold_left3 f accu l1 l2 l3 =
   | a1::l1, a2::l2, a3::l3 -> fold_left3 f (f accu a1 a2 a3) l1 l2 l3
   | _ -> raise (Invalid_argument "fold_left3")
     
+(*------------------------------------------------------------*)
+(* Sequence extraction helpers *)
+let as_seq1 = function [e] -> e | _ -> assert false
+let as_seq2 = function [e1;e2] -> (e1,e2) | _ -> assert false
+let as_seq3 = function [e1;e2;e3] -> (e1,e2,e3) | _ -> assert false
+
```
eponier: already exists in utils.ml

compiler/safety/run.ml
```diff
@@ -31,11 +31,13 @@ let load_file name =
   |> Compile.preprocess Arch.reg_size Arch.asmOp
 
 let analyze ~fmt pd asmOp source_f_decl f_decl p =
-  let module AbsInt = SafetyInterpreter.AbsAnalyzer (struct
+  let module PW = struct
+    type extended_op = X86_extra.x86_extended_op
```
eponier: `pd` and `asmOp` should be in the functor. `PW`, on the contrary, should be normal function arguments.

compiler/safetylib/safetyPreanalysis.ml
```diff
@@ -143,28 +143,27 @@ end
 (* Pre Analysis *)
 (****************)
 
-module Pa : sig
+module type PreAnalysisSig = sig
```
eponier: We could have kept `Pa` and `FSPa` as modules and functorize the whole file.

compiler/src/dune
```diff
@@ -11,10 +11,10 @@
  (name jasmin)
  (public_name jasmin.jasmin)
  (flags (:standard -w -9-27-32-33-34-37-39-50-67))
- (modules :standard \ main_compiler x86_safety)
+ (modules :standard \ main_compiler x86_safety arm_safety riscv_safety)
```
eponier: When we get rid of `PW`, `x86_safety` and the like should be removed, because they won't be of any use anymore.

compiler/safetylib/safetyArch.ml
```diff
@@ -0,0 +1,48 @@
+open Jasmin
```
eponier, line 1: `pd` and `asmOp` should be in the functor rather than function arguments. Apart from the that, the signature is convincing.

compiler/safetylib/safetyUtils.ml
clebreto: Should be addressed by 0a1876336127eb045dbff9b1f657717281eaa8de

compiler/safety/run.ml
eponier: In the end, we kept `PW` because it was hard to get rid of.

compiler/safetylib/safetyPreanalysis.ml
eponier: done

compiler/src/dune
eponier: We kept `PW` but still removed `x86_safety`.

compiler/safetylib/safetyArch.ml
eponier: done

eponier:
Should be ready. I updated the initial message of the PR to reflect the current status.

PR #1328, opened 2025-11-20
Add declassification operators

### LINKED ISSUE #746: Declassify annotation lost after inlining.
The following example passes CT checking when done on the default compiler pass. However, when CT checking is done after lowering `-checkCTafter=lowering` (with this #736 PR) it fails.

```jasmin
inline fn something(#secret reg u64 u) -> #secret reg u64
{
  reg u64 a;
  a = u;
  a *= 45;
  return a;
}

export fn thing() -> #[public] reg u64
{
  reg u64 b;
  reg u64 u;

  #declassify b = something(u);

  if (b == 0) {
    b += 1;
  } else {
    b += 2;
  }
  return b;
}
```

The error message is:
```
/* Security types:
*/
"", line -1 (-1):
constant type checker: __of__ has type secret it needs to be public
```

Furthermore, when one prints all of the passes, it is clear that the declassify gets lost:

```shell
jasminc -pall t.jazz | grep -E "After|declassify"
/* After typing */
  #[inline, declassify]
/* After param expansion */
  #[inline, declassify]
/* After array copy */
  #[inline, declassify]
/* After add array initialisation */
  #[inline, declassify]
/* After lower spill/unspill instructions */
  #[inline, declassify]
/* After inlining */
/* After remove unused function */
/* After unrolling */
/* After remove unused function */
/* After liverange splitting */
/* After variable renaming to remove copies */
/* After remove phi nodes introduced by splitting */
/* After dead code after variable renaming to remove copies */
/* After remove array initialisation */
/* After add assignments before and after call to ensure that arguments and results are ref ptr */
/* After expansion of register arrays */
/* After liverange splitting */
/* After variable renaming to remove copies */
/* After remove phi nodes introduced by splitting */
/* After dead code after variable renaming to remove copies */
/* After remove globals variables */
/* After lowering of instructions */
/* After selective load hardening lowering of instructions */
/* After propagate inline variables */
/* After stack allocation */
/* After remove unused returned values */
/* After register allocation */
/* After dead code after register allocation */
/* After linearization */
/* After stack zeroization */
/* After tunneling */
```

A quick fix is to declassify an assignment after the call:
```jasmin
b = something(u);
#declassify b = b;
```

However, when checking the above after the inlining pass a different error is emitted, which I am not sure about. Just pointing out that different compiler passes in the CT checker influence its decisions.:
```
/* Security types:
something(#secret u) -> #secret
*/
"", line -1 (-1):
constant type checker: u has type #strict #secret it cannot receive a value of type #public
```

vbgl:
# Description

Fixes #746

The current handling of declassification through an annotation has flaws:

1. what is being declassified is not clear
2. it sometimes require to add extra assignments only to bear the `declassify` annotation
3. annotations may be duplicated or erased during compilation

This PR introduces a dedicated (pseudo)-operator which applies to a single value, has no other effect, and is preserved by compilation until asm-gen (future work: leave a visible trace in the assembly, e.g., a comment).

Compilation of declassification of arrays requires a specific treatment during stack-allocation and an other (pseudo)-operator expressing the declassification of a memory range. This `declassify_mem` operator is not exposed to users (and therefore not taken into account by the constant-time checkers).

# Checklist

### DIFF
--- a/changes/01-feature/1328-declassify-operator.jazz
+++ b/changes/01-feature/1328-declassify-operator.jazz
@@ -1,0 +1,5 @@
+- Declassification can be expressed using a pseudo-operator that applies to a
+  single variable: `() = #declassify(x)`; the legacy annotation `#[declassify]`
+  on assignements is deprecated
+  ([PR #1328](https://github.com/jasmin-lang/jasmin/pull/1328);
+  fixes [#746](https://github.com/jasmin-lang/jasmin/issues/746)).
--- a/compiler/safetylib/safetyInterpreter.ml
+++ b/compiler/safetylib/safetyInterpreter.ml
@@ -397,6 +397,7 @@ let safe_opn pd asmOp safe opn es =
   let id =
     Sopn.get_instr_desc
       pd
+      Arch.msf_size
       asmOp
       opn
   in
@@ -1333,13 +1334,13 @@ end = struct
   (* -------------------------------------------------------------------- *)
   let num_instr_evaluated = ref 0
 
-  let print_ginstr pd asmOp ginstr abs_vals =
+  let print_ginstr pd msfsize asmOp ginstr abs_vals =
     Format.eprintf "@[<v>@[<v>%a@]@;*** %d Instr: nb %d, %a %a@;@;@]%!"
       (AbsDom.print ~full:true) abs_vals
       (let a = !num_instr_evaluated in incr num_instr_evaluated; a)
       ginstr.i_info.i_instr_number
       L.pp_sloc ginstr.i_loc.L.base_loc
-      (Printer.pp_instr ~debug:false pd asmOp) ginstr
+      (Printer.pp_instr ~debug:false pd msfsize asmOp) ginstr
 
   let print_binop fmt (cpt_instr,abs1,abs2,abs3) =
     Format.fprintf fmt "@[<v 2>Of %d:@;%a@]@;\
@@ -1354,7 +1355,7 @@ end = struct
   let print_if_join cpt_instr ginstr labs rabs abs_r =
     Format.eprintf "@;@[<v 2>If join %a for Instr:@;%a @;@;%a@]@."
       L.pp_sloc ginstr.i_loc.L.base_loc
-      (Printer.pp_instr ~debug:false Arch.pointer_data Arch.asmOp) ginstr
+      (Printer.pp_instr ~debug:false Arch.pointer_data Arch.msf_size Arch.asmOp) ginstr
       (print_binop) (cpt_instr,
                      labs,
                      rabs,
@@ -1409,7 +1410,7 @@ end = struct
   let rec aeval_ginstr : ('ty,minfo,Arch.extended_op) ginstr -> astate -> astate =
     fun ginstr state ->
       debug (fun () ->
-        print_ginstr Arch.pointer_data Arch.asmOp ginstr state.abs);
+        print_ginstr Arch.pointer_data Arch.msf_size Arch.asmOp ginstr state.abs);
 
       (* We stop if the abstract state is bottom *)
       if AbsDom.is_bottom state.abs
--- a/compiler/src/ct_checker_forward.ml
+++ b/compiler/src/ct_checker_forward.ml
@@ -527,34 +527,47 @@ let get_annot ensure_annot f =
 (* -----------------------------------------------------------*)
 let sdeclassify = "declassify"
 
-let is_declassify annot =
+let is_declassify ~loc annot =
   Annot.ensure_uniq1 sdeclassify Annot.none annot <> None
+  && (warning Always loc "#[declassify] annotations are deprecated: use the #declassify() operator instead"; true)
 
-let declassify_lvl annot lvl =
-  if is_declassify annot then Public
+
+let declassify_lvl ~loc annot lvl =
+  if is_declassify ~loc annot then Public
   else lvl
 
-let declassify_lvls annot lvls =
-  if is_declassify annot then List.map (fun _ -> Public) lvls
+let declassify_lvls ~loc annot lvls =
+  if is_declassify ~loc annot then List.map (fun _ -> Public) lvls
   else lvls
+
+let declassify_expr ~loc env =
+  function
+  | Pvar { gs = Slocal ; gv } -> Env.set env gv Public
+  | _ ->
+     warning Always loc "ignored #declassify: only local variables are supported";
+     env
 
 (* [ty_instr env i] return env' such that env |- i : env' *)
 
 let rec ty_instr is_ct_asm fenv env i =
+  let loc = i.i_loc in
   let env1 =
   match i.i_desc with
   | Cassgn(x, _, _, e) ->
     let env, lvl = ty_expr ~public:false env e in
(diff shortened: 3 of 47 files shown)


changes/01-feature/1328-declassify-operator.jazz
```diff
@@ -0,0 +1,4 @@
+- Declassification can be expressed using a pseudo-operator that applies to a
+  single variable: `() = #declassify(x);`
+  ([PR #1328](https://github.com/jasmin-lang/jasmin/pull/1328);
```
bgregoir, line 4: Should we say that #[declassify] annotation is deprecated. Or we still use it?

proofs/lang/pseudo_operator.v
```diff
@@ -27,6 +27,8 @@ Canonical spill_op_eqType := @ceqT_eqType _ eqTC_spill_op.
 Variant pseudo_operator :=
 | Ospill    of spill_op & seq atype
 | Ocopy     of wsize & positive
+| Odeclassify of atype
+| Odeclassify_mem of positive
```
bgregoir, line 31: I think it is not possible to define correctly the semantics of Declassify_mem, if it is an operator because the semantics of operator is not parametrised by the memory.

compiler/tests/success/common/declassify.jazz
```diff
@@ -0,0 +1,18 @@
...
+export fn load(reg ptr u32[1] a) -> reg u32 {
+  reg u32 x = a[0];
+  () = #declassify(x);
+  return x;
+}
+
+/* The input array contains mixed-sensitivity data:
+  the beginning is secret whereas the end is public.
+  This program illustrates that only a slice can be declassified.
+*/
```
bgregoir, line 18: this is really nice

compiler/safetylib/safetyInterpreter.ml
```diff
@@ -409,6 +409,7 @@ let split_div sg ws es =
 let safe_opn safe opn es =
   let id =
     Sopn.get_instr_desc
+      X86_decl.x86_decl.reg_size
```
bgregoir: For the moment, I have not be able to understand why we have an extra reg_size

bgregoir: Maybe to define the type of Declassify_mem ?

compiler/src/ct_checker_forward.ml
```diff
@@ -538,6 +538,13 @@ let declassify_lvls annot lvls =
   if is_declassify annot then List.map (fun _ -> Public) lvls
```
bgregoir: Should we print a warning for deprecated ? Maybe it is done somewhere else

compiler/src/printCommon.ml
```diff
@@ -151,7 +151,7 @@ let string_of_op2 = function
   | Owi2(sg, ws, o) -> string_of_wiop2 sg ws o
 
 (* -------------------------------------------------------------------- *)
-let pp_opn pd asmOp fmt o = pp_string fmt (Sopn.string_of_sopn pd asmOp o)
+let pp_opn pd asmOp fmt o = pp_string fmt (Sopn.string_of_sopn pd pd asmOp o)
```
bgregoir: Again strange, the original one corresponds to msf_size ?

compiler/src/sct_checker_forward.ml
```diff
@@ -938,6 +938,13 @@ let declassify_tys env annot tys = if is_declassify annot
   then List.map (declassify env) tys
   else tys
 
+let declassify_expr ~loc env ((msf, venv) as msf_e) =
```
bgregoir, line 936: Same comment should we add a warning for deprecated in declassify_expr and declassify_tys

docs/source/tools/ct.md
```diff
@@ -211,8 +211,10 @@ possible way around this limitation is to use some “declassification”, i.e.,
 explicitly mark some data as public although the rules of the type-system would
 label them as secret.
 
-The type-checker supports a `declassify` annotation on assignments and labels
-the written values as public.
+The type-checker supports a `declassify` pseudo-operator that applies to local
+variables and does not produce any output. It tells the checker to label its
+argument as public. It also support a `declassify` annotation on assignments and
+labels the written values as public.
```
bgregoir: Should we deprecate the second version with annotation?

proofs/compiler/stack_alloc.v
```diff
@@ -1699,6 +1699,24 @@ Definition alloc_array_swap rmap rs t es :=
...
+  if es is [:: Pvar x ] then
+    Let _ := assert (is_lvar x) (stk_ierror_no_var "global reg ptr ...") in
+    let xv := gv x in
+    Let: (sr, status) := get_sub_region_status rmap xv in
+    Let _ := check_valid xv status in
+    if get_local xv is Some pk then
+      Let: (p, ofs) := addr_from_pk xv pk in
+      let e := add (Plvar p) (cast_const ofs) in
+      let len := Z.to_pos (size_of xv.(vtype)) in
+      ok (Copn [::] AT_keep (Opseudo_op (pseudo_operator.Odeclassify_mem len)) [:: e ])
```
bgregoir, line 1716: Maybe it will be more uniform is declassify_mem take a ws and a len. As the other array operators.

proofs/lang/sopn.v
```diff
@@ -222,6 +222,56 @@ Definition Ocopy_instr ws p :=
...
+    semi     := fun=> ok tt;
+    semu     := @declassify_semu cty;
+    i_safe   := [:: ];
+    i_valid  := true;
+    i_safe_wf    := refl_equal;
+    i_semi_errty := fun _ => (@sem_prod_ok_error _ [:: cty ] _ ErrType);
+    i_semi_safe  := fun _ => (@sem_prod_ok_safe _ [:: cty ] _);
+  |}.
+
+Definition Odeclassify_mem_instr len :=
```
bgregoir, line 265: How to really define the semantic ?

bgregoir: Maybe event can solve the problem, but I am not sure, because I think that the interpretation of event will not have the memory.

compiler/safetylib/safetyInterpreter.ml
vbgl: Exactly. The type of of this operator is `Uptr → ()`. But this might need to change: there sure is a better design awaiting for us to find it!

changes/01-feature/1328-declassify-operator.jazz
vbgl: The legacy annotation is now deprecated.

compiler/src/ct_checker_forward.ml
vbgl: A warning has been added.

compiler/src/sct_checker_forward.ml
vbgl: Yes, done.

docs/source/tools/ct.md
vbgl: Yes. Done.

compiler/src/printCommon.ml
vbgl: That’s true. This is addressed in the latest commit.

PR #1333, opened 2025-11-24
SCT tests: minor cleaning

vbgl:
1. make sure annotations use the preferred syntax with square brackets
2. better reporting of non-failures in negative tests


bgregoir (approved):
Thank

PR #1340, opened 2025-12-01
Print less brackets

vbgl:
Pretty-printing of Jasmin programs and expressions use less parentheses, taking into account priority and associativity of operators.

### DIFF
--- a/compiler/src/printCommon.ml
+++ b/compiler/src/printCommon.ml
@@ -242,3 +242,74 @@ let pp_var fmt x =
   fprintf fmt "%s" y.v_name
 
 let pp_var_i fmt x = pp_var fmt x.E.v_var
+
+(* -------------------------------------------------------------------- *)
+type priority =
+  | OpPmin
+  | OpPternary
+  | OpPbor
+  | OpPband
+  | OpPeq
+  | OpPcmp
+  | OpPor
+  | OpPxor
+  | OpPand
+  | OpPshift
+  | OpPsum
+  | OpPprod
+  | OpPunary
+
+let priority_min = OpPmin
+let priority_ternary = OpPternary
+
+let priority_of_wop1 = function
+  | E.WIwint_of_int _ | WIint_of_wint _ | WIword_of_wint _ | WIwint_of_word _
+  | WIwint_ext _ ->
+      OpPunary
+  | WIneg _ -> OpPsum
+
+let priority_of_op1 = function
+  | E.Oword_of_int _ | Oint_of_word _ | Osignext _ | Ozeroext _ | Onot | Olnot _
+    ->
+      OpPunary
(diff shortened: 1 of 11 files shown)


compiler/src/printCommon.ml
```diff
@@ -242,3 +242,72 @@ let pp_var fmt x =
...
+  | OpPbor | OpPband | OpPeq | OpPcmp | OpPor | OpPxor | OpPand | OpPshift
+  | OpPsum | OpPprod ->
+      Left
+  | OpPunary -> Right
+
+let same_side a b =
+  match (a, b) with Left, Left | Right, Right -> true | _, _ -> false
+
+let optparent fmt ctxt side prio p =
+  if prio < ctxt || ((not (same_side (associativity ctxt) side)) && prio = ctxt)
```
bgregoir: Maybe it is dangerous to relate on the implementation of the polymorphic comparison here.

compiler/src/printFexpr.ml
```diff
@@ -1,24 +1,36 @@
 open Fexpr
 open PrintCommon
 
-let rec pp_fexpr fmt = function
+let rec pp_fexpr side prio fmt = function
```
bgregoir, line 4: Some code that come from EC: 
```
let pp_if c pp1 pp2 fmt x =
  match c with
  | true  -> Format.fprintf fmt "%a" pp1 x
  | false -> Format.fprintf fmt "%a" pp2 x

let pp_maybe c tx pp fmt x =
  pp_if c (tx pp) pp fmt x

let pp_enclose ~pre ~post pp fmt x =
  Format.fprintf fmt "%(%)%a%(%)" pre pp x post

let pp_paren pp fmt x =
  pp_enclose ~pre:"(" ~post:")" pp fmt x

let pp_maybe_paren c pp =
  pp_maybe c pp_paren pp
```
I think then the code become simpler to read

bgregoir: But it is up to you

compiler/src/printCommon.ml
vbgl: I thought that this is a documented feature. If it is not, is there an elegant way of achieving the same thing?

compiler/src/printFexpr.ml
vbgl: I’ll give it a try, thanks for the suggestion.

vbgl: In the end, I’m not sure about what you were suggesting. What’s your opinion about the commit I’ve just added?

bgregoir: I think it is good, from time to time error message are harder to read without parenthesis but globally it is cleaner

vbgl: I agree.

PR #1341, opened 2025-12-03
wint-int: handle (un)-spill operators

### LINKED ISSUE #1335: EC extraction: crash when spilling word-int
The following function causes `jasmin2ec` to crash:

~~~
export fn wint() {
  reg ui32 x = 0;
  () = #spill(x);
}
~~~

> "spill.jazz", line 24 (2-17):
internal compilation error in function wint:
  wint_to_int: invalid expr in Copn

vbgl:
# Description

Process arguments to spill / unspill in the wint-int pass.

Remove the wint-int-proof that is unused and not relevant (we need backward, not forward).

Fixes #1335.

# Checklist


proofs/compiler/lower_spill.v
```diff
@@ -1,7 +1,7 @@
 (* ** Imports and settings *)
 From mathcomp Require Import ssreflect ssrfun ssrbool eqtype.
 From Coq Require Import ZArith Uint63.
-Require Import pseudo_operator expr compiler_util.
+Require Import sopn pseudo_operator expr compiler_util.
```
bgregoir, line 4: Why sopn is now required ?

PR #1342, opened 2025-12-04
Enable wints as argument to #swap

vbgl:
# Description

Currently, pre-typing reject applications of `#swap` to word-sized integers (for no good reason, AFAICT).

Apart from relaxing a test in pre-typing, this PR adapts the wint-int pass to properly handle swap operators.

# Checklist

### DIFF
--- a/compiler/tests/success/common/swap.jazz
+++ b/compiler/tests/success/common/swap.jazz
@@ -51,3 +51,10 @@ export fn truncate(reg u32 x) -> reg u32, reg u16 {
 export fn nolhs(reg u32 x y) {
   #[keep] _, _ = #swap(x, y);
 }
+
+export fn wint(reg ui32 a) -> reg ui32 {
+  reg ui32 b = 0;
+  a, b = #swap(a, b);
+  a += b;
+  return a;
+}
--- a/proofs/compiler/wint_int.v
+++ b/proofs/compiler/wint_int.v
@@ -22,10 +22,10 @@ Module Import E.
   Definition ierror := pp_internal_error pass.
 
   Definition ierror_e e :=
-    ierror (pp_nobox [:: pp_s "ill typed expression"; pp_e e]).
+    ierror (pp_nobox [:: pp_s "ill typed expression "; pp_e e]).
 
   Definition ierror_lv lv :=
-    ierror (pp_nobox [:: pp_s "ill typed left value"; pp_lv lv]).
+    ierror (pp_nobox [:: pp_s "ill typed left value "; pp_lv lv]).
 
 End E.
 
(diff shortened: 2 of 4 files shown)


compiler/tests/success/common/swap.jazz
```diff
@@ -51,3 +51,10 @@ export fn truncate(reg u32 x) -> reg u32, reg u16 {
 export fn nolhs(reg u32 x y) {
   #[keep] _, _ = #swap(x, y);
 }
+
+export fn wint(reg ui32 a) -> reg ui32 {
+  reg ui32 b = 0;
+  a, b = #swap(a, b);
+  a += b;
+  return a;
```
bgregoir, line 59: this is an interesting program for register allocation :-)

bgregoir: At leat on x86, does the compiler introduce extra assignment?

proofs/compiler/wint_int.v
```diff
@@ -278,33 +278,45 @@ Fixpoint wi2i_ir (ir:instr_r) : cexec instr_r :=
...
+          let etys := map etype_of_expr es in
+          let tys' := map2 (fun ety ty => to_etype (sign_of_etype ety) ty) etys tys in
+          Let _ := assert (size tys == size es) (E.ierror_s "ill typed spill") in
+          Let _ := assert (all2 esubtype tys' etys) (E.ierror_s "ill typed spill (arguments)") in
+          (* We patch the type of the operator *)
+          let tys := map2 (fun ty e => wi2i_type (sign_of_expr e) ty) tys es in
+          Let es := mapM wi2i_e es in
+          ok (Copn [::] t (Opseudo_op (Ospill k tys)) es)
+      | Opseudo_op (Oswap ty) =>
+          if es is [:: e1; e2 ] then
```
bgregoir, line 293: Maybe the code can be follow more the pattern used for Ccall.
let ety := etype_of_expr m e1 in
let sig := [ety; ety] in
Let _ := assert (all2 (fun ety e => esubtype ety (etype_of_expr m e)) sig.1 es)
                           (E.ierror_s "invalid args in Copn") in
Let es := wi2i_es wi2i_e es in
Let xs := wi2i_lvs "invalid dest in Copn" true sig xs in
let ty := wi2i_type (sign_of_expr e1) ty in
ok (Copn xs t (Opseudo_op (Oswap ty)) es)

I think the proof will be simpler, because existing lemmas can be reused.

proofs/compiler/wint_int.v
```diff
@@ -278,33 +278,45 @@ Fixpoint wi2i_ir (ir:instr_r) : cexec instr_r :=
...
+          ok (Copn xs t o es)
+      end
 
   | Csyscall xs o es =>
     Let _ := assert (all (fun e => sign_of_expr e == None) es)
                     (E.ierror_s "invalid args in Csyscall") in
     Let es := mapM wi2i_e es in
     let xtys := map (to_etype None) (syscall_sig_u o).(scs_tout) in
-    Let xs := mapM2 (E.ierror_s "invalid dest in Copn") wi2i_lv xtys xs in
+    Let xs := mapM2 (E.ierror_s "invalid dest in Csyscall") wi2i_lv xtys xs in
```
bgregoir, line 318: Nice catch

compiler/tests/success/common/swap.jazz
vbgl: Yes:

~~~
wint:
	movl	$0, %eax
	xchg	%eax, %edi
	addl	%eax, %edi
	movl	%edi, %eax
	ret
~~~

proofs/compiler/wint_int.v
vbgl: Done. Thanks for the suggestion.

PR #1349, opened 2025-12-10
Document namespaces

eponier:
# Description

Add some documentation for namespaces

(plus add a copy button next to code blocks in the documentation)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 sphinx-rtd-theme
 myst-parser
+sphinx-copybutton
--- a/docs/shell.nix
+++ b/docs/shell.nix
@@ -6,6 +6,6 @@ mkShell {
   packages = [
     graphviz
     (python3.withPackages
-      (ps: [ ps.myst-parser ps.sphinx ps.sphinx-rtd-theme ]))
+      (ps: [ ps.myst-parser ps.sphinx ps.sphinx-rtd-theme ps.sphinx-copybutton ]))
   ];
 }
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,7 +13,7 @@ author = "The Jasmin developer team"
 # -- General configuration ---------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
 
-extensions = ["myst_parser", "sphinx_rtd_theme", "sphinx.ext.graphviz"]
+extensions = ["myst_parser", "sphinx_rtd_theme", "sphinx.ext.graphviz", "sphinx_copybutton"]
 
 myst_enable_extensions = ["colon_fence"]
 myst_heading_anchors = 3


vbgl:
I think it must be documented, otherwise you cannot have an export function within a namespace.

docs/source/language/syntax/structure.md
```diff
@@ -171,3 +173,54 @@ is a local function that copies the contents of a register `x` to an address
...
+      return x;
+    }
+  }
+}
+```
+defines `f` inside the namespace `B`, which is itself inside the namespace `A`. To refer to `f` outside `A`,
+one would write `A::B::f`.
+
+A namespace can be closed and reopened, but accesses to objects defined in previous blocks
+of the same namespace must be prefixed.
```
vbgl, line 212: Unrelated to this PR: this “must” looks odd. Should we fix name resolution so as to be able to refer to objects defined in the current namespace without prefix?

docs/source/misc/faq.md
```diff
@@ -59,3 +59,109 @@ while (i < 256) {
     i += 1;
 }
 ```
+
+## How to achieve code reuse?
+
+There is ongoing work to have a proper module system in Jasmin,
+but for the time being the only solution is to use `require` directives.
```
vbgl: I would not claim this is “the only solution”. Don’t underestimate user creativity.

docs/source/language/syntax/structure.md
eponier: Indeed, this seems a bug to me. I discovered this when I played with namespaces to know what to write in the documentation.

I've just played a bit with C++ to check the differences with Jasmin. Indeed, in C++, reopening a namespace also makes all the names defined in that namespace accessible without prefix. But there is a second difference with what we do in Jasmin: one can always use the long name. For instance the following is valid C++ (but the equivalent in Jasmin is not valid).
```
namespace A {
  void f () {};
  void g () {
    A::f ();
  }
}
```
I don't know if we want that in Jasmin too. One good property of this is that it doesn't make a difference to have everything in one namespace block or split in several blocks. We don't have this property in Jasmin at the moment, but even opening the scope when we reopen a namespace won't be enough, there will still be a difference, since
```
namespace A {
  fn f () {}
}
namespace A {
  fn g () {
    () = A::f();
  }
}
```
would be valid, but
```
namespace A {
  fn f () {}
  fn g () {
    () = A::f();
  }
}
```
would not.

vbgl: Thanks. Let’s address these issues in a separate PR.

docs/source/misc/faq.md
eponier: I rewrote the beginning, because I also want to mention the `require` system for its original purpose.

I pushed an additional commit, I'll squash it when we agree on the PR.

eponier: With a correct link, this is better...

eponier:
done (and rebased)

vbgl (approved):

PR #1356, opened 2025-12-16
Trusted automated spilling

vbgl:
# Description

This adds a trusted auto-spill transformation, that is naive but systematic. It features two modes that can be fine-tuned using annotations.

# Checklist

### DIFF
--- a/docs/source/compiler/passes/auto_spill.md
+++ b/docs/source/compiler/passes/auto_spill.md
@@ -1,0 +1,57 @@
+# Automated spilling
+
+This is an experimental feature of the `jasminc` command-line tool, enabled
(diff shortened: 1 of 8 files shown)


bgregoir (commented):
This is very nice, I think it will be very useful for teaching/learning Jasmin

docs/source/compiler/passes/auto_spill.md
```diff
@@ -0,0 +1,57 @@
...
+  y = arg[1];
+  z = x ^ y;
+  return z;
+}
+~~~
+
+Automated spilling can be disabled in a whole function by annotating said
+function as `nospill`.
+
+A `msf` annotation is treated as a `nospill` annotation.
```
bgregoir, line 57: nice catch

PR #1360, opened 2025-12-17
doc: add auto_spill to the list of passes

eponier:
The documentation page of auto spill was not referenced from anywhere.

### DIFF
--- a/docs/source/compiler/passes/passes.dot
+++ b/docs/source/compiler/passes/passes.dot
@@ -11,7 +11,6 @@ preprocess[label="Preprocess",fillcolor=moccasin];
 typing[label="Type-check",style="filled,dotted",fillcolor=moccasin];
 jasmin[label="Jasmin",shape=oval,fillcolor=white];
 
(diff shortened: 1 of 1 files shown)


docs/source/compiler/passes/passes.dot
```diff
@@ -62,9 +63,10 @@ subgraph line1 {
   parse -> preprocess;
   preprocess -> typing;
   typing -> jasmin;
+  jasmin -> autospill;
 }
 
-jasmin -> wintword;
+autospill -> wintword;
```
vbgl: I’d put autospill before jasmin.

eponier: I just followed the code, should we move `do_auto_spill` before
```
(* The source program, before any compilation pass. *)
let source_prog = prog in
```
?

vbgl: The auto-spill feature will undergo some refactoring soon. In particular I’ll move it later in the compilation pipeline.

eponier: So what should I do with this PR? Should I just wait for the refactoring?

PR #1370, opened 2026-01-21
Parser: allow expressions in annotations

vbgl:
This change has been extracted from #1170.


eponier:
If I understand correctly, this duplicates type `annotations`. Is it on purpose that the old one is not removed?

PR #1371, opened 2026-01-23
Enable strict warnings for handwritten OCaml

eponier:
Before this PR, we were forced to turn off some OCaml warnings, because we did not have an easy way to disable some only for files extracted from Rocq. The solution was [OCaml code injection!](https://github.com/rocq-prover/rocq/issues/20193#issuecomment-2688604159)

I left warning 9 because enabling it implied to do a lot of changes, and I am not sure it is worth in practice. But I do not have a strong opinion either, so if someone thinks enabling warning 9 is important, I will do it.

### DIFF
--- a/compiler/src/dune
+++ b/compiler/src/dune
@@ -14,7 +14,7 @@
 (library
  (name jasmin)
  (public_name jasmin.jasmin)
- (flags (:standard -w -9-27-32-33-34-37-39-50-67))
+ (flags (:standard -w -9))
  (modules :standard \ main_compiler)
  (libraries angstrom batteries.unthreaded zarith menhirLib uint63))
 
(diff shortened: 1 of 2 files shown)


compiler/src/toEC.ml
```diff
@@ -352,7 +352,7 @@ type ec_module = {
 type ec_item =
     | IrequireImport of string list
     | Iimport of string list
-    | IfromImport of string * (string list)
+    | IfromImport of string * (string list) [@warning "-37"]
```
eponier: Here I decided to preserve the type, but if we think that removing the constructor is a better fix, I can do that instead.

vbgl: The annotation is not working as intended (see CI failure). It might be OK to drop the unused constructor.

eponier: Ok I understood the issue. The per-constructor annotation exists only since 5.1 (https://github.com/ocaml/ocaml/commit/8712acc51cd7b3d83dbf0fb756511afcff5090a0). Either I use the type-level annot (with two @), or I remove the constructor. What do you think?

vbgl: Remove constructor. Benjamin, any opinion?

eponier: I removed the constructor.

PR #1372, opened 2026-01-26
No ring in extraction

eponier:
# Description

Following changes in mathcomp (PR https://github.com/math-comp/math-comp/pull/1256, see https://github.com/jasmin-lang/jasmin/pull/1093 for discussions about it), extraction to OCaml fails when encountering a ring structure from mathcomp. In particular, Jasmin is not compatible with mathcomp 2.5.

This PR removes the ring structures from all the problematic places. Most of the removed instances involve the ring structure of words, but a few occurrences of the ring structure of `Z` are removed too.

This PR goes together with changes to coqword that I'm about to push. CI will tell if it works with the base version of coqword.

This is on top of https://github.com/jasmin-lang/jasmin/pull/1371, just to avoid the conflict between both PRs.

### DIFF
--- a/proofs/lang/word.v
+++ b/proofs/lang/word.v
@@ -2,7 +2,8 @@
 
 (* ** Imports and settings *)
 
-From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat seq eqtype tuple.
+From HB Require Import structures.
+From mathcomp Require Import ssreflect ssrfun ssrbool ssrnat seq eqtype choice tuple.
 From mathcomp Require Import div fintype order ssralg ssrnum word_ssrZ word.
 From Coq Require Zquot.
 From Coq Require Import ZArith.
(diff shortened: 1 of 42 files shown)


eponier:
>     * Now, words are extracted in OCaml into `coq_Z` instead of `Obj.t`. While arguably nicer, this may need some care to avoid using them as `coq_Z` and use only functions on words instead. It would be better if types were different. We may want to use the same kind of hack as we use for syscalls.

What about this though? Do we want to use the trick at extraction?

proofs/lang/word.v
```diff
@@ -176,42 +177,90 @@ Coercion nat_of_pelem (pe: pelem) : nat :=
   | PE128 => nat_of_wsize U128
   end.
 
-Definition word sz : comRingType := (wsize_size_minus_1 sz).+1.-word.
+Definition word sz : Type := (wsize_size_minus_1 sz).+1.-word.
+HB.instance Definition _ sz := Equality.on (word sz).
+HB.instance Definition _ sz := GRing.ComRing.on (word sz).
```
vbgl, line 182: Why is this needed?

eponier: This is meant to copy the structures from `mathcomp.word.word` to `Jasmin.word`. But since I don't really understand what I'm doing when I have to touch HB things, I don't claim this is the right way to do it, nor really needed.

Additionally, if I don't add the first line, an Equality instance is inferred from the ring structure, meaning the ring structure appears each time we use `==` on words. This means we cannot use `==` on words in extracted parts of Jasmin (otherwise extraction fails with mathcomp 2.5). With the first line, this is not the case.

vbgl: Thanks. I think we can also `Import` `mathcomp.word` and `word_ssrZ`  where needed but it is simpler as implemented.

vbgl: Also, there is no real duplication: this is only an alias (with a cool name: `HB_unnamed_factory_2`).

PR #1375, opened 2026-01-30
Literal byte arrays

### LINKED ISSUE #510: Add constant string literals
This is a **feature request** to add constant string literals.

As mentioned in #507, it is desired to define u8 arrays by string literals. A example for this could look like this:

```
string test = "Teststring";
```

The resulting string `test` should then internally be treated as being of type u8[10]. Non-ascii characters should also be representable. For example by `\xff` with `ff` being the hex value of the character.

Some functions might expect null terminated strings. For example system calls. The compiler must keep track of this function and display an error message if an input string to that function is not null terminated.

vbgl:
# Description

This adds literal (byte) arrays to the language of expressions. In source code they can be written using string syntax (e.g., `"data"`) or array syntax (e.g., `{ x, y + z }`). Both syntax were already accepted for top-level definitions of global variables.

Fixes #510

~This PR is marked as draft as it needs a bit of polishing (see checklist below).~

Future work (not for this PR): generalize the type of cells (currently limited to `u8`).

# Checklist


vbgl:
I’ve doubts about the `rev` in the semantics: which operator does cancel it? Either `curry` or `WArray.fill`.

vbgl:
1. PR #1170 uses them.
2. “scoped globals” (the one declared within an inline function) are currently restricted to word

eponier:
What should happen in these cases? `reg u8[3] r = { 1, 2, 3 };` and `stack u8[3] s = { 1, 2, 3 }`? For now, we get an error in array expansion for the first one, and an assert false for the second one.

proofs/compiler/constant_prop.v
```diff
@@ -288,10 +288,14 @@ Definition app_sopn := app_sopn of_expr.
...
 
 Definition s_opN (op:opN) (es:pexprs) : pexpr :=
+  if op is Oarray len then
+    PappN op es
+  else
   match app_sopn _ (sem_opN_typed op) es with
   | Ok r =>
     match op return sem_t (eval_atype (type_of_opN op).2) -> _ with
     | Opack ws _ => fun w => Papp1 (Oword_of_int ws) (Pconst (wunsigned w))
+    | Oarray len => fun _ => Pbool false
```
eponier: To be sure I understand, this is dead code and would typically be an `assert false` if it were possible?

vbgl: Yes, this is dead code. I’m not sure there is a more elegant way to write the same code (here, there are two nested pattern-matching twice on the same value).

docs/source/language/syntax/expressions.md
```diff
@@ -4,11 +4,11 @@
...
   | <var>  // Variable.
-  | (<wsize>)[<var> + <expr>] // Memory access.
-  | (<wsize>)[<var> - <expr>] // Memory access.
+  | [<expr>] // Memory access.
+  | [:<wsize> <expr>] // Memory access.
   | <var>[<expr>]  // Array access.
-  | <var>[<wsize> <expr>]  // Array access.
-  | <var>.[<wsize> <expr>]  // Unscaled array access.
+  | <var>[:<wsize> <expr>]  // Array access.
+  | <var>.[:<wsize> <expr>]  // Unscaled array access.
```
eponier, line 14:
```suggestion
  | <var>.[<expr>]  // Unscaled array access.
  | <var>.[:<wsize> <expr>]  // Unscaled array access.
```

docs/source/language/syntax/expressions.md
```diff
@@ -4,11 +4,11 @@
...
   | <var>[<wsize> <expr> : <expr>]  // Subarray.
   | <var>.[<wsize> <expr> : <expr>]  // Unscaled subarray.
```
eponier:
```suggestion
  | <var>[<expr> : <expr>]  // Subarray.
  | <var>[:<wsize> <expr> : <expr>]  // Subarray.
  | <var>.[<expr> : <expr>]  // Unscaled subarray.
  | <var>.[:<wsize> <expr> : <expr>]  // Unscaled subarray.
```

proofs/compiler/constant_prop.v
eponier: Actually, Rocq is able to elaborate a bit. So you could write
```
Definition s_opN (op:opN) (es:pexprs) : pexpr :=
  match op, app_sopn _ (sem_opN_typed op) es with
  | Opack ws _, Ok w => Papp1 (Oword_of_int ws) (Pconst (wunsigned w))
  | Ocombine_flags _, Ok b => Pbool b
  | _, _ => PappN op es
  end.
```

eponier: (and this has no impact on the proof it seems)

vbgl: Thanks. I’ll apply your suggestion.

vbgl: Done.

PR #1382, opened 2026-02-12
subst.ml was wrong for unsigned div and rem

eponier:
It turns out that the operations provided by (OCaml) ZArith do not match the ones provided by (Rocq) ZArith...

# Checklist


eponier:
This should finally be correct...

PR #1388, opened 2026-02-19
Array lengths are now integers in Z

eponier:
# Motivation

Motivation is PR https://github.com/jasmin-lang/jasmin/pull/1325. I'm about to introduce variables and some arithmetic in array lengths. And the question is what to do when one of the variables has no associated value (impossible, but it's easier to manipulate a definition where this is allowed) or when the length expression evaluates to a non-positive integer.

# Description

The solution implemented here is to allow array types of non-positive lengths. Array values are also generalized to avoid requiring positive lengths. An array of non-positive size has no valid cell.

The change was easy to implement, except for the proof of stack allocation. The idea of the change in the proof is to notice that non-positive slots are never used for something meaningful, like reading a value in them, because they are empty. I thus weakened a lot of lemmas to mention only slots of positive sizes.

I did not remove the check introduced by https://github.com/jasmin-lang/jasmin/pull/1158, so in concrete syntax arrays of non-positive lengths are still not allowed.

I am not sure how this impacts the semantics of Jasmin. Also, I did not check whether this breaks the extraction to EC or the safety analysis.

# Alternative ideas

- use `N` rather than `Z` for the type of array lengths: negative lengths do not make a lot of sense, and for https://github.com/jasmin-lang/jasmin/pull/1325 just having arrays of length `0` should be enough, but at least in this PR everything happens in `Z` (nearly no coercions)
- define array values only for arrays of positive lengths: just an idea, but maybe we can allow types of arrays of non-positive lengths, but make sure these types are empty (no value exists for such a type)
- make `eval_atype` capable of failing: my initial idea, but @bgregoir suggested something closer to this PR, we feared that the change on the code base would be large

### DIFF
--- a/compiler/src/evaluator.ml
+++ b/compiler/src/evaluator.ml
@@ -226,19 +226,19 @@ let pp_val fmt v =
   match v with
   | Vbool b -> Format.fprintf fmt "%b" b
   | Vint z  -> Format.fprintf fmt "%a" Z.pp_print (Conv.z_of_cz z)
-  | Varr(p,t) ->
-    let ip = Conv.int_of_pos p in
+  | Varr(n,t) ->
+    let ni = Conv.int_of_cz n in
     let pp_res fmt = function
       | Ok w               -> pp_word fmt U8 w
       | Error ErrAddrUndef -> pp_undef fmt (Coq_cword U8)
       | Error _            -> assert false in
     Format.fprintf fmt "@[[";
-    for i = 0 to ip-2 do
+    for i = 0 to ni-2 do
       let i = Conv.cz_of_int i in
-      Format.fprintf fmt "%a;@ " pp_res (WArray.get p Aligned AAscale U8 t i);
+      Format.fprintf fmt "%a;@ " pp_res (WArray.get n Aligned AAscale U8 t i);
     done;
-    if 0 < ip then
-      pp_res fmt (WArray.get p Aligned AAscale U8 t (Conv.cz_of_int (ip-1)));
+    if 0 < ni then
+      pp_res fmt (WArray.get n Aligned AAscale U8 t (Conv.cz_of_int (ni-1)));
     Format.fprintf fmt "]@]";
   | Vword(ws, w) -> pp_word fmt ws w
   | Vundef ty -> pp_undef fmt ty
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -583,16 +583,23 @@ let check_array env x =
 (* ------------------------------------------------------------------- *)
 (* Formatting to string helpers *)
 
+(* we print negative lengths with a "m"-prefix to create valid module names *)
+let pp_length fmt i =
+  if i < 0 then
+    Format.fprintf fmt "m%d" (-i)
+  else
+    Format.fprintf fmt "%d" i
+
 let fmt_array_theory at = match at with
-  | Array n -> Format.sprintf "Array%i" n
-  | WArray n -> Format.sprintf "WArray%i" n
-  | ArrayWords aw -> Format.sprintf "ArrayWords%iW%i" aw.sizea (8*aw.sizew)
-  | SubArray x -> Format.sprintf "SubArray%i_%i" x.sizes x.sizeb
-  | SubArrayDirect x -> Format.sprintf "SubArrayDirect%i_%iW%i" x.sizes x.sizeb (8*x.sizew)
-  | SubArrayCast x -> Format.sprintf "SubArrayDirect%iW%i_%iW%i" x.sizes (8*x.sizews) x.sizeb (8*x.sizewb)
-  | ArrayAccessCast x -> Format.sprintf "ArrayAccessCastW%i_%iW%i" (8*x.sizews) x.sizeb (8*x.sizewb)
-  | ByteArray n -> Format.sprintf "BArray%i" n
-  | SubByteArray x -> Format.sprintf "SBArray%i_%i" x.sizeb x.sizes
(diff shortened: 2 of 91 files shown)


eponier:
This should be ready now.

> I did not remove the check introduced by #1158, so in concrete syntax arrays of non-positive lengths are still not allowed.

Now the check is weaker: only negative sizes are rejected.

> I am not sure how this impacts the semantics of Jasmin. Also, I did not check whether this breaks the extraction to EC or the safety analysis.

This is still true.

compiler/src/evaluator.ml
```diff
@@ -226,19 +226,19 @@ let pp_val fmt v =
...
     let pp_res fmt = function
       | Ok w               -> pp_word fmt U8 w
       | Error ErrAddrUndef -> pp_undef fmt (Coq_cword U8)
       | Error _            -> assert false in
     Format.fprintf fmt "@[[";
-    for i = 0 to ip-2 do
+    for i = 0 to ni-2 do
       let i = Conv.cz_of_int i in
-      Format.fprintf fmt "%a;@ " pp_res (WArray.get p Aligned AAscale U8 t i);
+      Format.fprintf fmt "%a;@ " pp_res (WArray.get n Aligned AAscale U8 t i);
```
bgregoir, line 238: Not related to this PR, but should we put Unaligned instead of Aligned here?

compiler/src/toEC.ml
```diff
@@ -1305,8 +1305,8 @@ let ty_expr = function
 let ty_sopn pd msfsz asmOp op es =
```
bgregoir, line 1333: I think, cloning of array theory, now 
require a little bit of care. The theory require that the size is greater than 0.
This was true by construction before I think we should now do the proof of the axiom when we clone, or remove the axiom in the theory (I think it is not needed).

compiler/src/typing.ml
```diff
@@ -20,9 +20,9 @@ let ty_var (x: var) =
   let ty = x.v_ty in
   begin match ty with
   | Arr(_, n) ->
-      if (n < 1) then
+      if (n < 0) then
         error (L.i_loc0 x.v_dloc)
-          "the variable %a has type %a, its array size should be positive"
+          "the variable %a has type %a, its array size must be non-negative"
```
bgregoir: This is really needed?

compiler/src/typing.ml
```diff
@@ -59,8 +59,8 @@ let check_int loc e te = check_type loc e te tint
 let check_ptr pd loc e te = check_type loc e te (tu pd)
 
 let check_length loc len =
-  if len <= 0 then
-    error loc "the length should be strictly positive"
+  if len < 0 then
+    error loc "the length must be non-negative"
```
bgregoir: Same here

bgregoir: I mean, the semantics imposes this ?

compiler/src/evaluator.ml
vbgl: With `AAscale`, that should be fine. And with `Aligned`, this is irrelevant.

eponier: I think Benjamin mentioned that only for the performance aspect. Functionally, everything is equivalent. (Maybe there are some unneeded checks with `Aligned`?)

compiler/src/typing.ml
eponier: No, but I have to check how good or bad the compiler deals with arrays of negative size.

compiler/src/evaluator.ml
vbgl: The cost of an alignment check with U8 is indeed non-zero. But it should not be a source for concern.

compiler/src/typing.ml
eponier: I removed the check, and tested a bit the compiler on arrays of negative length. I have found a few bugs or sub-optimal behaviors, and tried to fix them.

compiler/src/toEC.ml
eponier: I removed the axiom. I hope that my changes to EC files make sense.

eponier:
I rebased on latest main.

vbgl:
Arrays of negative size are still scary. Should extraction to EasyCrypt just fail when such a thing happen?

eponier:
For the moment, it seems that some files do not even parse in EC. Let me have a look.

vbgl:
The issue is that the sizes end up in identifiers (e.g., `Array-1`).

eponier:
Thanks. A syntactic fix shouldn't be hard, but the question is: is this the way to go?

vbgl:
> is this the way to go?

Hence my comment above: just refuse to translate programs with arrays of negative size.

eponier:
But the next step is to allow symbolic lengths, that could at runtime be instantiated with negative lengths, but we won't know it statically. What is the move that prepares best for that second change?

vbgl:
Disable the failing test and wait for the future?

bgregoir:
When length will become expression we will have anyway to generate names from expressions. I think it is reasonable to already generate name like Arraym10.

eponier:
> When length will become expression we will have anyway to generate names from expressions. I think it is reasonable to already generate name like Arraym10.

I followed this suggestion.

vbgl:
Some `require import AllCore.` might be missing in the EasyCrypt files generated for each array library with negative sizes.

eponier:
> Some `require import AllCore.` might be missing in the EasyCrypt files generated for each array library with negative sizes.

Thanks for the analysis. I added `require import Int.` to the array files systematically.

vbgl (approved):

PR #1390, opened 2026-02-24
Array lengths are now nonneg integers in N

eponier:
Alternative to https://github.com/jasmin-lang/jasmin/pull/1388, using `N` instead of `Z`.

Arrays of length 0 are now accepted.


bgregoir:
Can we close this PR, since we have decide to use #1388 instead ?

eponier:
Sure!

PR #1424, opened 2026-03-27
Emit vmovdqa instructions on x86

vbgl:
# Description

The compiler emits `VMOVDQA`  instructions instead of `VMOVDQU` when the underlying memory access is known to be aligned.

# Checklist

### DIFF
--- a/proofs/compiler/x86_lowering.v
+++ b/proofs/compiler/x86_lowering.v
@@ -226,6 +226,17 @@ Definition check_signed_range (m: option wsize) sz' (n: Z) : bool :=
       if h <=? z then z <? -h else false)%Z
   else false.
 
+Definition classify_pget ty e x (al: aligned) sz v : lower_cassgn_t :=
+    if (sz ≤ U64)%CMP
+    then LowerMov (if is_var_in_memory v then is_lval_in_memory x else false)
+    else if ty is aword szo
+    then if (U128 ≤ szo)%CMP then
+           let al := min_aligned al (aligned_of_lval x) in
+           LowerCopn (Ox86 (VMOVDQ al szo)) [:: e ]
+    else if (U32 ≤ szo)%CMP then LowerCopn (Ox86 (MOVV szo)) [:: e ]
+    else LowerAssgn
+    else LowerAssgn.
+
 (* x =(ty) e *)
 Definition lower_cassgn_classify ty e x : lower_cassgn_t :=
   let chk (b: bool) r := if b then r else LowerAssgn in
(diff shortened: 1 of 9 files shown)


proofs/compiler/x86_lowering.v
```diff
@@ -226,6 +226,17 @@ Definition check_signed_range (m: option wsize) sz' (n: Z) : bool :=
       if h <=? z then z <? -h else false)%Z
   else false.
 
+Definition classify_pget ty e x (al: aligned) sz v : lower_cassgn_t :=
+    if (sz ≤ U64)%CMP
+    then LowerMov (if is_var_in_memory v then is_lval_in_memory x else false)
+    else if ty is aword szo
+    then if (U128 ≤ szo)%CMP then
+           let al := min_aligned al (aligned_of_lval x) in
```
eponier, line 234: What does it mean to have `al = Aligned` and `aligned_of_lval x = Unaligned`? The user/another pass of the compiler asked for an aligned access, but actually this is not possible? Can it happen in practice?

proofs/lang/fexpr.v
```diff
@@ -33,6 +33,9 @@ Variant lexpr :=
 Notation rexprs := (seq rexpr).
 Notation lexprs := (seq lexpr).
 
+Definition aligned_of_lexpr (x: lexpr) : aligned :=
+  if x is Store al _ _ then al else Aligned.
```
eponier, line 37: Just to be sure, here `Aligned` is a dummy value, or it's what we want for the `LLvar` case?

proofs/compiler/x86_lowering.v
eponier: Oh, I think I understood. The source is known to be aligned, but not the target. But in practice it cannot happen, can it? If we load from memory, the result is a register, so there is no notion of alignment there. And vice versa, if it's a store, we can check the alignment of the target address, but the source is a register. Am I correct?

proofs/lang/fexpr.v
vbgl: This is not dummy. It is the right default for accessing a variable: no alignment constraint.

proofs/compiler/x86_lowering.v
vbgl: We have to take into account the alignment constraints from the source and destination. If both are variables, no problem. If both are memory, asm-gen will fail later (too bad). So the interesting cases are load to register and stores from register, for which one of the alignment is trivially `Aligned` (the register) and shall be ignored by the `min_aligned` (which will return the other).

eponier: Makes sense

proofs/lang/fexpr.v
eponier: Thanks!

PR #1430, opened 2026-04-07
CI: use Coq 9.0

vbgl:
and update the installation guide accordingly.


eponier (dismissed):

eponier:
Should we try to get rid of the warnings?
- `Warning, feedback message received but no listener to handle it!`: we already had this one I think, but it disappeared at some point
- `Warning: Deprecated environment variable COQPATH, use ROCQPATH instead.`: should we use `ROCQPATH`? or just mute it for the time being?
- The debug thing for `compiler/stack_alloc_proof_2.v`: I don't understand where it comes from

eponier:
(and btw, why did the CI run twice on the same commit?)

vbgl:
> (and btw, why did the CI run twice on the same commit?)

Fausse manip’

eponier:
The first two warnings are related it seems: https://github.com/rocq-prover/rocq/issues/20229. This was fixed in Rocq 9.1.

eponier:
If we use `ROCQPATH` instead of `COQPATH` (no idea of what to change and how difficult it is), maybe both are fixed.

vbgl:
I think it would be nice to silence these warnings, I’ve just pushed a patch in this direction. Does it look OK? Does it work in your environment too?

eponier:
Actually, on my machine, I don't see anything in both cases. I think the env var is not used, because Rocq is installed in the default location. Maybe it is used in nix because the location is non-standard?

vbgl:
That’s correct : `nix` “installs” Coq/Rocq libraries by populating the `COQPATH` environment variable. In your case, both are undefined therefore all is fine. Thanks for testing.

proofs/Makefile
```diff
@@ -3,6 +3,11 @@
 # --------------------------------------------------------------------
 SUBDIRS :=
 
+# --------------------------------------------------------------------
+#  Silence some rocq warnings
+ROCQPATH := $(COQPATH)
```
eponier, line 8:
```suggestion
ROCQPATH ?= $(COQPATH)
```
Thinking of it, maybe the `?=` is better, at least more future-proof? If someone on purpose sets `ROCQPATH`, it would be overwritten if we use `:=`.

eponier: (sorry to trigger one more round of review)

vbgl: For the time being, they can be told to set the `COQPATH` variable instead. Soon enough we might be able to properly move to Rocq and get rid of all of this.

eponier:
>     * The debug thing for `compiler/stack_alloc_proof_2.v`: I don't understand where it comes from

As you spotted (cf. https://github.com/jasmin-lang/jasmin/pull/1428), this is not introduced in this PR, so I'll merge without a fix for this one.

proofs/Makefile
eponier: Ok, fair enough!

eponier (approved):

PR #1431, opened 2026-04-08
Adapt to math-comp/math-comp#1456

eponier:
Get rid of ring on the problematic line, and rely only on lia

~Restores the original patch proposed by Pierre Roux, now that we require at least Rocq 9.0~

(cf. https://github.com/jasmin-lang/jasmin/pull/1428)

### DIFF
--- a/proofs/compiler/stack_alloc_proof_2.v
+++ b/proofs/compiler/stack_alloc_proof_2.v
@@ -159,9 +159,8 @@ Proof.
   + by have := size_slot_gt0 x; lia.
   + exists (l0 ++ rdata ++ vdata).
     rewrite heq -!catA; split => //.
-    have ? := size_slot_gt0 x.
-    rewrite -Nat2Z.inj_iff !size_cat !Nat2Z.inj_add Z2Nat.id //; last by lia.
-    by rewrite hsz1 hsz2 heqsz !Z2Nat.id //; first ((rewrite ?(addZE, subZE); ring) || ring); lia.  (* remove the last "|| ring" when requiring MC >= 2.7 *)
+    have slot_x_gt0 := size_slot_gt0 x.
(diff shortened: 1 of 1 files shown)


eponier:
Let's see if CI is happy.

vbgl:
Note that CI does not checks the future changes to math-comp…

eponier:
Yes, but I believe in this case that Pierre's original patch was compatible with the PR he's working on.

eponier:
Anyway, it fails.

eponier:
Let's try this version.

vbgl:
I made a similar try: https://gitlab.com/jasmin-lang/jasmin/-/commit/1bebe0e4ad06cbd6f157d5824a310d5d43b778f5 But it’s hard to know whether it’s OK regarding the future math-comp changes.

eponier:
Yours is cleaner, I'll pick it.

eponier:
Ok, in the end, lia is clever enough to reason about conversions from Z to nat and from nat to Z, so I could make the proof even shorter.

eponier:
> I made a similar try: https://gitlab.com/jasmin-lang/jasmin/-/commit/1bebe0e4ad06cbd6f157d5824a310d5d43b778f5 But it’s hard to know whether it’s OK regarding the future math-comp changes.

We reason on Z, so I don't see how Pierre's PR could have an impact, but to be safe, let's ping @proux01 .

bgregoir (approved):

bgregoir:
Maybe I was to fast on this merge. Sorry.

proux01:
Indeed, apparently it does break on https://github.com/math-comp/math-comp/pull/1456/ (c.f., https://github.com/math-comp/math-comp/actions/runs/24086349000/job/70736502192?pr=1456 )

PR #1432, opened 2026-04-10
Adapt to https://github.com/math-comp/math-comp/pull/1456

proux01:
Adapt to https://github.com/math-comp/math-comp/pull/1456/ (again)

The previous https://github.com/jasmin-lang/jasmin/pull/1431 was implicitly relying on mczify. If we remove that dependency, we need one more step for Stdlib's lia to work.


eponier:
Oh, indeed, the `+` hid `addn`, not `Z.add`.

eponier (approved):
Hopefully this is the last round on that! (sorry for the multiple round trips on this one)

proux01:
Thanks, and sorry that it took so many iterations

PR #1438, opened 2026-04-16
Warn about assumed alignment of arguments to export functions

vbgl:
The compiler infers the required alignment for function arguments. For export functions, the inferred alignment should not excess the declared type of said argument. When the inferred alignment excesses the bound, a warning is emitted.

The new command-line flag `-only-trivial-alignment` restrict the bound to `U8` (i.e., warns more often).

# Checklist

### DIFF
--- a/changes/01-feature/1438-weaker-align-assumptions.md
+++ b/changes/01-feature/1438-weaker-align-assumptions.md
@@ -1,0 +1,6 @@
+- The compiler now warns when array arguments to export functions are assumed to
+  have some unexpected alignment. Such arrays are expected to be aligned to the
+  declared size of their elements, unless the `-only-trivial-alignment`
+  command-line flag is used, in which case they are not expected to have any
+  specific alignment.
+  ([PR 1438](https://github.com/jasmin-lang/jasmin/pull/1438)).
--- a/compiler/src/glob_options.ml
+++ b/compiler/src/glob_options.ml
@@ -29,6 +29,7 @@ let introduce_array_copy = ref true
 let introduce_export_renaming = ref true
 let print_dependencies = ref false
 let lazy_regalloc = ref false
+let only_trivial_alignment = ref false
 
 let verbosity = ref 1
 
@@ -48,6 +49,7 @@ let set_linting_level i =
 
 let enable_all_warnings () =
   linting_level := 2;
+  only_trivial_alignment := true;
   set_all_warnings ()
 
 let do_auto_spill = ref None
@@ -223,6 +225,7 @@ let options = [
     "-wall", Arg.Unit enable_all_warnings, " Enable all warnings";
     "-nowarning", Arg.Unit (nowarning), " Do no print warnings";
     "-linting-level", Arg.Int set_linting_level, "[n] Set linting level to n (defaults to 1; disable linting when set to 0)";
+    "-only-trivial-alignment", Arg.Set only_trivial_alignment, " Warn if export functions assume some alignment for their array arguments";
     "-color", Arg.Symbol (["auto"; "always"; "never"], set_color), " Print messages with color";
     "-help-intrinsics", Arg.Set help_intrinsics, " List the set of intrinsic operators (and exit)";
     "-auto-spill", Arg.Unit set_auto_spill, " Enable naive spilling of #[spill]-annotated variables";
--- a/compiler/src/varalloc.ml
+++ b/compiler/src/varalloc.ml
@@ -274,7 +274,7 @@ let init_slots pd stack_pointers alias coloring fv =
   !slots, lalloc
 
 (* --------------------------------------------------- *)
-let all_alignment pd (ctbl: alignment) alias params lalloc : param_info option list * wsize Hv.t =
+let all_alignment ~is_export pd (ctbl: alignment) alias params lalloc : param_info option list * wsize Hv.t =
 
   let get_align c =
     try Hv.find ctbl c.Alias.in_var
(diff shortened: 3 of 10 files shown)


eponier:
Hi Vincent, is it related to https://github.com/jasmin-lang/jasmin/pull/1424?

vbgl:
Yes, sort of.

changes/01-feature/1438-weaker-align-assumptions.md
```diff
@@ -0,0 +1,6 @@
+- The compiler now warns when array arguments to export functions are assumed to
+  have some unexpected alignment. Such arrays are expected to be aligned to the
+  declared size of their elements, only the `-only-trivial-alignment`
```
eponier:
```suggestion
  declared size of their elements, unless the `-only-trivial-alignment`
```
?

compiler/src/glob_options.ml
```diff
@@ -48,6 +49,7 @@ let set_linting_level i =
 
 let enable_all_warnings () =
   linting_level := 2;
+  only_trivial_alignment := true;
```
eponier, line 52: `only_trivial_alignment` is already set by `set_all_warnings ()` if I understand correctly

compiler/src/varalloc.ml
```diff
@@ -291,6 +291,19 @@ let all_alignment pd (ctbl: alignment) alias params lalloc : param_info option l
         | _ | exception Not_found -> assert false in
       let pi_writable = w = Writable in
       let pi_align = get_align c in
+      (* Check that the assumed alignment for array arguments to export functions does not excess some bound.
```
eponier:
```suggestion
      (* Check that the assumed alignment for array arguments to export functions does not exceed some bound.
```

eponier (commented):
The warning message would be clearer with something like `but is declared with type ...`. If `only_trivial_alignment` is set, the message has to be tweaked, though.

compiler/src/glob_options.ml
vbgl: No, it is not.

vbgl:
I’ve fixed the typos. Thanks for spotting them.

I’m not sure how to improve the warning message, though.

compiler/src/glob_options.ml
eponier: Oh, I got confused between the option to default to `U8` and the printing of the warning, my bad.

eponier:
I've just pushed some suggestion.

eponier:
(I'm unsure where to put "only" in the sentence)

vbgl:
Your use of “assumed” is wrong: the compiler assumes a larger alignment that the user might expect, hence the warning.

eponier:
`but it is inferred to have only %a-bit alignment`?

vbgl:
I’ve fixed the PR title (and message commit accordingly).

eponier:
Your proposal for the warning message was ok, I think. Did you remove it on purpose?

eponier:
(if you push again, you can also excess -> exceed in the commit message itself)

bgregoir:
Hi, 
  I will this PR but I have some remark on the default behaviour.
  First I think that the default should be -only-trivial-alignment.
  The reason is that we can not ensure that external caller will provide the desired alignment. 
  Second I think it is preferable to have an explicite annotation for providing a different alignment, 
  instead of using the type information. 
  The main motivation is for proper documentation of the code, and also to disconnect the type information
  from the alignement (maybe the type information is set to this value for proving the program).

  For that we can put the annotation on the parameters declaration of the functions:
      export fn foo (#[align=u64] reg ptr u32[n] t1, #[align=u32] reg ptr u16[k] t2, #[align=u32] reg ptr u16[k] t2) ....
  or in the annotation of the function:
      #[align_arg = { t1 = U64; t2 = U32; t3 = U32 }]
  Or any other syntax ...
  If the alignment is not specified the u8 will be used. 
  

  Maybe it is problematic to do this change in one go for compatibility reason, so I did not say that we should 
  do this directly.

vbgl:
> First I think that the default should be -only-trivial-alignment.

86 cases in the compiler test suite would rise a warning.

That being said, I agree that any serious development should only have byte arrays as arguments.

Note that for `ptr` arrays, the type annotation is mostly irrelevant; it only tells what the default access size is. This PR treats it as a hint for the documented expected alignment.

eponier:
> Your proposal for the warning message was ok, I think. Did you remove it on purpose?

?

vbgl:
Yes.

eponier:
> Yes.

But why? It was much clearer than the current message I think.

vbgl:
> It was much clearer than the current message I think.

I disagree. I think the current version is the best attempt so far: short and to the point.

The purpose of this warning is to let the user know that something is fishy in the required alignment. As Benjamin rightly noted, code that is meant to be executed should not have any non-trivial alignment constraints, so the warning just warns about that: “hej, your code has some non-trivial alignment requirement (of xx-bit, for pointer p)”.

The details on how the compiler decides to show the warning are (to me) just internal cuisine that should not leak to the output message.

eponier:
I'm ok not being as specific as your temporary trial, but currently for me the message is not helpful for users. It could either give pointers to the command-line option or `#unaligned`, or at least mention that the alignment is (or might be) too high. Your proposal
```quote
“hej, your code has some non-trivial alignment requirement (of xx-bit, for pointer p)”.
```
works too.

eponier:
If I draw a (probably bad) parallel with typing, on this stupid example
```
export fn main (reg u32 x) -> reg ptr u64[2] {
  return x;
}
```
In the spirit of the current error message in this PR, we would get
```
typing error: the expression has type u32
```
missing the crucial part that allows the user to really understand the problem and fix it
```
typing error: the expression has type u32 instead of u64[2]
```
A useful message could also be that it does not match the provided return type.

bgregoir:
I send a message to try to reactivate this PR. 
I don't have a strong opinion on the error message, but I think we can have something short and meaning full (so both of you will be happy). 
On my side I don't like the fact to use the type as an hint or an annotation for the alignment. 
Mostly because the type is already used as an annotation for the extraction to indicate how to extract array (if we do not use barray). And it seems that people (mostly Manuel) prefers this version. 
In any case, I think this change can be done latter.

vbgl:
Let’s focus on #1478 first and consider again rejecting programs with non-trivial alignment requirements.

vbgl:
>  On my side I don't like the fact to use the type as an hint or an annotation for the alignment.

This is not what this PR does. In this PR, the type is used as an heuristic to decide whether a warning is displayed or not.

vbgl:
Let me try again in a different PR. Thanks for all the feed-back.

PR #1439, opened 2026-04-17
Regalloc hints

sarranz:
# Description

Experimental hints for the register allocator requested by @bgregoir 

# Checklist


sarranz:
I removed the "conflict" annotation. One thing worth noting:
```
export fn main() -> reg u64 {
    #[force_regalloc=RCX] reg u64 x = 0;
    return x;
}
```
fails even if the compiler introduces renaming, because the renames inherit the annotations. I don't know if it's a feature or a bug.

PR #1445, opened 2026-04-29
Implement error resilience in pretyping by accumulating typing errors

clebreto:


vbgl:
Did you consider using effects? They look like the right tool for this task.

clebreto:
> Did you consider using effects? They look like the right tool for this task.

Jean-Christophe clearly mentioned it, but as a feature of Ocaml 5 ? @eponier ?

eponier:
The changes we had to do were painful, and yes it looks like something algebraic effects are good for. I see two limits though:
- in the way we implemented this, we don't catch every error; basically we report at most one error per instruction (the idea was to avoid reporting too many errors, I don't know if this is really a good idea)
- most of the time, when there is an error, there is a real question of what to return. For now, we drop most of the wrong things. After discussing with Enrico, it would make better sense to add an error node instead.

Basically, there is some design choice in the details, and I don't know whether algebraic effects would allow to do the specific things we may want to do in some cases.

eponier:
After discussing with Benjamin, the plan is to always generate (as much as possible) an AST, even if some terms are ill-typed. If we manage to do that, then it may possible to implement a switch choosing between failing at the first error and accumulating all the errors, using algebraic effects.

PR #1447, opened 2026-05-05
CI: update nixpkgs (elpi 3.7.1 for coq-master)

vbgl:
The `coq-master` has been failing. This fixes it by updating `elpi` to version 3.7.1 (which is only available in recent `nixpkgs`).

PR #1452, opened 2026-05-13
CI: fix coq-master job

vbgl:


eponier (approved):
I thought I had already merged this one...

vbgl:
Do you imply that this `coq-master` job breaks quite often? Cf. #1447.

eponier:
When I saw this PR, my reaction was indeed "this job breaks often". But I am not sure that this breaks that often. And additionally, I don't understand why the breakage was not reported by upstream (now that Jasmin is back in Rocq's CI).

As for my comment, it was much simpler than that. I had already had a look at this PR, and literally thought I had already accepted it.

PR #1453, opened 2026-05-18
CI: compile and check for CT formosa/mldsa

vbgl:
This adds a CI job that compiles the current development version of formosa-mldsa and runs the CT-checker.

### DIFF
--- a/scripts/test-mldsa.sh
+++ b/scripts/test-mldsa.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+set -e
+
+REPO=formosa-crypto
+NAME=formosa-mldsa
+BRANCH=main
+
+FILE="$NAME.tar.gz"
+ROOT=$(echo -n $NAME-$BRANCH | tr / -)
+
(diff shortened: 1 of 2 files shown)


scripts/test-mldsa.sh
```diff
@@ -0,0 +1,39 @@
...
+REPO=formosa-crypto
+NAME=formosa-mldsa
+BRANCH=main
+
+FILE="$NAME.tar.gz"
+ROOT=$(echo -n $NAME-$BRANCH | tr / -)
+
+if [ -z "$JOBS" ]
+then
+   JOBS=1
```
eponier: `JOBS` is not used, isn't it?

vbgl: Indeed. Fixed.

eponier (dismissed):

vbgl:
This was missing the SCT check for the x86/ref implementations. Sorry about that.

eponier (approved):

PR #1458, opened 2026-05-26
Update the x86 DOIT instruction list

vbgl:
# Description

Update the list of DOIT instructions for x86 according to #1422 and change the insecure example from the documentation (as rotations are now DOIT).

Fixes #1422.

# Checklist

### DIFF
--- a/changes/01-feature/1458-doit-update.md
+++ b/changes/01-feature/1458-doit-update.md
@@ -1,5 +1,5 @@
 - More instructions are considered “DOIT” on `x86_64`: byte swap (`BSWAP`),
-  sign-extend (`CDQ`, `CQO`), rotations (`ROL`, `RORX`, `ROR`), shifts (`SHLD`,
+  sign-extend (`CQO`), rotations (`ROL`, `RORX`, `ROR`), shifts (`SHLD`,
   `SHRD`)
   ([PR 1458](https://github.com/jasmin-lang/jasmin/pull/1458);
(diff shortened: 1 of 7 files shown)


changes/01-feature/1458-doit-update.md
```diff
@@ -0,0 +1,5 @@
+- More instructions are considered “DOIT” on `x86_64`: byte swap (`BSWAP`),
+  sign-extend (`CDQ`, `CQO`), rotations (`ROL`, `RORX`, `ROR`), shifts (`SHLD`,
```
bgregoir: It seems that only CQO is and not CDQ, did I miss-understand ?
Anyway the commit only add CQO to the list.

vbgl: The commit does add both `CDQ` and `CQO` to the list. And also `CWD`. These are three ways of spelling “sign-extend” in the Intel™ parlance.

Do you think only the 64-bit case is doit?

bgregoir: I am not sure I look to the good place, but I did not find CDQ and CWD to the list. 
https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/data-operand-independent-timing-instructions.html

vbgl: That webpage really is not precise. Let’s be safe and remove CDQ and CWD from the doit list.

vbgl: (done in a fresh commit)

bgregoir (approved):

PR #1461, opened 2026-05-27
CI: update nixpkgs and fix coq-master job

vbgl:
Follows the merge of https://github.com/math-comp/math-comp/pull/1456


eponier:
@vbgl `ring` was barely used in that file, so I saw the opportunity to make a more future-proof change. I force-pushed on your PR, I hope you don't mind. Don't hesitate to rollback if you don't like my changes.

vbgl:
No problem. Thanks for your patch. I think we can get even further and drop a dependency that is no longer needed.

eponier (approved):
Cool!

PR #1464, opened 2026-05-27
Accept Lnone as destination for #randombytes

### LINKED ISSUE #1462: Unexpected randombytes lval typing error
This
```
export fn main() {
  stack u8[1] s;
  _ = #randombytes(s);
}
```
gives
```
line 3 (2-3):
typing error: _ lvalue not accepted here
```
which doesn't look like a typing error and maybe should be accepted?

vbgl:
# Description

Although it is arguably OK to reject programs that simply discard the result of a call to `#randombytes`, claiming that this is a “typing error” is wrong.

Fixes #1462

# Checklist


eponier (approved):
Accepting `Lnone` in stack alloc would, I think, be cleaner, but this is more work.

PR #1465, opened 2026-05-27
Safety: do not use the source type as an assumption

vbgl:
# Description

This essentially removes dead code.

# Checklist


eponier (approved):
Less code 🎉

PR #1473, opened 2026-06-08
Enforce output constraints in regalloc

sarranz:
# Description

Register allocation should look into argument constraints of operators for both inputs and outputs. The new code has one difference: we used to `assert false` on implicit arguments that are not `Lvar`s, now we ignore them.

### DIFF
--- a/compiler/src/regalloc.ml
+++ b/compiler/src/regalloc.ml
@@ -717,37 +717,32 @@ module Regalloc (Arch : Arch_full.Arch)
       in
       allocate_one nv vars loc cnf x i y a
     in
-    let mallocate_one x y a =
-      match x with Pvar x when is_gkvar x -> allocate_one x.gv y a | _ -> ()
+    let var_of_expr e =
+      match e with Pvar x when is_gkvar x -> Some x.gv | _ -> None
+    in
+    let var_of_lval lv =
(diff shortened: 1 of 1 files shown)


compiler/src/regalloc.ml
```diff
@@ -717,37 +717,32 @@ module Regalloc (Arch : Arch_full.Arch)
...
+      let xarg = get_var arg in
+      let add = conflicts_add_one Arch.pointer_data Arch.reg_size Arch.asmOp in
+      match ad with
+      | ADImplicit v
+      | ADExplicit (_, ACR_exact v) ->
+        Option.may (fun x -> allocate_one x (Conv.var_of_cvar v) a) xarg;
+        cnf
+      | ADExplicit (_, ACR_any) -> cnf
+      | ADExplicit (_, ACR_subset rs) ->
+        match xarg with
```
bgregoir, line 736: Why do you use a option.may line 732 and a match here ?

sarranz: Because it returns different `cnf`s and the body is two lines; it would be an `omap` and then an `odflt` and I thought it looked less clear. I have no particular preference, I can change it

bgregoir (approved):

PR #1474, opened 2026-06-08
Various fixes to the printer 

eponier:
# Description

In the test suite, we test the printer just after pretyping (for `pprog`s), but not the printer used after param expansion (for `prog`s). It turns out that it was bugged.

Started from a bug report by @FantomeBeignet, with help from @bgregoir 

# Checklist

### DIFF
--- a/compiler/src/conv.ml
+++ b/compiler/src/conv.ml
@@ -406,6 +406,13 @@ let fresh_var_ident =
     | x -> x
     | exception Not_found ->
         let ty = ty_of_cty st in
-        let x = V.mk n r ty i_loc.L.base_loc [] in
+        let v_annot =
+          (* When generating an mmx variable, we add the #[mmx] annotation
+             for the sake of clarity when printing programs *)
+          match r with
+          | Wsize.Reg (Extra, _) -> [(L.mk_loc L._dummy "mmx", None)]
+          | _ -> []
+        in
+        let x = V.mk n r ty i_loc.L.base_loc v_annot in
         Hashtbl.add memo k x;
         x
--- a/compiler/src/printer.ml
+++ b/compiler/src/printer.ml
@@ -293,6 +293,15 @@ let pp_contract ~debug pp_len pp_var fmt fc =
 let pp_ocontract ~debug pp_len pp_var fmt fc =
   Option.may (pp_contract ~debug pp_len pp_var fmt) fc
 
+let pp_header_ pp_size pp_var fmt fd =
+  let pp_vd =  pp_var_decl pp_var pp_size in
+  let ret = List.map L.unloc fd.f_ret in
+  let set_var_type x ty = GV.mk x.v_name x.v_kind ty x.v_dloc x.v_annot in
+  F.fprintf fmt "fn %s @[(%a)@] -> @[(%a)@]"
+    fd.f_name.fn_name
+    (pp_list ",@ " pp_vd) fd.f_args
+    (pp_return_type pp_size) (List.combine fd.f_ret_info.ret_annot (List.map2 set_var_type ret fd.f_tyout))
+
 let pp_gfun ~debug (pp_size:F.formatter -> 'size -> unit) pp_opn pp_var fmt fd =
   let ds = ScopeTree.get_declaration_sites fd in
   let pp_vd =  pp_var_decl pp_var pp_size in
@@ -301,7 +310,6 @@ let pp_gfun ~debug (pp_size:F.formatter -> 'size -> unit) pp_opn pp_var fmt fd =
     |> Spv.iter (F.fprintf fmt "%a;@ " pp_vd)
   in
   let ret = List.map L.unloc fd.f_ret in
-  let set_var_type x ty = GV.mk x.v_name x.v_kind ty x.v_dloc x.v_annot in
   let pp_ret fmt () =
     (* In the rare case where there is no instruction in the function,
        the results that are not arguments are not get properly declared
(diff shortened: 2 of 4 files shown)


compiler/src/printer.ml
```diff
@@ -512,7 +512,7 @@ let pp_sprog ~debug pd msfsize asmOp fmt ((funcs, p_extra):('info, 'asm) Prog.sp
   let pp_opn = pp_opn pd msfsize asmOp in
   let pp_var = pp_var ~debug in
   let pp_f_extra fmt f_extra =
-    Format.fprintf fmt "(* @[<v>alignment = %s; argument alignment = [%a];@ stack size = %a + %a; max stack size = %a;@ max call depth = %a;@ saved register = @[%a@];@ saved stack = %a;@ return_addr = %a@] *)"
+    Format.fprintf fmt "/* @[<v>alignment = %s; argument alignment = [%a];@ stack size = %a + %a; max stack size = %a;@ max call depth = %a;@ saved register = @[%a@];@ saved stack = %a;@ return_addr = %a@] */"
```
bgregoir, line 515: At least some of this information can be expressed has valid annotations (like stack size --> stacksize)

eponier:
Actually, since we keep the same representation until stack alloc, I tried the following experiment: for every pass, compile until that pass (program `p1`), print the result, reparse it and process it until after param expansion (program `p2`), and compare `p1` and `p2`. It worked rather well. Here are the reasons when it failed (meaning `p1` and `p2` are not equal, or the compiler fails on `p2`):
- parse error:
  - instructions that can be printed but not parsed (`NOP`, `MOVZX32`, x86 SLH-operators)
- type error:
  - reject some `SWAP` instruction where all operands and results do not have the same type (is pretyping too strict?)
  - ARM instructions with shifts (printed as an additional argument, but parsed with `<<` syntax)  
  - no way to distinghish between the versions of `#protect` that apply to standard registers or MMX registers
  - `_ is mutable, it should be returned` after liverange splitting
- not the same program:
  - instructions that are printed the same (pseudo-op `SWAP` vs arm-m4 and risc-v `SWAP`, x86 `protect` was also printed the same as SLH `protect`, I added a prefix to distinguish them)
  - negative integers are reparsed as the operator `-` applied to a positive integer
  - tags of instructions are printed as comments, they cannot be reparsed as is
  - types of assignments were compatible but not equal (`u16[2]` vs `u8[4]` for instance)
  - MMX variables introduced by the compiler were not printed with the corresponding `#[mmx]` annotation, I fixed that
  - `#copy` are added to the reparsed program by `insert_copy_and_fix_length`

Not sure that I am exhaustive, but this gives the idea.

eponier:
And by the way, some commits are more debatable than others. I can drop the controversial ones if needed.

bgregoir:
For the swap, there is two difficulties:
- The arguments do not share the same type, this should be accepted by pretyping (if it they are bigger than the swap type)
- The destinations do not share the same type, if one is strictly bigger that the swap type if should be rejected. But remark that this is only true because pretyping is for the initial semantics without subword. So the compiler can generate this cituation.

For the tag, maybe the solution will be to print them as annotation. I think some annotation are already done for that like #[keep].

compiler/src/conv.ml
```diff
@@ -406,6 +406,13 @@ let fresh_var_ident =
     | x -> x
     | exception Not_found ->
         let ty = ty_of_cty st in
-        let x = V.mk n r ty i_loc.L.base_loc [] in
+        let v_annot =
+          (* When generating an mmx variable, we add the #[mmx] annotation
+             for the sake of clarity when printing programs *)
```
vbgl, line 411: This does not look like the correct place to fix a printing issue. When printing a declaration of an extra variable, the printer could add the `#[mmx]` annotation (unless it is already there).

vbgl:
> Started from a bug report

If this fixes some bug, a non-regression test case would be welcome.

compiler/src/printer.ml
eponier: Compiler info and annotations feel different to me, so I find it normal that they are printed differently.

eponier:
> For the swap, there is two difficulties:

Regarding `#swap`, I think it works more or less like that since Vincent fixed the pretyping (see https://github.com/jasmin-lang/jasmin/pull/1285).

And I misread what I did. The problem arises after lowering, when `SWAP` has been turned into `XCHG`. Maybe we are just too strict about intrinsics in general? For instance
```
export fn main () {
  reg u32 x;
  reg u64 y;
  ?{}, x = #ADD_64(y, y);
}
```
is rejected with `typing error: the expression has type u32 instead of u64`. Maybe we should tolerate it, since `u32` is shorter than `u64`?

compiler/src/conv.ml
eponier: Initially this is what I wanted to do. Then I thought we could have as an invariant that every variable of kind `Extra` has the annotation `#[mmx]`. But I can make a fix limited to the printer.

eponier:
> > Started from a bug report
> 
> If this fixes some bug, a non-regression test case would be welcome.

The bug was that the function annotations were not printed, it's difficult to add a test case without writing the same kind of code as `tests/printing.ml`.

compiler/src/conv.ml
vbgl: Reading this again, I think the proposed patch makes sense. Thanks.

vbgl (approved):
Should I squash all these commits or preserve the history?

eponier:
You can squash

eponier:
Thanks!

PR #1475, opened 2026-06-09
Add overlay

bgregoir:
# Description
The first commits fix some error in the code of alias.
The seconds add the overlay annotation allowing the user to force the compiler to merge stack arrays (possibly of different size).

This is done in collaboration with Jean-Christophe Léchenet.
Fixes # (issue) 

# Checklist

### DIFF
--- a/compiler/src/alias.ml
+++ b/compiler/src/alias.ml
@@ -130,19 +130,19 @@ let compare_gvar params x gx y gy =
     let sx = size_of x.v_ty in
     let sy = size_of y.v_ty in
     match gx, gy with
-    | E.Sglob, E.Sglob -> 
+    | E.Sglob, E.Sglob ->
       hierror_no_loc "cannot merge two globals (%a and %a)" pp_var x pp_var y
-    | E.Sglob, E.Slocal -> 
+    | E.Sglob, E.Slocal ->
       if (Sv.mem y params) then hierror_no_loc "cannot merge a global and a param (%a and %a)" pp_var x pp_var y;
       check_size "global" y sy x sx;
       1
-    | E.Slocal, E.Sglob -> 
+    | E.Slocal, E.Sglob ->
       if (Sv.mem x params) then hierror_no_loc "cannot merge a param and a global (%a and %a)" pp_var x pp_var y;
       check_size "global" x sx y sy;
       -1
     | E.Slocal, E.Slocal ->
       match Sv.mem x params, Sv.mem y params with
-      | true, true -> 
+      | true, true ->
         hierror_no_loc "cannot merge two params (%a and %a)" pp_var x pp_var y;
       | true, false -> check_size "param" y sy x sx; 1
       | false, true -> check_size "param" x sx y sy; -1
@@ -234,11 +234,11 @@ let link_array_return params a xs es cc =
         )
         a xs cc
 
-let opn_cc o = 
+let opn_cc o =
   match o with
   | Sopn.Oslh (SLHprotect_ptr_fail _) -> Some [Some 0]
   | Sopn.Opseudo_op(Pseudo_operator.Oswap _) -> Some [Some 1; Some 0]
-  | _ -> None 
+  | _ -> None
 
 let rec analyze_instr_r params cc a =
   function
@@ -246,10 +246,10 @@ let rec analyze_instr_r params cc a =
   | Ccall (xs, fn, es) -> link_array_return params a xs es (cc fn)
   | Csyscall (xs, o, es) -> link_array_return params a xs es (syscall_cc o)
   | Cassgn (x, _, ty, e) -> if is_ty_arr ty then assign_arr params a x e else a
-  | Copn (xs, _, o, es) -> 
+  | Copn (xs, _, o, es) ->
     (* A special case for operators that can return array *)
-    begin match opn_cc o with 
-    | None -> a 
+    begin match opn_cc o with
+    | None -> a
     | Some l -> link_array_return params a xs es l
     end
   | Cassert _ -> a
--- a/compiler/src/alias.ml
+++ b/compiler/src/alias.ml
@@ -8,16 +8,23 @@ let hierror = hierror ~kind:"compilation error" ~sub_kind:"stack allocation"
    by catching the exception and reemiting it with more information. *)
 let hierror_no_loc ?funname = hierror ~loc:Lnone ?funname
 
+type range = int * int
 type sub_slice_kind =
-  | Exact
+  | Exact of range
     (* the range is exact *)
   | Sub of Wsize.wsize
     (* the precise offset is not known, we remember that it is a subpart
        and its alignment *)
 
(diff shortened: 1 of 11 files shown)


eponier:
It seems you forgot to call `dune promote` in the latest commit.

compiler/src/varalloc.ml
```diff
@@ -239,15 +239,23 @@ let init_slots pd stack_pointers alias coloring fv =
...
         let c = Alias.normalize_var alias v in
+        let range =
+          match c.kind with
+          | Exact range -> range
+          | Sub _ ->
+              hierror ~loc:(Lone v.v_dloc) "cannot allocate in the stack the variable “%a” to “%a” with non constant start index"
+                    (Printer.pp_var ~debug:false) v
+                    (Printer.pp_var ~debug:false) c.in_var in
         if c.scope = E.Sglob then
           (* TODO: do we need to check that we are exact and fail otherwise? *)
```
eponier: The comment should be removed (we do now check that we are `Exact`)

compiler/src/varalloc.ml
```diff
@@ -239,15 +239,23 @@ let init_slots pd stack_pointers alias coloring fv =
...
-          add_local v (Direct (c.in_var, r2i c.range, E.Sglob))
+
+          add_local v (Direct (c.in_var, r2i range, E.Sglob))
         else
           begin
             let slot = get_slot coloring c.in_var in
             add_slot slot;
             (* TODO: do we need to check that we are exact and fail otherwise? *)
```
eponier: idem

compiler/src/varalloc.ml
bgregoir: done

compiler/src/varalloc.ml
bgregoir: done

compiler/src/alias.ml
```diff
@@ -130,51 +136,81 @@ let compare_gvar params x gx y gy =
...
 (* Precondition: s1 and s2 are normal forms (aka roots) in a *)
+(* We want to unify the slice of s1 and the slice of s2,
+   Assume that [size_of s1.in_var <= size_of s2.in_var]
+   We need to find express the position of the slice
+    [{in_var = s1.in_var; kind : Exact (0, size_of s1.in_var)}]
+    in term of [s2.in_var], in such way that s1 and s2 are aligned
+
+  s1.in_var   =               [......   | s1 | ...]
+  s2.in_var   =   [...........|...| s2 | ... | ......]
+  s1.in_var ->  [...|            | ......]
```
eponier: Why two lines with `s1.in_var`?

compiler/src/alias.ml
```diff
@@ -130,51 +136,81 @@ let compare_gvar params x gx y gy =
...
+
+
+let mk_sub sz1 sz2 ws =
+  if sz1 = sz2 then Exact(0,sz2)
+  else Sub ws
 
 (* Precondition: s1 and s2 are normal forms (aka roots) in a *)
+(* We want to unify the slice of s1 and the slice of s2,
+   Assume that [size_of s1.in_var <= size_of s2.in_var]
+   We need to find express the position of the slice
```
eponier:
```suggestion
   We need to express the position of the slice
```

compiler/src/alias.ml
```diff
@@ -130,51 +136,81 @@ let compare_gvar params x gx y gy =
...
+    [{in_var = s1.in_var; kind : Exact (0, size_of s1.in_var)}]
+    in term of [s2.in_var], in such way that s1 and s2 are aligned
```
eponier: I don't like `aligned` here, it could create a confusion with alignment constraints

compiler/src/alias.ml
```diff
@@ -269,14 +304,51 @@ let rec analyze_instr_r params cc a =
        else loop (merge params a' a |> normalize_map )
      in loop (analyze_stmt params cc a s1 |> normalize_map)
 and analyze_instr params cc a { i_loc ; i_desc } =
+ (* Format.eprintf "analyze_instr %a: %a@."
+    Location.pp_iloc i_loc pp_alias a;*)
```
eponier: Leftover debug, to be removed

compiler/src/alias.ml
```diff
@@ -269,14 +304,51 @@ let rec analyze_instr_r params cc a =
...
 and analyze_stmt params cc a s =
   List.fold_left (analyze_instr params cc) a s
 
+let is_stack_direct k = k = Stack Direct
+
+let has_overlay annot =
+  Annot.ensure_uniq1
+    "overlay"
+    (Annot.on_attribute ~on_id:(fun _ _ s -> s) ~on_string:(fun _ _ s -> s)
+       (fun loc id -> Annot. error ~loc "attribute for “%s” should be a string" id))
```
eponier, line 317:
```suggestion
       (fun loc id -> Annot.error ~loc "attribute for “%s” should be a string" id))
```

compiler/src/alias.ml
bgregoir: overlay ?

eponier: Actually, I'm not even sure of what you want to express here

bgregoir: I think that s1 and s2 will perfectly overlap.

compiler/src/alias.ml
bgregoir: done

compiler/src/alias.ml
bgregoir: pushed some changes

compiler/src/alias.ml
bgregoir: done

eponier:
I commented the form, but it'd be better if @vbgl did the real review, since I co-authored part of the PR.

vbgl:
Is this a regression?
~~~
export fn main(reg ptr u8[32] m) -> reg u32 {
  reg u32 x = m[:u32 0];
  reg ptr u32[1] slice;
  if x != 0 {
    slice = m[4:4];
  } else {
    slice = m[x:4];
  }
  x = slice[#unaligned 0];
  return x;
}
~~~

Prior to this change, we get an informative error message:

> stack allocation: cannot merge distinct slices of the same array: m[4; 8[ and m[0; 4[

After this change, we get a cryptic one:

> stack allocation: no region associated to variable slice

eponier:
Actually, this is a fix! Only half joking, since `m[0; 4[` was clearly wrong.

eponier:
The idea is that we emit the error `cannot merge distinct slices of the same array` only when we are sure that the slices are distinct, i.e. when the two slices to merge are in the `Exact` case. In your example, we are not in the `Exact` case in the `else` branch (we were before the PR, but this was a bug), so we don't raise an error. The OCaml code succeeds, but then the Rocq code fails. It's not clear to me how to make better in this case.

vbgl (approved):
The algorithm for discovering what are the overlays does not seem very efficient (optimization left as future work).

PR #1478, opened 2026-06-10
Array #copy does not assume alignment

vbgl:
# Description

Currently, the `#copy` operator assumes that its source and destination are aligned and its translation emits “aligned” memory accesses. This imposes some constraints on the arrays that are involved in `#copy` operation but programmers have no control over it and these constraints are often spurious (emitted code would be the same without the alignment assumption).

# Checklist


bgregoir (approved):
I think that it will be nice to allow the user to specify if the #copy as to be align. 
But I am not sure what is be best way to provide this information. 
``` 
  #[align] t = #copy(s)
```
or we introduce a #copy_aligned operator. I don't know.

vbgl:
Usually, the assumptions about source and destination are different. Typical example: copy from memory to stack, source is not aligned but destination is.

PR #1480, opened 2026-06-12
Catch annotation errors

eponier:
# Description

Annotation errors were not caught in `main_compiler.ml`. I also added a test case, even though I realized later that for such test cases, we do not call `jasminc` but Jasmin as a library (and thus it does not exercise the change I did). It's good to have anyway.

Does it need a changelog entry?

The first commit is unrelated.

# Checklist

### DIFF
--- a/compiler/tests/negative.expected
+++ b/compiler/tests/negative.expected
@@ -4,6 +4,10 @@ fail/annotation/x86-64/calldepth.jazz:
 compilation error in function f:
 stack allocation: the maximum call depth is 1 (expected: 2)
 
+fail/annotation/x86-64/returnaddress.jazz:
+
+"fail/annotation/x86-64/returnaddress.jazz", line 3 (2-15): attribute for “returnaddress” should be one of “stack”, “reg”
+
 fail/annotation/x86-64/stackalign.jazz:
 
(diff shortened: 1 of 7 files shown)


compiler/tests/negative.expected
```diff
@@ -4,6 +4,11 @@ fail/annotation/x86-64/calldepth.jazz:
 compilation error in function f:
 stack allocation: the maximum call depth is 1 (expected: 2)
 
+fail/annotation/x86-64/returnaddress.jazz:
+
+"fail/annotation/x86-64/returnaddress.jazz", line 3 (2-15): attribute for “returnaddress” should be (
+stack | reg)
```
vbgl: You may want to fix the formatting here (looks like removing boxes should help).

vbgl: Also remove parentheses and esoteric symbols.

eponier: The thing is that we don't print the same in the compiler and in the tests. In the compiler, we wrap everything in a box and there is no problem. In the tests, we just call `printf` and we get these strange lines.

But anyway, it was simpler to remove the box, so I followed this path. I rephrased the message as well.

vbgl (approved):
Thanks.

PR #1482, opened 2026-06-16
Do not assume non-trivial alignment of array arguments to export functions

vbgl:
# Description

Array arguments to export functions are turned into memory addresses after compilation. They usually cannot be assumed to have any specific alignment. Such assumptions can be expressed explicitly through a `required_alignment` annotation.

# Checklist

# TODO

- [x] Discuss libjade changes
- [x] Write documentation

### DIFF
--- a/changes/01-feature/1482-enforce-export-unaligned.md
+++ b/changes/01-feature/1482-enforce-export-unaligned.md
@@ -1,0 +1,4 @@
+- The compiler no longer assumes non-trivial alignment of array arguments to
+  export functions; alignment requirements can be expressed through the
+  `required_alignment` annotation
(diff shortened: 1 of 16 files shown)


bgregoir (commented):
Thank for this work. 
I think we should add some documentation (but maybe first merge the other PR).
It will be nice to extend the ability of declaring an alignment to stack variables (not necessary in this PR).

bgregoir:
This PR can be closed?

vbgl:
Let me write some doc and it will be ready. Due to #1478 and #1485, no change to existing code is necessary.

vbgl:
Done.

bgregoir (approved):

PR #1483, opened 2026-06-16
Document #aligned and #unaligned

vbgl:
Some words about these `#aligned` and `#unaligned` annotations introduced in #748.

### DIFF
--- a/docs/source/language/syntax/expressions.md
+++ b/docs/source/language/syntax/expressions.md
@@ -5,12 +5,12 @@
   | <bool>  // Boolean constant.
   | <string> // String constant.
   | <var>  // Variable.
-  | [<expr>] // Memory access.
-  | [:<wsize> <expr>] // Memory access.
-  | <var>[<expr>]  // Array access.
-  | <var>[:<wsize> <expr>]  // Array access.
-  | <var>.[<expr>]  // Unscaled array access.
-  | <var>.[:<wsize> <expr>]  // Unscaled array access.
+  | [<align> <expr>] // Memory access.
+  | [<align>:<wsize> <expr>] // Memory access.
+  | <var>[<align> <expr>]  // Array access.
+  | <var>[<align>:<wsize> <expr>]  // Array access.
+  | <var>.[<align> <expr>]  // Unscaled array access.
+  | <var>.[<align>:<wsize> <expr>]  // Unscaled array access.
   | <var>[<expr> : <expr>]  // Subarray.
   | <var>[:<wsize> <expr> : <expr>]  // Subarray.
   | <var>.[<expr> : <expr>]  // Unscaled subarray.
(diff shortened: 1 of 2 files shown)


docs/source/language/syntax/expressions.md
```diff
@@ -167,13 +172,38 @@ type-casting are supported: `a.[i:N]`, `a[u16 i:N]`, `a.[u16 i:N]`.
...
-Suppose a register `reg u64` whose value is a memory address.
+Suppose a register variable `reg u64 pointer` whose value is a memory address.
 Then, for loading values from memory the notation is as follows:
 ```
 reg u64 var;
-var = [ptr + offset];
+var = [pointer + offset];
 // or for better understanding
-var = (64u)[ptr + offset];
+var = [:u64 ptr + offset];
```
eponier:
```suggestion
var = [:u64 pointer + offset];
```

docs/source/language/syntax/expressions.md
```diff
@@ -167,13 +172,38 @@ type-casting are supported: `a.[i:N]`, `a[u16 i:N]`, `a.[u16 i:N]`.
...
 ```
-`offset` is measured in bytes.
+`offset` is measured in bytes and may be an expression, depending on the
+addressing modes available on the target platform.
+
+## Alignment assumptions
+
+Memory and array accesses may have alignment annotations, either `#aligned` or
+`#unaligned`. They denote whether, *after compilation*, the pointer (i.e.,
+memory address) underlying the access is aligned, i.e., a multiple of the size
```
bgregoir, line 190: Maybe: the size -> the size in bytes

bgregoir (dismissed):

docs/source/language/syntax/expressions.md
vbgl: Thanks for the suggestion. Done.

docs/source/language/syntax/expressions.md
```diff
@@ -167,13 +172,38 @@ type-casting are supported: `a.[i:N]`, `a[u16 i:N]`, `a.[u16 i:N]`.
...
+1. when selecting instructions and generating assembly, to pick appropriate
+   instructions and discharge their preconditions regarding alignment (if any);
+2. when allocating local variables into the stack, the layout of each frame is
+   computed so that these assumptions are met (knowing that each `export`
+   function enforces that the stack pointer is sufficiently aligned initially).
 
+Therefore, some of these annotations hold by virtue of the correctness of the
+compiler, but the other ones are **assumptions**, used as trusted hints by the
+compiler. Notably the ones that involve pointers or arrays that are arguments to
+export functions. The [safety checker](../../tools/safety_checker) may prove
```
eponier, line 207: I'm not sure I understand. What can the safety checker prove about alignment?

docs/source/language/syntax/expressions.md
vbgl: Indeed. Thanks.

docs/source/language/syntax/expressions.md
vbgl: That memory accesses are safe (under the assumptions reported at the end of the analysis).

eponier: So it always ends up with assumptions about the alignment of arguments. I don't think it is really clear, but anyway it probably will change soon, and I don't know what to suggest, so I'm ok with it.

eponier (approved):

PR #1484, opened 2026-06-16
Fewer stack zeroization tests

eponier:
# Description

We used to test stack zeroization on (nearly) all .jazz files in `examples` and `tests`. Now, we test it only on the files in one subfolder `stack_zeroization`. The main goal was to reduce the number of tests: it goes down from around 1700 to 1000.

One might argue that we now don't test the feature enough... We might want to add a few more files in this subfolder.

cc @bgregoir


eponier:
Bonus: we test stack zeroization on risc-v

vbgl:
I’m not sure I understand the last commit: if the test is meaningful for zeroization, it might be moved to the proper directory instead.

eponier:
This test had been there before we added stack zeroization. When we decided to systematically check stack zeroization on every file, we had to add the annotation, because otherwise it did not compile. But the file was not meant for stack zeroization particularly.

We might think that the check was interesting and add something similar to the `stack-zeroization` directory. But the thing is that we know that we don't deal with `randombytes` properly, since we do not take into account the stack size it might use...

vbgl (approved):

vbgl:
Thanks for the clarification (and the nice clean-up!).

bgregoir (commented):
I am clearly in favor of these kinds of changes.

bgregoir:
Sorry I didn't see that I was already merged ...

eponier:
> Sorry I didn't see that I was already merged ...

*you* are merged? :open_mouth:

PR #1487, opened 2026-06-23
Linter: check that no “inline” variable depend on a non-inline

vbgl:
# Description

New linting analysis & check that inline variables do not depend too much on non-inline data.

# Checklist

### DIFF
--- a/compiler/linter/Analysis/TaintAnalysis.ml
+++ b/compiler/linter/Analysis/TaintAnalysis.ml
@@ -1,0 +1,244 @@
+open Jasmin.Syscall_t
+open Jasmin.Utils
+open Jasmin.Prog
+open Annotation
+
+type dependencies = Sv.t Mv.t
+(** Maps each variable to the set of variables it depends on. *)
+
+type writeset = { get_writeset : Sv.t } [@@unboxed]
+(** The set of variables that are written by a statement. *)
+
+type signatures = (var list * var list) Mf.t
+(** Maps each function name to its signature (parameters and returned
+    variables). *)
+
+let empty_writeset = { get_writeset = Sv.empty }
+
+let merge_writesets w1 w2 =
+  { get_writeset = Sv.union w1.get_writeset w2.get_writeset }
+
+let get_dependencies (deps : dependencies) (x : var) : Sv.t =
+  Mv.find_default Sv.empty x deps
+
+let kill_dependencies (deps : dependencies) (x : var) : dependencies =
+  Mv.remove x deps
+
+let set_dependencies (deps : dependencies) (x : var) (d : Sv.t) : dependencies =
+  if Sv.is_empty d then kill_dependencies deps x else Mv.add x d deps
+
+let join_dependencies (deps : dependencies) (x : var) (d : Sv.t) : dependencies
+    =
+  Mv.modify_def Sv.empty x (Sv.union d) deps
+
+let merge_dependencies d d' : dependencies =
+  Mv.merge
+    (fun _x o1 o2 ->
+      match (o1, o2) with
+      | Some s1, Some s2 -> Some (Sv.union s1 s2)
+      | None, o | o, _ -> o)
+    d d'
+
+let is_included d d' : bool =
+  Mv.for_all (fun x s -> Sv.subset s (get_dependencies d' x)) d
+
+let write_dependencies (d : Sv.t) deps (x : lval) : dependencies =
+  match x with
+  | Lnone _ | Lmem _ -> deps
+  | Lvar x -> set_dependencies deps (L.unloc x) d
+  | Laset (_, _, _, x, e) | Lasub (_, _, _, x, e) ->
+      let d' = vars_e e in
+      join_dependencies deps (L.unloc x) (Sv.union d d')
+
+let written_lvs w xs =
+  { get_writeset = List.fold_left written_lv w.get_writeset xs }
+
+let after_join deps d w : dependencies =
+  Sv.fold (fun x deps -> join_dependencies deps x d) w.get_writeset deps
+
+let analyse_call_args deps args es =
+  List.fold_left2 (fun d x e -> set_dependencies d x (vars_e e)) deps args es
+
+let analyse_stmt (context : signatures) dom stmt =
+  let rec analyse_stmt dom stmt = List.fold_left_map analyse_instr dom stmt
+  and analyse_instr ((_w, deps) as dom) instr =
+    let dom, i_desc = analyse_instr_r dom instr.i_desc in
+    (dom, { instr with i_desc; i_info = Annotation deps })
+  and analyse_instr_r (w, deps) ir =
+    match ir with
+    | Cassgn (x, tg, ty, e) ->
+        let d = vars_e e in
+        ( ( { get_writeset = written_lv w.get_writeset x },
+            write_dependencies d deps x ),
+          Cassgn (x, tg, ty, e) )
+    | Copn (xs, tg, op, es) ->
+        let d = vars_es es in
+        ( (written_lvs w xs, List.fold_left (write_dependencies d) deps xs),
+          Copn (xs, tg, op, es) )
+    | Ccall (xs, fn, es) ->
+        let args, rets = Mf.find fn context in
+        let w =
+          {
+            get_writeset =
+              (let w =
+                 List.fold_left (fun s x -> Sv.add x s) w.get_writeset args
+               in
+               List.fold_left (fun s x -> Sv.add x s) w rets);
+          }
+        in
+        let deps = analyse_call_args deps args es in
+        let deps =
+          List.fold_left2
+            (fun d lv x -> write_dependencies (Sv.singleton x) d lv)
+            deps xs rets
+        in
+        ((written_lvs w xs, deps), Ccall (xs, fn, es))
+    | Csyscall ([ x ], RandomBytes len, es) ->
+        (* The result of #randombytes does not depend on its argument. *)
+        ( ( { get_writeset = written_lv w.get_writeset x },
+            write_dependencies Sv.empty deps x ),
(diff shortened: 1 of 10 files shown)


docs/source/tools/linter.md
```diff
@@ -123,3 +124,22 @@ fn only_one_live_result(reg u64 x) -> reg u64 {
   return x;
 }
 ~~~
+
+## Flows from non-inline variables to inline variables
+
+This checks warns about inline variables whose values depend on the values of
```
eponier:
```suggestion
This check warns about inline variables whose values depend on the values of
```

docs/source/tools/linter.md
```diff
@@ -123,3 +124,22 @@ fn only_one_live_result(reg u64 x) -> reg u64 {
...
+non-inline variables. This situation usually leads to compilation failure. For
+instance in the following program, the number of iteration of a `for` loop
+depends on a run-time value (the argument `up`); therefore the loop cannot be
+unrolled an the program cannot be compiled.
```
eponier:
```suggestion
unrolled and the program cannot be compiled.
```

compiler/linter/Analysis/TaintAnalysis.ml
```diff
@@ -0,0 +1,240 @@
...
+  { get_writeset = Sv.union w1.get_writeset w2.get_writeset }
+
+let get_dependencies (deps : dependencies) (x : var) : Sv.t =
+  Mv.find_default Sv.empty x deps
+
+let kill_dependencies (deps : dependencies) (x : var) : dependencies =
+  Mv.remove x deps
+
+let set_dependencies (deps : dependencies) (x : var) (d : Sv.t) : dependencies =
+  if Sv.is_empty d then kill_dependencies deps x else Mv.add x d deps
```
bgregoir, line 28: minor remark: why do you do this test instead of doing simply Mv.add

compiler/linter/Analysis/TaintAnalysis.ml
```diff
@@ -0,0 +1,240 @@
...
+let written_lvs w xs =
+  { get_writeset = List.fold_left written_lv w.get_writeset xs }
+
+let after_join deps d w : dependencies =
+  Sv.fold (fun x deps -> join_dependencies deps x d) w.get_writeset deps
+
+let analyse_call_args deps args es =
+  List.fold_left2 (fun d x e -> set_dependencies d x (vars_e e)) deps args es
+
+let analyse_stmt (context : signatures) dom stmt =
```
bgregoir, line 62: Other stupid question, this seems to be a forward analysis. 
There is no functor that allows to do that ?

compiler/linter/Analysis/TaintAnalysis.mli
```diff
@@ -0,0 +1,20 @@
...
+    it is OK to assign Low (inline) variables in High contexts as long as the
+    Low variable is not used outside of said context.
+
+    Dependencies are not transitively closed; for instance, if X depends on Y
+    and Y depends on Z, the dependency of X on Z might be overlooked. This is
+    meant to reduce the amount of reported warnings and produce more focused
+    diagnostic messages.
+
+    Current limitation: flows through memory are ignored, therefore loading a
+    value from memory into an inline variable will not cause any warning to be
```
bgregoir, line 14: Question: is it difficult to add a boolean indicating that the variable depend on loaded value ?

compiler/tests/negative.expected
```diff
@@ -1482,6 +1482,12 @@ internal compilation error:
   unrolling: loop iterator too small
 Please report at https://github.com/jasmin-lang/jasmin/issues
 
+fail/unroll/x86-64/run-time-bound.jazz:
+
+"fail/unroll/x86-64/run-time-bound.jazz", line 5 (2) to line 7 (3):
+compilation error in function count:
```
bgregoir, line 1542: This is really the error message you want to see? or you want to see the one generate by the linter ?

compiler/tests/warnings.t
```diff
@@ -89,3 +89,15 @@
   warning: the variable g is already declared at "warning/common/bug_1347.jazz", line 2 (6-7)
   "warning/common/bug_1347.jazz", line 8 (20-21)
   warning: the variable C::x is already declared at "warning/common/bug_1347.jazz", line 7 (20-21)
+
+  $ ../jasminc warning/common/inline.jazz
+  "warning/common/inline.jazz", line 5 (2-12)
+  warning: Inline parameter “x” depends on (non-inline) variable “a”
+  "warning/common/inline.jazz", line 7 (2-23)
+  warning: Inline variable “b” depends on (non-inline) variable “z”
```
bgregoir: Do you think it is possible to indicate that z come from the function zero or maybe simply print the position where z is declared

compiler/tests/warnings.t
```diff
@@ -89,3 +89,15 @@
...
+  "warning/common/inline.jazz", line 16 (2-9)
+  warning: Inline variable “t” depends on (non-inline) variable “c”
```
bgregoir: Is it possible to print this message line 15 ?

compiler/tests/warnings.t
```diff
@@ -89,3 +89,15 @@
...
+  "warning/common/inline.jazz", line 22 (2-19)
+  warning: For loop counter “i” depends on (non-inline) variable “p”
+  "warning/common/inline.jazz", line 42 (2-9)
```
bgregoir: same remark, line 39 will be simpler to follow

compiler/linter/Analysis/TaintAnalysis.ml
vbgl: To ensure that no variable is bound to an empty set (makes the map smaller).

compiler/linter/Analysis/TaintAnalysis.ml
vbgl: I think this analysis is not an instance of the generic one provided by the functor. Please correct me if I’m wrong.

compiler/linter/Analysis/TaintAnalysis.mli
vbgl: I don’t think it is difficult. But for the purpose of this check of inline variables, this is really useless.

compiler/tests/negative.expected
vbgl: The linter is not run there, so it is expected that the warning does not show up.

compiler/tests/warnings.t
vbgl: Dead assignments are fine: warnings are triggered only when the variable is used. I think you are right, but this is not easy to solve. Do you have any suggestion?

bgregoir: maybe a combination with a liveness analysis. If an inline variable is assigned, its value depend on none inline variable and and the variable in live, then report an error.

vbgl: Consider the following program.

~~~
fn f(reg u32 x) -> reg u32 {
  inline int i = 0;
  if x == 0 {
    i = 1; // does i depends on x ? not yet
    x = i; // this is fine
  }
  // here i depends on x
  x += i; // this is bad
  return x;
}
~~~

Where do you want a warning? How is liveness supposed to help?

compiler/tests/warnings.t
vbgl: I’ve updated the message to include declaration position, e.g.,

> "/home/gitlab-runner/builds/_6kTXlVEu/0/jasmin-lang/jasmin/libjade/arm-m4/ref/common/hashing.jinc", line 339 (4-59)
> warning: Inline parameter “i” depends on (non-inline) variable “j (declared at "/home/gitlab-runner/builds/_6kTXlVEu/0/jasmin-lang/jasmin/libjade/arm-m4/ref/common/hashing.jinc", line 302 (14-15))”

compiler/tests/warnings.t
bgregoir: For this example, I think on the if. After merging the two branches and add the dependency on `x`, `i` depends on `x` and it is assigned by the if instruction. 
The liveness analysis allows to know that `i` is live after the if.

compiler/tests/warnings.t
bgregoir: This is much better

compiler/linter/Analysis/TaintAnalysis.ml
```diff
@@ -0,0 +1,241 @@
...
+        Format.fprintf fmt "%s “%a” depends on (non-inline) variable%a" kind
+          pp_var x pp_varset y);
+  }
+
+let check_func (context : signatures) fd : CompileError.t list =
+  let errors = ref [] in
+  let check_set ~kind ~loc x d =
+    let ys =
+      Sv.filter
+        (fun y -> match y.v_kind with Stack _ | Reg _ -> true | _ -> false)
```
bgregoir: I think it is better to replace the _ by the explicite constructor.

vbgl: Fixed.

compiler/tests/warnings.t
vbgl: Done (see last commit). Note that the following program produces two warnings. Is that right?

~~~
fn twice(reg u32 a) -> reg u32 {
  inline u32 b = 0;
  if a == 0 {
    b = a;
    a = b;
  }
  a += b;
  a += b;
  return a;
}
~~~

bgregoir: I think it is ok to provide two warnings here. 
But, if you don't want you can maintain a set of variables where a warning is already printer (each time a variable is assigned remove it from the set)

bgregoir: Do you want to improve this or we can merge ?

vbgl: I think it is OK to merge as is.

bgregoir (approved):

PR #1491, opened 2026-06-26
"Remove unused returns" also updates annotations

eponier:
The pass used to update the list of returned values (and the list of output types), but never touched the associated list of annotations, while they should be kept in sync.

This made the compiler crash when trying to print the program just after the pass (after https://github.com/jasmin-lang/jasmin/pull/1474, before the annotations were incorrectly not printed).

### DIFF
--- a/compiler/src/compile.ml
+++ b/compiler/src/compile.ml
@@ -134,7 +134,7 @@ let compile (type reg regx xreg rflag cond asm_op extra_op)
        and type asm_op = asm_op
        and type extra_op = extra_op) visit_prog_after_pass prog cprog =
   let module RA = Regalloc.Regalloc (Arch) in
-  let module StackAlloc = StackAlloc.StackAlloc (Arch) in
+  let module SA = StackAlloc.StackAlloc (Arch) in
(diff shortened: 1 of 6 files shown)


eponier:
According to @bgregoir , these annotations were used for (S)CT. But a different kind of annotations was introduced for (S)CT, so an alternative patch could be to just remove these annotations.

eponier:
I don't like these `Obj.t` and `Obj.magic`, but I didn't find how to avoid them.

eponier:
Does it need a changelog entry?

eponier:
Oops, I had forgotten to update the proof...

vbgl:
> I don't like these `Obj.t` and `Obj.magic`, but I didn't find how to avoid them.

Here is a proposal (stay first order).

eponier:
> > I don't like these `Obj.t` and `Obj.magic`, but I didn't find how to avoid them.
> 
> Here is a proposal (stay first order).

Why not. What I like in the first version is that it's clearer that we apply the same operation on the 3 lists, but the `Obj` occurrences are dreadful.

bgregoir (approved):

PR #1489, opened 2026-06-25
One semantics

bgregoir:
This PR remove the big step semantics and keep only the semantics based on itree.

PR #1495, opened 2026-07-02
Compatibility with Rocq 9.2 (and master)

eponier:
Fix warning -level-tolerance, mute others.

Depends on coq-core rather than coq in opam.

Note that there is currently no mathcomp release compatible with Rocq 9.2, so we are still not compatible with Rocq 9.2 because of that.

### DIFF
--- a/opam
+++ b/opam
@@ -33,7 +33,8 @@ depends: [
   "yojson" {>= "1.6.0"}
   "angstrom" {>= "0.14.0"}
   "ocamlfind" { build }
-  "coq" {>= "9.0" & < "9.2~"}
+  "coq-core" {>= "9.0" & < "9.3~"}
+  "coq-stdlib"
   "coq-elpi" {>= "2.5.1"}
   "coq-mathcomp-ssreflect" {>= "2.3" & < "2.6~"}
   "coq-mathcomp-algebra"
--- a/proofs/3rdparty/xseq.v
+++ b/proofs/3rdparty/xseq.v
@@ -67,7 +67,7 @@ case: eqP=> [->|/eqP ne_xt] t_notin_s; last first.
 + by rewrite in_cons eqE /= (negbTE ne_xt).
 rewrite inE eqE /= eqxx /=; case: eqP => [->|ne_wu] _ /=.
 + by constructor.
-suff ->: (t, w) \in s = false by constructor; case=> /esym.
+suff ->: ((t, w) \in s) = false by constructor; case=> /esym.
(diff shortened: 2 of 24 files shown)


eponier:
For the first commit, here are some remarks:
- I don't understand most of the notation stuff, adding `right associativity` made some warnings disappear, so I used that, but I'm not sure it was the right thing to do
- the situation of `>>` is a bit painful. We have to pick level 30 to be compatible with the standard library where it is defined with `no associativity` (see https://github.com/jasmin-lang/jasmin/pull/204), but level 30 is right associative because of notation `x ^ y` defined in Corelib. According to [this footnote](https://rocq-prover.org/doc/master/refman/user-extensions/syntax-extensions.html#no-associativity) in the doc of Rocq, `no associativity` is actually `left associativity`... The only case where it plays a role is in `linearization.v`, in function `check_i`. Printing the parsed term, it seems that Rocq chooses left associativity after all... I added parentheses to enforce right associativity (I needed to add parentheses anyway to avoid the warning), but I don't know what choice is better. Proofs pass with both choices (probably thanks to `t_xrbindP`).
- surprisingly, `=> //` and `move => //` are not the same. The latter seems to simplify terms in some cases.

eponier:
On my machine, extraction fails similarly to [here](https://github.com/jasmin-lang/jasmin/pull/1413#issuecomment-4040851515). @vbgl , I guess you circumvented the issue in the CI. Do you remember what is the culprit?

eponier:
CI revealed other warnings in the coq-master job. Let's see if I caught them all.

vbgl:
> Do you remember what is the culprit?

Rocq 9.2. Fortunately Rocq 9.3 is around the corner.

eponier:
> > Do you remember what is the culprit?
> 
> Rocq 9.2. Fortunately Rocq 9.3 is around the corner.

Oh ok. Around the corner is in a few months, though :sweat_smile:

eponier:
I forgot to deal with `at-level-200-changed`, but actually I don't know what the proper fix is for the few occurrences (follow the advice? just mute the warning?). Any idea?

eponier:
> I forgot to deal with `at-level-200-changed`, but actually I don't know what the proper fix is for the few occurrences (follow the advice? just mute the warning?). Any idea?

There were 3 of them. @bgregoir suggested to turn two of them into abbreviations. The last one is `linearization_proof.v`, that is about to be removed in https://github.com/jasmin-lang/jasmin/pull/1489, so I leave it as it is.

This is good to go!

vbgl (approved):
Thanks!

PR #1501, opened 2026-07-07
ARM: add instructions SXTB and SXTH

### LINKED ISSUE #805: Support instructions SXTB and SXTH
These are the same as UXTB and UXTH that are already implemented.

vbgl:
# Description

Adds two sign-extension instructions of the ARMv7-M architecture.

Fixes #805

# Checklist

### DIFF
--- a/proofs/compiler/arm_instr_decl.v
+++ b/proofs/compiler/arm_instr_decl.v
@@ -130,6 +130,8 @@ Variant arm_mnemonic : Type :=
 | UXTB                           (* Extract a byte and zero extend *)
 | UXTH                           (* Extract a halfword and zero extend *)
 | SBFX                           (* Extract a sub-word and sign extend *)
+| SXTB                           (* Extract a byte and sign extend *)
+| SXTH                           (* Extract a halfword and sign extend *)
 | CLZ                            (* Count leading zeros. *)
 
 (* Comparison *)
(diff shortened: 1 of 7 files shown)


proofs/compiler/arm_instr_decl.v
```diff
@@ -2078,6 +2082,60 @@ Definition arm_SBFX_instr : instr_desc_t :=
     id_semi_safe := fun _ => @bit_field_extract_semi_safe sh;
   |}.
 
+Definition sign_extend_bits_semi
+  (len: wsize) (wn: wreg) (wroram: u8) : wreg :=
+  sign_extend reg_size (wrepr len (Z.shiftr (wunsigned wn) (wunsigned wroram))).
```
bgregoir: I think the Z.shiftr is wrong it should be a rotation. 
maybe something like: 
```
sign_extend reg_size (zero_extend len (wror wn (wunsigned wroram)))
```

vbgl: OMG, you are right. I’m genuinely curious about any serious use of this feature (e.g., rotate by 24, extract 16, sign-extend).

bgregoir: For 8, it seems clearer. you can sign extend any byte of the word. I assume it is the same to 16bits if you rotate by 0 or 16.

vbgl: I might be wrong, but I understand that in the case of 8 (SXTB), rotation and shift would be equivalent. For 16, the difference between shift and rotation is only relevant when the rotation amount is 24. In this case, you start from four bytes [ A, B, C, D ], rotate to get [ B, C, D, A ], extract and get [ D, A ] and sign-extend. Hence my question: to which problem is this operation the solution?

bgregoir: for the shift I agree with you. For your question, no idea :-), maybe it is just for uniformity.

vbgl: Thanks. I’ve applied your suggestion.

bgregoir (approved):

PR #1504, opened 2026-07-07
Remove unused definitions and lemmas

vbgl:
Small clean-up.

Note that there is probably no need to keep so many small commits.

### DIFF
--- a/proofs/compiler/inline_proof.v
+++ b/proofs/compiler/inline_proof.v
@@ -79,46 +79,6 @@ Proof.
   by apply: rbindP => ? /Hrec -> /=;apply:rbindP => ?? [] <-.
 Qed.
 
-Lemma inline_progP p p' f fd' :
-  uniq [seq x.1 | x <- p] ->
-  inline_prog' p = ok p' ->
-  get_fundef p' f = Some fd' ->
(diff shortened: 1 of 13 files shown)


proofs/compiler/inline_proof.v
```diff
@@ -264,30 +212,6 @@ Qed.
 
 Section IT.
 
-Section AUX.
-
-(* Technical lemma: we don't have equality of the type signatures anymore,
-   they are only convertible. We still manage to prove some equality, so luckily
-   we don't have to mess with itrees.
-   The lemma is put inside a different section, because we want to pick generic
-   instances of the typeclasses involved. *)
```
eponier, line 273: According to this comment, the lemma was carefully crafted, but it's not needed in the end?

vbgl: Its use have been removed there: https://github.com/jasmin-lang/jasmin/commit/14d5c1c7ce27a07ca003387817b8a572671d7189

eponier: Oh, I wonder who wrote that commit :sweat_smile:

eponier (approved):

proofs/compiler/inline_proof.v
eponier: (and apparently I added the lemma initially, frankly, I don't recall at all)

PR #1505, opened 2026-07-08
Heuristic register-allocation oracle

vbgl:
# Description

This provides a way to skip the first run of register-allocation.

# Checklist


changes/01-feature/1505-callee-saved.md
```diff
@@ -0,0 +1,3 @@
+- Experimental heuristics for computing how many callee-saved registers are
+  needed by each export function
```
bgregoir: maybe add the fact that the heuristics can be controled using the `-callee-saved` option

vbgl: Done.

bgregoir (approved):

PR #1506, opened 2026-07-08
Experiment: add .vos targets

eponier:
for fast compilation of proofs (it can be useful in some cases when developing)

The first two commits are useful whatever is decided about this PR.
The third commit adds `Proof using` everywhere. Without them, `.vos` is useless in the presence of sections.
The last commit adds `.vos` targets.

The question is whether we consider the `.vos` targets useful enough to justify maintaining `Proof using` clauses (not a huge burden, but still adds some rigidity).


bgregoir:
Can we use : Default Proof Using. To reduce the impact of the 3th commit?

eponier:
We could add `Set Default Proof Using "".` so that `Proof.` is actually `Proof using.` (meaning we just need to make changes when there are section variables that are captured). Is it what you think about? I have no strong opinion about it.

bgregoir:
Yes, I view to advantages:
 - Reduce the size of the commit. 
 - I think it is more stable, we will do new lemma starting with "Proof." and Rocq will complains if we use other section hypothesis. No risk that I will forgot to put a "Proof using."

eponier:
I followed your suggestion, and added a line to the `_CoqProject` file. The syntax is a bit strange, but I think this works.

vbgl:
LGTM. What will happen if I forget to write `Proof using`?

eponier:
With the option set in `_CoqProject`, `Proof.` is actually `Proof using.`. So if the lemma does not use any section variable, it will be fine. If it uses some, the proof will be rejected with a message listing the section variables used in the proof. The only risk is not to write `Proof.` at all, but the risk is really small, and there is a warning coming in Rocq 9.3 about this.

vbgl (approved):

PR #1508, opened 2026-07-08
Remove unused definitions & lemmas

vbgl:
Some more cleaning.

### DIFF
--- a/proofs/compiler/stack_alloc.v
+++ b/proofs/compiler/stack_alloc.v
@@ -939,7 +939,7 @@ Definition get_global (x:var_i) :=
 
 Definition get_local (x:var) := Mvar.get pmap.(locals) x.
 
-Definition get_var_kind x :=
+Definition get_var_kind x : cexec var_kind :=
   let xv := x.(gv) in
   if is_glob x then
     Let z := get_global xv in
--- a/compiler/tests/success/x86-64/empty_string.jazz
+++ b/compiler/tests/success/x86-64/empty_string.jazz
@@ -2,8 +2,8 @@
 
 param int N = 0;
 
-inline fn f(inline u8[N] a) {}
+inline fn f(inline u8[N] a) -> inline u8[N] { return a; }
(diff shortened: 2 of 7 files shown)


proofs/compiler/stack_alloc.v
```diff
@@ -939,7 +939,7 @@ Definition get_global (x:var_i) :=
 
 Definition get_local (x:var) := Mvar.get pmap.(locals) x.
 
-Definition get_var_kind x :=
+Definition get_var_kind x : cexec var_kind :=
```
bgregoir, line 942: Simple question: it is just for documentation or there is some interest in doing this ?

proofs/lang/varmap.v
```diff
@@ -897,12 +885,6 @@ Section REL_EQUIV.
     by apply value_uincl_vm_truncate.
   Qed.
 
-  Lemma vm_uincl_set vm1 vm2 x v1 v2 :
```
bgregoir, line 900: I am always a little bit worry of removing this kind of lemma that normally should allows to reduce other proofs. But if we really don't use those lemmas...

bgregoir (dismissed):

proofs/compiler/stack_alloc.v
vbgl: `var_kind` has been defined in 7aa47e694337726edae447fc54e1965647154fad but never used. Should I just remove its definition instead?

proofs/lang/varmap.v
vbgl: We can drop the commit about `varmap`.

vbgl: Done.

vbgl:
I’ve removed the changes in `varmap.v`, cleaned the history, and added an unrelated commit (which is also some cleaning) that silences a warning in the test suite. So please don’t squash.

bgregoir (approved):

PR #1509, opened 2026-07-09
Regalloc: use “inplace” annotation as hint

vbgl:
# Description

Register allocation understands an “inplace” annotation on operations as follows. The first argument shall be allocated to the same register as the destination. The annotation is ignored if the operation has several destinations.

# Checklist


bgregoir:
I think it will be nice to add a little bit of documentation for the feature.
Maybe something in "Living in harmony with the Jasmin compiler"/"registers allocation".

vbgl:
I’ve added some documentation as suggested.

bgregoir (approved):

PR #1513, opened 2026-07-10
arch: make immediate-argument conditions architecture-local

clebreto:
The validity conditions for immediate operands used to be a single inductive in arch_decl.v mixing every architecture's constructors, and each backend's checker had to enumerate the other backends' variants to map them to false: adding an architecture meant editing every other architecture's files. The type existed globally only to keep decidable equality over arg_kind.

Give each architecture its own condition type instead, following the pattern already used for conditions ([cond]): arch_decl gains a [caimm_cond] type field with an [eqTypeC] instance, a printer for assembly-generation error messages and [check_CAimm] over it, and [arg_kind] carries an [option caimm_cond] (None replacing the shared CAimmC_none). Decidable equality over arg_kind is written by hand through the eqTypeC bridge, which removes the reason the type was global. The per-architecture condition types, checkers and printers move next to their architectures: arm and riscv each define exactly their own constructors, and x86, which has no immediate conditions, uses the empty type. No architecture mentions another's conditions anywhere, and adding an architecture no longer touches existing ones. The generic assembly-generation proofs needed no change.


proofs/arch/arch_decl.v
```diff
@@ -305,14 +300,36 @@ Definition check_oreg or ai :=
...
  *)
-#[only(eqbOK)] derive
 Variant arg_kind :=
 | CAcond
 | CAreg
 | CAregx
 | CAxmm
 | CAmem of bool (* true if Global is allowed *)
-| CAimm of caimm_checker_s & wsize.
+| CAimm of option caimm_cond & wsize.
```
bgregoir: why there is an option type, if needed it can be done by the arch directly.

clebreto: Good point — done in cf8cef4c. I dropped the option: CAimm now carries caimm_cond directly, and each architecture provides a distinguished caimm_none : caimm_cond value (a new arch_decl field) that every immediate satisfies. The generic CAimm_sz helper uses caimm_none, so the shared ak_*_imm argument-kind helpers still work for the architectures that accept an unconstrained immediate — which is why the "none" case needed a generic representation in the first place. Now it's handled by each architecture directly: arm and riscv add a none constructor to their own condition type, and x86 (which has only unconstrained immediates) uses a single-constructor type. No generic option, and no architecture references another's conditions.

bgregoir: it is in fact used on all arch

bgregoir (approved):

PR #1517, opened 2026-07-15
fs_uincl_initialize and fs_uincl_finalize: relax hypotheses on f_params & f_res

vbgl:
There is no need to constrain the metadata that occurs in the parameters & returned variables.

This makes these theorems reusable even in contexts where the equality on metadata is not available (remember: `var_i` is not `EqType`).

### DIFF
--- a/proofs/lang/psem.v
+++ b/proofs/lang/psem.v
@@ -644,33 +644,55 @@ Lemma eq_initialize p' fd fd' fs s:
   initialize_funcall p' ev fd' fs = ok s.
 Proof. by rewrite /initialize_funcall => <- <- <- <-. Qed.
 
+Lemma get_eq_var_is wdb xs ys vm :
+  eq_var_is xs ys →
+  get_var_is wdb vm xs = get_var_is wdb vm ys.
+Proof.
+  elim: xs ys; first by case.
+  by move => x xs ih [] // y ys /andP[] /eqP /= -> /ih ->.
+Qed.
+
+Lemma write_vars_eq_var_is wdb xs ys vs s :
+  eq_var_is xs ys →
+  write_vars wdb xs vs s = write_vars wdb ys vs s.
+Proof.
+  elim: xs ys vs s; first by case.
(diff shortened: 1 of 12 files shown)


proofs/lang/psem.v
```diff
@@ -644,33 +644,55 @@ Lemma eq_initialize p' fd fd' fs s:
...
+  write_vars wdb xs vs s = write_vars wdb ys vs s.
+Proof.
+  elim: xs ys vs s; first by case.
+  move => x xs ih [] // y ys [] // v vs s /andP[] /eqP h /ih{}ih.
+  rewrite /= /write_var h.
+  case: set_var => ?; last done.
+  exact: ih.
+Qed.
+
 (* TODO: Can we generalize this to different semantic ? *)
```
bgregoir, line 666: I don't understand why you don't provide 2 lemmas: 
```
Lemma fs_uincl_initialize_eq_var_is p' fd fd' fs fs' s:
  f_tyin fd = f_tyin fd' ->
  f_extra fd = f_extra fd' ->
  eq_var_is (f_params fd) (f_params fd') ->
  p_extra p = p_extra p' ->
  fs_uincl fs fs' ->
  initialize_funcall p ev fd fs = ok s ->
  exists2 s', initialize_funcall p' ev fd' fs' = ok s' & st_uincl tt s s'.
```
and 
```Lemma fs_uincl_initialize p' fd fd' fs fs' s:
  f_tyin fd = f_tyin fd' ->
  f_extra fd = f_extra fd' ->
  f_params fd = f_params fd' ->
  p_extra p = p_extra p' ->
  fs_uincl fs fs' ->
  initialize_funcall p ev fd fs = ok s ->
  exists2 s', initialize_funcall p' ev fd' fs' = ok s' & st_uincl tt s s'.
Proof. (* Direct consequence of the first lemma *)
``` 
Thus you don't need to change the other files.
It is the same remark for `fs_uincl_finalize`.

vbgl: Because changing the other files was fairly minimal: usually fix the spelling of the reflexivity lemma.

PR #1519, opened 2026-07-15
BT, BTS, and BTR now set OF, SF, and PF to undef

sarranz:
# Description

`BT`, `BTS`, and `BTR` should set `OF`, `SF`, and `PF` to undefined. This means that these instructions now return three more results, so existing code will break. I don't know if I can add more tests other than updating the existing positive ones.

# Checklist


vbgl:
You probably need to update the EasyCrypt model of these instructions.

sarranz:
Sorry, I forgot. Now it should be fine.

vbgl (approved):

Issue #1013, opened 2025-01-21
RISC-V: CT and doit are wrong


eponier:
We modelled RISC-V with every instruction being both CT and doit. Now that we have DIV and REM, this is certainly false.

This is not hard to do but I don't have the time right now. In the RISC-V spec, there is something called Zkt which, if mentioned by a device, means that the instructions listed are guaranteed to be CT. There is no flag, unlike x86 and ARM.

There is also a notion of hints, encoded with instructions whose destination register is x0. The instruction can be in Zkt (e.g. it can be ADDI). One must just check that this is not problematic for our modeling of CT and doit.

J08nY:
Related links:
https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/exts/Zkt.html
source:
https://github.com/riscv-software-src/riscv-unified-db/blob/main/arch/ext/Zkt.yaml

PR #1520, opened 2026-07-16
arch: carry the DOIT classification in the instruction descriptors

clebreto:
The DOIT/DIT classification used by jasmin-ct --doit was kept in hand-written OCaml tables (is_doit_asm_op / is_doit_asm_extra in {x86,arm,riscv}_arch_full.ml), disconnected from the instruction descriptions.

Move this information into the descriptors:
- add a boolean field id_doit to arch_decl.instr_desc_t and a corresponding field i_doit to sopn.instruction_desc; base ops forward id_doit to i_doit (arch_extra.get_instr_desc);
- x86: the per-operator classification is transcribed from the OCaml table into x86_op_is_doit and stamped into the descriptor by x86_instr_desc;
- arm: every instruction description sets id_doit, transcribing the per-mnemonic OCaml table; mk_cond, mk_shifted and idt_dropn preserve the field;
- riscv: all instructions are classified as DOIT, as before;
- extra operators (x86_extra, arm_extra, riscv_extra) are all classified as DOIT, as in the OCaml tables; pseudo and SLH operators are classified as DOIT, which matches the CT checkers that treat them as constant-time;
- delete the OCaml tables; Arch.is_ct_sopn ~doit:true now reads i_doit from the instruction descriptor.


vbgl (approved):
LGTM. Let’s squash & fix the commit message.

eponier:
(reminder, orthogonal to this PR: DOIT is wrong for RISC-V, see https://github.com/jasmin-lang/jasmin/issues/1013)

proofs/compiler/x86_extra.v
```diff
@@ -123,7 +123,7 @@ Definition Ox86SLHinit_instr :=
       [:: ty_msf ]
       [:: E 0 ]
       se_init_sem
-      true.
+      true NOT_DOIT.
```
eponier, line 126: Sorry, late to the party, but why is this `NOT_DOIT` while it was before?

bgregoir:
Because it use LFENCE which is not DOIT (which does not really make sense since the instruction take 0 argument),
so it change nothing.

eponier:
Thanks!

PR #1521, opened 2026-07-16
x86: SSE instructions need aligned memory

vbgl:
# Description

Type-4 instruction according to table 2-15.

# Checklist


bgregoir (commented):
Should we also add negative test ?

bgregoir (approved):

PR #1530, opened 2026-07-31
Minor fix in pretyping

eponier:
As it already happened, some arguments needed to be swapped.

I added test files, but I'm unsure they are really useful.

### DIFF
--- a/compiler/src/pretyping.ml
+++ b/compiler/src/pretyping.ml
@@ -1960,7 +1960,7 @@ let create_is_mem_init pd loc args =
     let e2 = cast_int loc None e2 t2 in
     P.Pis_mem_init (e1,e2)
   else
-    rs_tyerror ~loc (InvalidArgCount(2, List.length args))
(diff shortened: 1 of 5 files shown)


compiler/src/alias.ml
```diff
@@ -156,6 +156,8 @@ let compare_gvar params x gx y gy =
          match is_ptr x.v_kind, is_ptr y.v_kind with
          | true, false -> check_size "local" "pointer" x sx y sy; -1
          | false, true -> check_size "local" "pointer" y sy x sx; 1
+         | true, true ->
+             hierror_no_loc "%a is a reg ptr that is neither an argument nor points to the stack" pp_var y
```
vbgl: Is this change intended as part of this commit & PR?

eponier: No, indeed.

vbgl (approved):
Well spotted. Thanks.

PR #1532, opened 2026-08-08
pretyping: reject vector shifts with conflicting sign prefix and suffix

### LINKED ISSUE #1441: Inconsistent signedness check for vectorized shift operators
The following program is accepted by the compiler, even though it contains a conflicting sign specification:
```jasmin
export fn main (reg u128 x) -> reg u128 {
  x = x >>s4u32 16;
  return x;
}
```
In this example, the vectorized shift right operator ```>>s4u32``` provides conflicting directives: ```s``` for signed and ```4u32``` for unsigned. The compiler currently seems to silently ignore the ```s``` prefix, relying solely on the ```4u32``` suffix to generate an unsigned vectorized shift right.

This behavior is misleading: without empirical testing or inspecting the assembly, a user cannot know a priori whether the compiler will favor the prefix or the suffix.

Furthermore, this inconsistent behavior is correctly caught and rejected for the scalar shift right. For instance, the following program:
```jasmin
export fn main (reg u32 x) -> reg u32 {
  x = x >>s32u 16;
  return x;
}
```
fails to compile with the expected typing error:
```shell
"foo.jazz", line 2 (11-14):
typing error: invalid operator >>s 32u
```

A look at the compiler source code suggests that while an explicit consistency check is performed for scalar shifts, it appears to be missing for their vectorized counterparts.

To avoid ambiguity and potential bugs, it would be preferable to enforce the same consistency checks for both scalar and vectorized shifts.

MavenRain:
# Description

The scalar shift `>>s32u` is rejected with `invalid operator >>s 32u`.  The vector form `>>s4u32` was accepted.  The compiler ignored the `s` prefix and used only the sign in the suffix.  It emitted a logical (unsigned) shift for an operator that asks for an arithmetic (signed) shift.

This change adds the same consistency check to the vector path.  The `CVS` arm of `op_info` in `compiler/src/pretyping.ml` now compares the prefix sign with the sign in the vector suffix.  If they conflict, the type checker rejects the program with the same `invalid operator` error as the scalar path.  If they agree, or if there is no prefix, the behavior does not change.  Only `>>` (`ShR`) is affected.  All other operators either take no sign prefix or reject vector suffixes before this point.

Notes on scope:

- `>>s4s32` and `>>u4u64` (agreeing signs) stay accepted.  This matches the scalar rule, where `>>s32s` is accepted.  Two such lines are added to the positive test `compiler/tests/success/x86-64/vector_expression.jazz` to pin this.
- @eponier proposed a stricter option in the issue: "In my opinion, we could even purely reject such syntax, even when both signs agree."  That variant is a small change to the same guard.  I can switch this PR to it if you prefer.  No file in this repository uses a sign prefix together with a vector suffix, so both options are compatible with the existing code and tests.

Tests: a new negative test `compiler/tests/fail/typing/x86-64/vector_shift_sign_mismatch.jazz` checks that `>>s4u32` is rejected, with its expected diagnostic in `compiler/tests/negative.expected`.

Fixes #1441

# Checklist

- [x] Add your name to `AUTHORS`

<sub>I used Claude to help prepare this change.  I reviewed and tested it.</sub>


compiler/src/pretyping.ml
```diff
@@ -829,8 +829,13 @@ let op_info exn op (s : W.signedness option) (castop:S.castop) ty ws_cmp vs_cmp
       in
       check_op_w loc op ty s ws_cmp
 
-    | CVS(vs,s,ve) ->
-      let s = tt_sign s in
+    | CVS(vs,sg,ve) ->
+      let s1 = tt_sign sg in
+      let s =
+        Option.map_default
```
vbgl, line 835: Using `map` here looks strange. Using `Option.may` might be clearer.

vbgl (commented):
Thanks for your contribution. I’ve left a minor comment about your implementation.

vbgl:
I’ve reworked this PR to address the comment and the conflicts with the main branch. Any feedback?

PR #1535, opened 2026-08-18
Require mathcomp 2.4

vbgl:
This raises the lower bound on the mathcomp dependencies and addresses the deprecation warnings introduced in that version (first commit), then updates mathcomp-word (second commit) so as to fix the `coq-master` CI job.


eponier (approved):
Thanks!

PR #1541, opened 2026-08-26
Add an ARMv8-A (AArch64) backend

clebreto:
New architecture `-arch armv8a` targeting A64 (AArch64), modeled at the full 64-bit register width.

- **Coq model**: registers x0..x30/xzr/sp, NZCV flags and conditions (shared with arm-m4 through `arm_common`), 43 mnemonics — exactly the set the compilation pipeline can emit (lowering, load-immediate and smart-address helpers, `mov_ofs`, stack zeroization, pseudo-op expansion). Each instruction carries its ARM ARM (DDI 0487 M.a) documentation block, and the DIT classification lives in the descriptors (`id_doit`), everything being DIT except SDIV/UDIV/ADR.
- **Immediate legality** is modeled by architecture-local `caimm_cond` checkers (arith imm12, bitmask immediates, MOV-alias immediates, shift amounts, MOVZ/MOVN/MOVK halfword shifts restricted per operand width).
- **Proofs**: the lowering pass and all ten `h_architecture_params` hypotheses are proved; the extraction cone contains no admitted statement.
- **OCaml**: GNU/ELF AArch64 printer, AAPCS64 calling convention, `-sp-min-align` (SP stays 16-byte aligned across calls, with a custom epilogue restore since an A64 LDR cannot target SP).
- **Tests**: `armv8a` and stack-zeroization check categories; a hardware differential test (`make -C compiler check-armv8a-semantics`, AArch64 hosts only) compares the extracted `id_semi` of every instruction form against the silicon — currently 509 forms / 21442 checks.

Validated end to end by compiling the ML-DSA reference implementations (formosa-mldsa) for the three parameter sets and running the full KAT suites.

### Previous review round

A first round by @bgregoir happened on the fork: clebreto/jasmin#5. Summary of the points and their resolutions:

- *"Do we need this file?"* (`compiler/doc/extract_armv8a_isa_docs.py`) — removed from tracking and history.
- *MOVZ/MOVN/MOVK at U32 with shifts 32/48 accepted by the checker but not encodable: "we should fix that in the model"* — fixed: the checker carries the operand width (`CAimmC_armv8a_halfword_shift`) and rejects 32/48 in the W form.
- *Stale skip note about ROR-shifted arithmetic forms* — removed; those forms are invalid in the model (`shift_allowed` in `id_valid`).
- *`jasmin_checksafety.ml` error reporting "should go in a separated PR"* — moved out (clebreto/jasmin#6); only the armv8a stub change in `safetyMain.ml` stays (error out instead of claiming safety).
- *`check-print-roundtrip` script "can go to a separated PR"* — moved out (clebreto/jasmin#7); the `armv8a-print-roundtrip` category left with it.

### DIFF
--- a/compiler/src/arm_arch_full.ml
+++ b/compiler/src/arm_arch_full.ml
@@ -63,5 +63,7 @@ module Arm (Lowering_params : Arm_input) : Arch_full.Core_arch
      would be silently rounded by the hardware. *)
   let sp_min_align = Wsize.U32
 
+  let max_store_size = Wsize.U32
+
   let internal_call_conv = Arm_decl.arm_internal_call_conv
 end
--- a/compiler/src/glob_options.ml
+++ b/compiler/src/glob_options.ml
@@ -80,6 +80,7 @@ let set_target_arch a =
     match a with
     | "x86-64" -> X86_64
     | "arm-m4" -> ARM_M4
+    | "armv8a" -> ARMv8A
     | "riscv" -> RISCV
     | _ -> assert false
   in target_arch := a'
@@ -238,7 +239,7 @@ let options = [
     "-intel", Arg.Unit (set_syntax `Intel), " Use intel syntax (default is AT&T)"; 
     "-ATT", Arg.Unit (set_syntax `ATT), " Use AT&T syntax (default is AT&T)"; 
     "-call-conv", Arg.Symbol (["windows"; "linux"], set_cc), " Select calling convention (default depends on host architecture)";
-    "-arch", Arg.Symbol (["x86-64"; "arm-m4"; "riscv"], set_target_arch), " Select target arch (default is x86-64)";
+    "-arch", Arg.Symbol (["x86-64"; "arm-m4"; "armv8a"; "riscv"], set_target_arch), " Select target arch (default is x86-64)";
     "-system", Arg.Symbol (["macosx"; "linux"], set_target_system), " Select target system (default is "^ Config.target_system^")";
     "-stack-zero",
       Arg.Symbol (List.map fst stack_zero_strategies, set_stack_zero_strategy),
--- a/compiler/src/pp_arm_v8a.ml
+++ b/compiler/src/pp_arm_v8a.ml
@@ -1,0 +1,230 @@
+(* Assembly printer for ARMv8-A (AArch64).
+
+   GNU assembler (ELF) syntax. General-purpose registers are printed in
+   their 64-bit X form except for the instruction operands that require the
+   32-bit W form (narrow loads and stores, 32-bit multiplies, zero
+   extensions). Immediate values are printed as nonnegative integers. *)
+
+open Arch_decl
+open Utils
+open PrintASM
+open Asm_utils
+
+(* Architecture imports*)
+open Arm_common
+open Armv8a_decl
+open Armv8a_instr_decl
+
+let arch = armv8a_decl
+
+let imm_pre = "#"
+
+(* We support the following AArch64 memory accesses.
+   Offset addressing:
+     - A base register and an immediate offset (displacement):
+       [<reg>, #<imm>].
+     - A base register and a register offset: [<reg>, <reg>].
+     - A base register and a scaled register offset:
+       [<reg>, <reg>, LSL #<imm>].
+*)
+let pp_reg_address_aux base disp off scal =
+  match (disp, off, scal) with
+  | None, None, None ->
+      Format.asprintf "[%s]" base
+  | Some disp, None, None ->
+      Format.asprintf "[%s, %s%s]" base imm_pre disp
+  | None, Some off, None ->
+      Format.asprintf "[%s, %s]" base off
+  | None, Some off, Some scal ->
+      Format.asprintf "[%s, %s, lsl %s%s]" base off imm_pre scal
+  | _, _, _ ->
+      hierror
+        ~loc:Lnone
+        ~kind:"assembly printing"
+        ~internal:true
+        "the address computation is too complex: an intermediate variable might be needed"
+
+let pp_imm = pp_imm imm_pre
+
+let pp_register = pp_register arch
+
+(* The 32-bit (W) view of a general-purpose register. *)
+let pp_wregister r =
+  match pp_register r with
+  | "xzr" -> "wzr"
+  | "sp" -> "wsp"
+  | s -> "w" ^ String.sub s 1 (String.length s - 1)
+
+let pp_reg_address addr =
+  let addr = parse_reg_address arch addr in
+  pp_reg_address_aux addr.base addr.displacement addr.offset addr.scale
+
+let pp_condt = hash_to_string string_of_condt
+
+(* A64 has no conditional execution: a condition is always a proper operand
+   (CSEL, CSET, ...), declared last in the argument list, so it is printed
+   in place like any other operand. *)
+let pp_asm_arg ?(wform = false) (arg : (register, Arch_utils.empty, Arch_utils.empty, rflag, condt) asm_arg) =
+  match arg with
+  | Condt ct -> pp_condt ct
+  | Imm (ws, w) -> pp_imm (Conv.z_unsigned_of_word ws w)
+  | Reg r -> if wform then pp_wregister r else pp_register r
+  | Regx _ -> .
+  | Addr (Areg ra) -> pp_reg_address ra
+  | Addr (Arip r) -> pp_rip_address r
+  | XReg _ -> .
+
+(* -------------------------------------------------------------------- *)
+
+let pp_shift_kind = hash_to_string string_of_shift_kind
+
+(* Append the optional shift to the last (immediate) operand:
+   [..., <shift> #<amount>]. *)
+let pp_shift (ARMv8A_op (_, opts)) args =
+  match has_shift opts with
+  | None ->
+      args
+  | Some sk ->
+      let sh = pp_shift_kind sk in
+      List.modify_last (Format.asprintf "%s %s" sh) args
+
+(* The last operand of MOVZ/MOVN/MOVK is the halfword shift: print it as
+   [LSL #<sh>], dropping it when it is zero. *)
+let pp_mov_wide_shift args =
+  match List.rev args with
+  | "#0" :: rest -> List.rev rest
+  | sh :: rest -> List.rev (Format.asprintf "lsl %s" sh :: rest)
+  | [] -> []
+
+let pp_mnemonic (ARMv8A_op (mn, _) as op) =
+  let id = instr_desc Armv8a_decl.armv8a_decl Armv8a_instr_decl.armv8a_op_decl (None, op) in
+  let pp = id.id_pp_asm [] in
+  match mn with
+  | UXTW -> "mov" (* [UXTW] is written [MOV <Wd>, <Wn>]. *)
+  | _ -> String.lowercase_ascii pp.pp_aop_name
+
+(* Split an [ADR] instruction to a global symbol into an [ADRP]/[ADD]
+   pair using :lo12: relocations. *)
+let pp_ADR args =
+  match args with
+  | dst :: addr :: _ ->
+      [ Instr ("adrp", [ dst; addr ]);
+        Instr ("add", [ dst; dst; ":lo12:" ^ addr ]) ]
+  | _ -> assert false
+
+module Armv8aTarget : AsmTargetBuilder.AsmTarget with
+  type reg = Armv8a_decl.register
+  and type regx = Arch_utils.empty
+  and type xreg = Arch_utils.empty
+  and type rflag = Arm_common.rflag
+  and type cond = Arm_common.condt
+  and type asm_op = armv8a_asm_op
+= struct
+
+  type reg = Armv8a_decl.register
+  type regx = Arch_utils.empty
+  type xreg = Arch_utils.empty
+  type rflag = Arm_common.rflag
+  type cond = Arm_common.condt
+  type asm_op = armv8a_asm_op
+
+  let headers = []
+
+  let data_segment_header =
+    [
+      Instr (".p2align", ["5"]);
+      Label global_datas_label
+    ]
+
+  (* A64 instructions must be 4-byte aligned. *)
+  let function_directives = [ Instr (".p2align", ["2"]) ]
+
+  (* The return address (X30) is saved on the stack around the body of
+     export functions; SP must stay 16-byte aligned. *)
+  let function_header =
+    [
+      Instr ("str", [ "x30"; "[sp, #-16]!" ])
+    ]
+
+  let function_tail =
+    [
+      Instr ("ldr", [ "x30"; "[sp], #16" ]);
+      Instr ("ret", [])
+    ]
+
+  let pp_instr_r fn i =
+    match i with
+    | ALIGN ->
+        (* A64 instructions are fixed-width: code is always 4-byte
+           aligned, there is nothing to emit. *)
+        []
+
+    | LABEL (_, lbl) ->
+        [ Label (string_of_label fn lbl) ]
+
+    | STORELABEL (dst, lbl) ->
+        [ Instr ("adr", [ pp_register dst; string_of_label fn lbl ]) ]
+
+    | JMP lbl ->
+        [ Instr ("b", [ pp_remote_label lbl ]) ]
+
+    | JMPI arg ->
+        begin match arg with
+        | Reg R30 -> [ Instr ("ret", []) ]
+        | Reg r -> [ Instr ("br", [ pp_register r ]) ]
+        | _ ->
+            hierror ~loc:Lnone ~kind:"assembly printing"
+              "unsupported indirect jump argument"
+        end
+
+    | Jcc (lbl, ct) ->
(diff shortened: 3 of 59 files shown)


tests/success/armv8a/add.jazz
```diff
@@ -0,0 +1,21 @@
+export
```
vbgl: This file looks wrong. Was it added by mistake?

clebreto: A stray duplicate of compiler/tests/success/armv8a/add.jazz indeed. Sorry. I'll remove it in the next commit (1507adcbe)

vbgl: Thanks!

compiler/src/glob_options.ml
```diff
@@ -238,14 +246,20 @@ let options = [
...
       " Select stack zeroization strategy for export functions";
     "-stack-zero-size",
       Arg.Symbol (List.map fst Annot.ws_strings, set_stack_zero_size),
       " Select stack zeroization size for export functions";
+    "-sp-min-align",
+      Arg.Symbol (List.map fst Annot.ws_strings, set_sp_min_align),
+      " Minimal alignment kept by the stack pointer across calls (default \
+       depends on the target architecture: u128, i.e. 16 bytes, on armv8a; \
+       no constraint otherwise). Lowering it below the architectural \
+       requirement may produce faulting code";
```
vbgl: What is the purpose of this option?

compiler/src/arm_arch_full.ml
```diff
@@ -56,5 +57,9 @@ module Arm (Lowering_params : Arm_input) : Arch_full.Core_arch
 
   let callstyle = Arch_full.ByReg { call = Some LR; return = false }
 
+  let sp_min_align = Wsize.U8
```
vbgl: Is this correct?

> The Armv7-M architecture guarantees that stack pointer values are at least 4-byte aligned

Source: Arm v7-M Architecture Reference Manual (version E.e), § B1.5.7)

proofs/arch/arch_decl.v
```diff
@@ -440,6 +440,9 @@ Record instr_desc_t := {
   id_semi_safe  : id_valid -> interp_safe_cond_lty id_tin id_safe id_semi;
 }.
 
+Definition doit := DOIT.
+Definition not_doit := NOT_DOIT.
+
```
vbgl: What is the purpose of these definitions?

compiler/src/glob_options.ml
clebreto: From what I understand, it's only useful in marginal cases: lowering the alignment on a platform where the OS has disabled SP alignment checking (SCTLR_ELx.SA0) in order to reclaim the stack padding, or raising it for experimentation purposes. 

Since it is really hypothetical for me, this can be removed for now, as long as we keep the enforce_sp_min_align which is compulsory for armv8a.

compiler/src/arm_arch_full.ml
clebreto: You're right. It can be an issue. Looking into it. Commit 04af3dfeb adds some tests, which were failing with ```let sp_min_align = Wsize.U8```.

proofs/arch/arch_decl.v
clebreto: A leftover of the pre-squash version of #1520 this branch was based on (that branch introduced the aliases; the squashed commit on main writes DOIT/NOT_DOIT directly). We can drop them. Thanks !
 
 c78e644aa For the removal.

compiler/src/glob_options.ml
vbgl: Considering that this feature seems completely independent from the other changes made in this PR, I suggest to remove it from now, as you proposed.

(Also note that this “enforce-sp-min-align” and the new architecture parameter are fairly independent and could be moved to a separate PR that could be merged quickly.)

clebreto: PR #1546 extracts the sp-min-align.
4917f3e48 on this branch remove the -sp-min-align command-line option, we can add it back later if needed.

#1546 must be merged before this PR.

vbgl:
I think that the directory  `armv8a-hw-semantics` does not belong to this repository. Can you please remove it from this PR? Thanks.

clebreto:
> I think that the directory  `armv8a-hw-semantics` does not belong to this repository. Can you please remove it from this PR? Thanks.

Removed in ac0c7f939.

For context: this harness tested the instruction semantics against "real hardware". It links against the extracted Coq model, evaluates each instruction's id_semi on a battery of inputs, and emits a C program that executes the same instruction forms on an AArch64 host (inline asm, reading NZCV back through mrs) and compares the results — 509 forms, 21442 checks. The correctness proofs are relative to id_semi, so this checks the one link they cannot: model vs "real hardware". It caught several model bugs during development (variable-shift masking, a ROR mis-model, MOVZ/MOVK W-form encodability). Since it needs an AArch64 host, which the CI does not have, it now lives out of tree on the fork: https://github.com/clebreto/jasmin/tree/armv8a-hw-semantics/compiler/tests/armv8a-hw-semantics — still runnable there with make -C compiler check-armv8a-semantics.

If for any reason these additional checks are pointless, please let me know.

vbgl:
I do think that these checks are worthwhile and useful. Thanks for writing and running them. Again, I don’t think that this repo is for know the right place to have this code.

Regarding this PR, I’m not sure whether you want to preserve the git history or not. I think it would help the reviewing process if you could split the change sets in two phases. First the refactoring of existing code that creates the “arm-common” things. Second the addition of new code. Indeed there is currently some issue reported by CI about `JModel_m4.ec` which could be due to the first phase only.

vbgl:
This needs a changelog entry.

I’m not convinced by the new `compiler/doc` directory: why isn’t the documentation part of the manual? Do we need to store that json data inside this source code repository?

clebreto:
> This needs a changelog entry.
> 
> I’m not convinced by the new `compiler/doc` directory: why isn’t the documentation part of the manual? Do we need to store that json data inside this source code repository?

- changelog:  Added in 96cb8f5ab.
- compiler/doc: Agreed. The register-model note is now a page of the manual (compiler/advanced/armv8a_register_model), and the JSON leaves the repository: its content is already carried, instruction by instruction, by the documentation blocks in armv8a_instr_decl.v (each lists its ARM ARM C6.2 section, page, syntax and Operation pseudocode), so the machine-readable extraction form only mattered at development time; it stays available on the fork with the other development artifacts. The compiler/doc directory is gone.

vbgl:
Thanks. Please put the changelog entry in a separate file in the `changes/` directory. More info: https://github.com/jasmin-lang/jasmin/tree/main/changes#unreleased-changelog

clebreto:
> Thanks. Please put the changelog entry in a separate file in the `changes/` directory. More info: https://github.com/jasmin-lang/jasmin/tree/main/changes#unreleased-changelog

Indeed, sorry.

 Done in 7624185ed — moved to changes/01-feature/1541-armv8a-backend.md.

vbgl:
Thanks. CI seems OK. Do you want to clean the git history before merging this into `main`?

clebreto:
> Thanks. CI seems OK. Do you want to clean the git history before merging this into `main`?

Before I rewrite the history into a sequence of self-contained commits: would you rather squash-merge?

vbgl:
Either way seems good to me. I can either squash what is there or keep what you’ll suggest (after review).

clebreto:
> Either way seems good to me. I can either squash what is there or keep what you’ll suggest (after review).

I can force push the refactoring of the history, but to check that the content has not changed you should keep a local copy of that branch, would that work for you ?

vbgl:
You may change the contents of the branch if you like. Please do force push your clean history.

clebreto:
> You may change the contents of the branch if you like. Please do force push your clean history.

It should be cleaner now.

proofs/compiler/armv8a_decl.v
```diff
@@ -0,0 +1,357 @@
...
+
+Definition armv8a_call_conv : calling_convention :=
+  {| callee_saved :=
+      map ARReg [:: R19; R20; R21; R22; R23; R24; R25; R26; R27; R28
+                  ; R29; RSP ]
+   ; callee_saved_not_bool := erefl true
+   ; callee_saved_has_rsp  := erefl true
+   ; call_reg_args  := [:: R0; R1; R2; R3; R4; R5; R6; R7 ]
+   ; call_xreg_args := [::]
+   ; call_reg_ret   := [:: R0; R1 ]
```
vbgl: I’d expect the list of return registers to be longer:

> The first eight registers, r0-r7, are used to pass argument values into a subroutine and to return result values from a function.

(https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#611general-purpose-registers)

proofs/compiler/armv8a_stack_zeroization_proof.v
```diff
@@ -0,0 +1,828 @@
...
+
+#[local]
+Lemma find_instr_skip p fn P Q :
+  is_linear_of p fn (P ++ Q) ->
+  forall scs m vm n,
+  find_instr p (Lstate scs m vm fn (size P + n)) = oseq.onth Q n.
+Proof. by eauto using find_instr_skip'. Qed.
+
+End FIXME.
+
```
vbgl: This is fishy. Who will fix this, and when?

Suggestion:

~~~
--- a/proofs/compiler/armv8a_stack_zeroization_proof.v
+++ b/proofs/compiler/armv8a_stack_zeroization_proof.v
@@ -28,23 +28,6 @@ Import seq_extra.

 Set SsrOldRewriteGoalsOrder.  (* change Set to Unset when porting the file, then remove the line when requiring MathComp >= 2.6 *)

-(* FIXME: We should use the higher-level [eval_lsem] lemmas. *)
-Section FIXME.
-
-Context
-  {asm_op syscall_state : Type}
-  {ep : EstateParams syscall_state}
-  {sip : SemInstrParams asm_op syscall_state}.
-
-#[local]
-Lemma find_instr_skip p fn P Q :
-  is_linear_of p fn (P ++ Q) ->
-  forall scs m vm n,
-  find_instr p (Lstate scs m vm fn (size P + n)) = oseq.onth Q n.
-Proof. by eauto using find_instr_skip'. Qed.
-
-End FIXME.
-
 #[local] Existing Instance withsubword.

 Section STACK_ZEROIZATION.
@@ -429,7 +412,7 @@ Proof using atoI call_conv fn halign hbody hlabel hsmall hstack lbl le_ws_ws_ali
     move: hn; rewrite Z.mul_1_l => ?; subst n.
     exists s3; split.
     + apply: (lsem_n_step_end hsem3).
-      by rewrite /step (find_instr_skip hbody) /= /eval_instr /=
+      by rewrite /step (find_instr_skip' hbody (n := 3)) // /= /eval_instr /=
          /get_var /= hzf3 /= GRing.addrN /ZF_of_word /= /setpc /=
          /lnext_pc /= -addnS.
     by move: hsr3; rewrite Z.sub_diag.
~~~

compiler/src/pp_arm_v8a.ml
```diff
@@ -0,0 +1,228 @@
...
+    [
+      Instr ("str", [ "x30"; "[sp, #-16]!" ])
+    ]
+
+  let function_tail =
+    [
+      Instr ("ldr", [ "x30"; "[sp], #16" ]);
+      Instr ("ret", [])
+    ]
+
```
vbgl, line 154: Why isn’t this handled with the other callee-saved registers in register allocation, linearization & co.?

compiler/src/pp_arm_v8a.ml
```diff
@@ -0,0 +1,228 @@
...
+  let pp_instr_r fn i =
+    match i with
+    | ALIGN ->
+        [ Instr (".p2align", ["2"]) ]
```
vbgl: This looks redundant. Aren’t instructions always aligned in this way?

compiler/src/pp_arm_v8a.ml
```diff
@@ -0,0 +1,228 @@
...
+        [ Instr (iname, [ string_of_label fn lbl ]) ]
+
+    | JAL (R30, lbl) ->
+        [ Instr ("bl", [ pp_remote_label lbl ]) ]
+
+    | CALL _
+    | JAL _ -> assert false
+
+    | POPPC ->
+        [ Instr ("ldr", [ "x30"; "[sp], #16" ]); Instr ("ret", []) ]
```
vbgl, line 191: Again, this handling of x30 is suspicious.

vbgl (commented):
Minor question: why are registers named Rn rather than Xn?

proofs/compiler/armv8a_params_proof.v
```diff
@@ -0,0 +1,1125 @@
...
+
+  rewrite /lip_lloads /armv8a_lloads.
+  rewrite /sem_fopns_args foldM_cat -!/sem_fopns_args.
+  case hall: (all (fun '(_, ofs1) => is_arith_small ofs1) to_save).
+
+  (* All offsets small: LDR directly from [rspi]. *)
+  - have [hsem1 hget1] := lloads_aux_correct armv8a_lload_correct hnin hget hf.
+    rewrite -[X in sem_fopns_args _ X]/(lloads_aux armv8a_lload rspi to_save).
+    rewrite hsem1 bind_ok_estate.
+    have [vm [-> hvm]] := hsp _ hget1.
```
vbgl:
```suggestion
    have [vm -> hvm] := hsp _ hget1.
```

proofs/compiler/armv8a_params_proof.v
```diff
@@ -0,0 +1,1125 @@
...
+  - exact: (hnin_reb _ hnin2).
+  have [hsem1 _] := lloads_aux_correct armv8a_lload_correct
+                      (rspi := vtmp2i)
+                      (to_restore := map (fun '(x1, ofs1) =>
+                                            (x1, (ofs1 - ofs0)%Z)) to_save)
+                      (s := with_vm s vma) hnin2' hgeta2 hf'.
+  rewrite -[X in sem_fopns_args _ X]/(lloads_aux armv8a_lload vtmp2i
+      (map (fun '(x1, ofs1) => (x1, (ofs1 - ofs0)%Z)) to_save)).
+  rewrite hsem1 bind_ok_estate.
+  have [vm [-> hvm]] := hsp _ hget_r'.
```
vbgl:
```suggestion
  have [vm -> hvm] := hsp _ hget_r'.
```

vbgl (commented):
Fix a pair of warnings.

proofs/compiler/armv8a_decl.v
clebreto: Right — the AAPCS64 register table marks r0-r7 as parameter/result registers. We had mirrored the two-register choice of the other backends; extending call_reg_ret to R0..R7 is faithful and only enlarges what is accepted. Will fix.

proofs/compiler/armv8a_stack_zeroization_proof.v
clebreto: I will apply your suggestion (the local wrapper had a single use site). For the record, this section was copied verbatim from arm_stack_zeroization_proof.v, which still carries the same FIXME on main, should we do the same fix there ?

vbgl: Let’s look at existing FIXMEs more generally in a different PR.

vbgl:
Please ignore my comments about x30. Benjamin gave me convincing explanation.

compiler/src/pp_arm_v8a.ml
clebreto: A64 instructions being fixed 4-byte, we can emit nothing instead, with a comment saying why.

proofs/compiler/armv8a_decl.v
clebreto: 917d1748a

proofs/compiler/armv8a_stack_zeroization_proof.v
clebreto: 050eb14eb

proofs/compiler/armv8a_params_proof.v
clebreto: 60795a265

proofs/compiler/armv8a_params_proof.v
clebreto: 60795a265

clebreto:
> Please ignore my comments about x30. Benjamin gave me convincing explanation.

Perfect.

(It could be moved to register allocation/linearization but this would impact all 3 backends, maybe in another PR to have a look at what that would look ?)

compiler/src/pp_arm_v8a.ml
clebreto: c08266ee4

vbgl:
> (It could be moved to register allocation/linearization but this would impact all 3 backends, maybe in another PR to have a look at what that would look ?)

Indeed, there is room for improvement but this is definitely worth a separate PR.

vbgl:
Thanks. Do you want to clean the history again?

clebreto:
> Thanks. Do you want to clean the history again?

It is back to the split with 14 commits.

vbgl (approved):

PR #1542, opened 2026-08-26
linearization: handle some remaining assignments

vbgl:
Linearization parameters already provide methods to handle moves, loads, and stores that appear as assignments (as opposed to target-specific `opn`).

Prior to this PR, linearization fails when any assignment remains. This PR recognizes some of these assignments and translate them using the appropriate parameters.

This is currently useless (dead code) but it makes it possible for the passes occurring after lowering and before linearization to introduce load / store / moves in a generic way (architecture dependent). I intend to re-implement the (currently experimental & trusted) automatic spilling pass using this facility.


eponier (approved):
Interesting!

PR #1544, opened 2026-08-26
riscv: DIV, DIVU, REM and REMU are neither constant-time nor DOIT

### LINKED ISSUE #1013: RISC-V: CT and doit are wrong

clebreto:
Fixes #1013.

RISC-V has no run-time flag akin to Intel DOIT or ARM DIT; the reference for data operand independent timing is the ratified **Zkt** extension ("Data-Independent Execution Latency Subset", [chapter Zkt of the Scalar Cryptography spec](https://github.com/riscv/riscv-crypto/blob/main/doc/scalar/riscv-crypto-scalar-zkt.adoc)). Its safe list covers, for the RV32IM subset modelled by Jasmin, all the base ALU instructions and `mul`/`mulh`/`mulhu`/`mulhsu`, and deliberately excludes division and remainder.

Changes:

- `riscv_instr_decl.v`: the `DOIT`/`NOT_DOIT` constructors are now passed explicitly to the descriptor helpers at every call site (as done for the other architectures in #1520); `DIV`, `DIVU`, `REM` and `REMU` become `NOT_DOIT`. A documentation block records the classification rationale against the Zkt list, including:
  - pseudo-instructions, classified by their canonical expansions (`MV` → `addi`, `NOT` → `xori`, `NEG` → `sub`, `LI` → `lui`+`addi`, `LA` → `auipc`+`addi`), all Zkt-listed;
  - loads and stores, which Zkt does not cover: the address is already leaked by the Jasmin leakage model, and independence from the transferred *values* is assumed, as on the other architectures (Intel's DOIT list contains `MOV`, ARM's DIT covers `LDR`/`STR`), so they stay `DOIT`;
  - the HINT caveat raised in the issue: Zkt exempts HINT encodings (typically `rd = x0`) from the latency requirement, but Jasmin can never emit one since `x0` is not part of the declared register file.
- `riscv_arch_full.ml`: `is_ct_asm_op` returns `false` for `DIV`/`DIVU`/`REM`/`REMU` (the `FIXME RISCV` is resolved), matching the x86/ARM convention of assuming everything but division to be constant-time.
- Tests: new `CCT-riscv` and `CCT-DOIT-riscv` categories with six tests: secret `#DIV`/`#REM` rejected by the plain CT checker, secret `/` and `%` rejected under `--doit` (both were wrongly accepted before this change), public division accepted, and secret `mul`/`add`/`xor`/shift/spill accepted under `--doit`.
- The Zkt reference is added next to the Intel and ARM ones in `docs/source/tools/ct.md`.

Validation: full proofs build, extraction and compiler build are green; `CCT`, `CCT-riscv`, `CCT-DOIT`, `CCT-DOIT-riscv`, `SCT` (43/43) and the `risc-v` compilation category (106/106) all pass.

### DIFF
--- a/changes/02-bugfix/1544-riscv-ct-doit.md
+++ b/changes/02-bugfix/1544-riscv-ct-doit.md
@@ -1,0 +1,7 @@
+- On RISC-V, the `DIV`, `DIVU`, `REM` and `REMU` instructions are no longer
+  considered constant-time nor DOIT: they have data-dependent latency on
+  typical implementations and are excluded from the Zkt ("Data-Independent
+  Execution Latency") safe list of the RISC-V specification; every other
+  instruction keeps its classification, now documented against the Zkt list
+  ([PR 1544](https://github.com/jasmin-lang/jasmin/pull/1544);
+  fixes [#1013](https://github.com/jasmin-lang/jasmin/issues/1013)).
--- a/compiler/CCT/fail/doit/riscv/div.jazz
+++ b/compiler/CCT/fail/doit/riscv/div.jazz
@@ -1,0 +1,11 @@
+/* Unsigned division lowers to DIVU, which is not in the Zkt safe list,
+   hence not DOIT. */
+#[ct="secret -> secret"]
+export fn div(reg u32 x) -> reg u32 {
+  reg u32 d q;
+  x = x;
+  d = 3;
+  q = x / d;
+  q = q;
+  return q;
+}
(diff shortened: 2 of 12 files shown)


bgregoir (approved):
Thank a lot

PR #1546, opened 2026-08-28
Stack allocation: enforce a per-architecture minimal stack-pointer alignment

clebreto:
Split out of #1541, as suggested in review: the mechanism is independent of the ARMv8-A backend.

- **First commit**: add an `sp_min_align` architecture parameter bounding the alignment the stack pointer keeps across calls; stack allocation raises the frame alignment of every function that actually uses the stack up to it (frame sizes are rounded up to the frame alignment, so this bounds the SP alignment itself). All architectures declare `U8`, so no generated code changes.
- **Second commit**: set it to `U32` on arm-m4. Armv7-M guarantees that stack pointer values are at least 4-byte aligned — writes to SP force bits [1:0] to zero (Arm v7-M Architecture Reference Manual, B1.5.7) — so a frame that is only 1- or 2-byte aligned lets the export prologue write a non-4-byte-aligned SP (`BIC r12, r12, #1` before `MOV sp, r12`), which the hardware silently rounds, shifting every subsequent SP-relative access. The new `tests/success/arm-m4/unaligned_stackalign.jazz` asserts the computed alignment with `#[stackalign=u32]` and fails without this commit; the `instack` function of `tests/success/common/unaligned.jazz` moves to per-architecture copies since the expected alignment is now architecture-dependent.

No command-line override: the per-architecture default is the only value with a real use case.

The ARMv8-A backend (#1541) relies on this mechanism with a 16-byte floor (AArch64 SP alignment checking); once this lands, the corresponding hunks disappear from that PR.

### DIFF
--- a/compiler/src/stackAlloc.ml
+++ b/compiler/src/stackAlloc.ml
@@ -106,6 +106,19 @@ module StackAlloc (Arch: Arch_full.Arch) = struct
 
 module Regalloc = Regalloc (Arch)
 
+(* Keep the stack pointer [Arch.sp_min_align]-aligned across calls: frame
+   sizes are rounded up to the frame alignment
+   ([stack_frame_allocation_size]), so raise the alignment of every function
+   that actually uses the stack. Functions with no stack footprint at all
+   keep their alignment (an export function may only use SavedStackNone when
+   its alignment is U8). *)
+let enforce_sp_min_align ~stack_size ~extra_size ~max_stk align =
+  if Z.equal max_stk Z.zero
+     && Z.equal stack_size Z.zero
+     && extra_size = 0
+  then align
+  else Utils0.cmp_max wsize_cmp align Arch.sp_min_align
+
 let memory_analysis pp_sr pp_err ~debug callee_saved_strategy up =
   if debug then Format.eprintf "START memory analysis@.";
   let p = Conv.prog_of_cuprog up in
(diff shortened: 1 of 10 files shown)


compiler/src/stackAlloc.ml
```diff
@@ -106,6 +106,20 @@ module StackAlloc (Arch: Arch_full.Arch) = struct
...
+   that actually uses the stack. Functions with no stack footprint at all
+   keep their alignment (an export function may only use SavedStackNone when
+   its alignment is U8). *)
+let enforce_sp_min_align ~stack_size ~extra_size ~max_stk align =
+  if Z.equal max_stk Z.zero
+     && Z.equal stack_size Z.zero
+     && extra_size = 0
+  then align
+  else if wsize_lt align Arch.sp_min_align then Arch.sp_min_align
+  else align
```
vbgl: This conditional seems more complex than it should: there are two possible results (`align` and `Arch.sp_min_align`) so a single `if` (with two branches instead of three) should be enough.

You may also prefer `cmp_max` to hide the nested `if` behind something clearer.

vbgl (approved):
LGTM. Why is this PR marked as “draft”?

clebreto:
> LGTM. Why is this PR marked as “draft”?
I wanted to have a last look at it with the github diffs, sorry. I remove the draft now.

PR #1549, opened 2026-09-03
Extract global

bgregoir:
# Description

Allows to have annotations on global variables.
Use it to select how global variables are extracted (abbrev or op).

# Checklist

### DIFF
--- a/changes/01-feature/1549-extract-global.md
+++ b/changes/01-feature/1549-extract-global.md
@@ -1,0 +1,8 @@
+- Global variables can now be extracted to EasyCrypt as operators
+  instead of abbreviations, and their machine words can be printed as
+  signed or unsigned integers. The default is selected by the
+  `--global-model` command line argument of `jasmin2ec` and can be
+  overridden for each global variable by the `#[abbrev]`, `#[op]`, and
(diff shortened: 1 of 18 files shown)


vbgl (commented):
Looks good.

Why is the command-line syntax (`op[opaque]`) different than the annotation syntax (`op = opaque`)? This seems confusing.

bgregoir:
so you propose `--global-model = op = "opaque smt_opaque"`, and `--global-model = op = opaque` ?
If yes, I not fane of the "= op =", but no strong oposition. Remark that op[opaque smt_opaque] is the ec syntax.
But I can not have this syntax in jasmin annotation.

vbgl:
I’d imagined: `--global-model "op = opaque smt_opaque"`. I think that with cmdliner the equal sign between argument name and value is optional.

vbgl:
You may also have exactly the same syntax writing: `--global-model op="opaque smt"`.

bgregoir:
I will try.

bgregoir:
I think it is done, I have also added another option "sign". This last one can really change the time easycrypt needs to load files with large table inside.

vbgl (approved):

PR #1550, opened 2026-09-04
Save the return address of export functions in the stack frame (arm64, risc-v)

clebreto:
**This PR is a proposal**, following up on the review discussion of #1541 ([this comment](https://github.com/jasmin-lang/jasmin/pull/1541) asked why the saving of X30 around export functions is done by the assembly printer rather than by register allocation / linearization). It is **based on #1541 and can only land after it**; until #1541 is merged, the diff shown here includes its commits — only the last four commits belong to this PR.

## What it does

On architectures where the caller passes the return address of an export function in a register (X30 on arm64, `ra` on RISC-V), that register is now saved and restored through the same verified mechanism as killed callee-saved registers — a slot in the stack frame, written and read by the to-save machinery of linearization — instead of by unverified glue emitted by the assembly printers around the function body. The printers now only emit the final return jump (`ret`), which puts them in the exact same position as the x86 printer and its `ret`.

- `calling_convention` gains a `call_reg_ra : option reg_t` field recording this register, and `one_varmap_info` derives the corresponding variable set.
- The one-varmap checker (`merge_varmaps`) enforces the new invariant for export functions: if the body kills the return-address register, it must be in `sf_to_save`. The write-set `W'` already accounts for that register at every call (`ra_vm`) and system call (`syscall_kill`), so the check integrates with the existing analysis; a single line of `it_merge_varmaps_proof` needed adaptation.
- The register-allocation oracle spills the return-address register like a killed callee-saved register, and it is removed from the general allocatable pool on these architectures: it now carries a live value across the whole body of export functions, and using it as a scratch register would needlessly consume a spill slot (it was only free under the old save-everything glue).

## Consequences

- Export functions that do not kill the return-address register (e.g. leaf functions) no longer touch the stack at all to save it; e.g. on arm64 an exported increment is now just `add w0, w0, #1` / `ret`.
- The `callee_saved = n` annotation must account for the return-address register when the function kills it (documented in the manual).

## Scope and limitations

- **arm-m4 keeps the printer glue** (`push {lr}` / `pop {pc}`). There, LR is also the linearization scratch register (`lip_tmp2`): the prologue of an export function kills it before any store into the frame is possible, and this is structural on T32 (SP cannot be aligned in place, and R0-R3 carry arguments), so `call_reg_ra` is `None` for arm-m4, with a comment explaining why.
- The final return jump itself remains a single unverified printer instruction on all architectures (as `ret` already is on x86): the semantic contract of export calls (`lsem_exportcall` / `asmsem_exportcall`) ends when the program counter reaches the end of the body, and modelling the jump back to an external caller would require changing those contracts. This is left as future work; this PR reduces the unverified surface to that single instruction and makes the restoration of the return-address register part of the checked, verified pipeline.

## Validation

- Coq proofs, extraction and OCaml build all green.
- Full test suite: 1066/1066 across all categories (x86-64 ATT/Intel/nolea, arm-m4, armv8a, risc-v, all stack-zeroization variants, CCT/CCT-DOIT/SCT); `dune runtest` green (one legitimate golden update: a RISC-V temporary formerly allocated to `ra` now goes to `x5`).
- Native execution of exported leaf/calling functions on AArch64.
- The formosa-mldsa ML-DSA AArch64 port (parameter sets 44, 65 and 87) passes its complete KAT suites when compiled with this branch.

### DIFF
--- a/changes/01-feature/1541-armv8a-backend.md
+++ b/changes/01-feature/1541-armv8a-backend.md
@@ -1,0 +1,3 @@
+- Add an ARMv8-A (AArch64) backend: new target architecture `-arch armv8a`,
+  modeled at the full 64-bit register width, with lowering and correctness
+  proofs ([PR #1541](https://github.com/jasmin-lang/jasmin/pull/1541)).
--- a/changes/03-other/1550-export-return-address-save.md
+++ b/changes/03-other/1550-export-return-address-save.md
@@ -1,0 +1,12 @@
+- On architectures where the caller passes the return address in a register
+  and that register is available (ARM64, RISC-V), export functions now save
+  and restore it through the same verified mechanism as callee-saved registers
+  (a slot in the stack frame), instead of unverified glue emitted by the
+  assembly printer around the function body. The one-varmap checker enforces
+  that an export function whose body kills the return-address register saves
+  it. Export functions that do not kill it (e.g. leaf functions) no longer
+  touch the stack at all to save it. Note that the `callee_saved = n`
+  annotation must now account for the return-address register when the
+  function kills it. On ARM (Cortex-M4), LR is used as a scratch register by
+  the compiler itself, so it is still saved around the body by the printer
+  ([PR #1550](https://github.com/jasmin-lang/jasmin/pull/1550)).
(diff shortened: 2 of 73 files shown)


bgregoir (commented):
I don't understand why this patch is applied only to risc-v and armv8 but not to armv7

PR #1554, opened 2026-09-09
Add jobs param to opam build config

rudynicolop:
# Description

Add a "jobs" flag to the `opam` file's `build` section. I believe this will help make installing Jasmin with the Rocq dependencies via `opam pin add` a little quicker.


eponier (approved):
It seems that lots of projects distributed with opam do that. Thanks for your contribution!

PR #1558, opened 2026-09-09
Safety assert withcatch

bgregoir:
# Description

This introduce the necessary change in the semantic to provide safety.v and safety_proof.v.
For the moment it is not used in the other part of the dev, but it is a first part for the intergration of the safety_assert2 branch.
I should fix the authors of the commit before final integration. 

# Checklist


vbgl:
There is something fishy going on in the `coq-master` CI job: extraction fails.

~~~
Error:
The informative inductive type prod has a Prop instance.
This happens when a sort-polymorphic singleton inductive type
has logical parameters, such as (I,I) : (True * True) : Prop.
Extraction cannot handle this situation yet.
~~~

eponier:
> There is something fishy going on in the `coq-master` CI job: extraction fails.
> 
> ```
> Error:
> The informative inductive type prod has a Prop instance.
> This happens when a sort-polymorphic singleton inductive type
> has logical parameters, such as (I,I) : (True * True) : Prop.
> Extraction cannot handle this situation yet.
> ```
Indeed strange, is mathcomp striking back?

bgregoir:
Sorry for the noise this is still a draft version. I try to be able to remove the "catch" operator from the semantics. I think I am close now. To be able to do it. I try to do so in a way that will allows us to review a PR with big change without need to read all the modification in the semantic

eponier:
I fixed the problem with job coq-master. It was indeed because of mathcomp. We should take care of not using ring-related definitions in the part that gets extracted.

The patch also includes two `Set SsrOldRewriteGoalsOrder.`, because some files that didn't need them before now do.

PR #1559, opened 2026-09-10
docs: Include `bool remain unallocated` into FAQ

iccuevas:
# Description

Added description and error triggering code example of  `register allocation: variables { b.222 } remain unallocated` into the docs.

### DIFF
--- a/docs/source/misc/faq.md
+++ b/docs/source/misc/faq.md
@@ -14,6 +14,34 @@ Which function? With the command-line flag `-debug` on, the compiler will output
 There are two possible kinds of fix to this kind of issue: either spill some registers to ensure there is at least one register that is free at *all* call sites; or change the calling-convention of the function so that return address is passed on the stack
 (this can be achieved through the `#returnaddress=stack` annotation before the function declaration).
 
+---
+
+> register allocation: variables { b.222 } remain unallocated
+
+triggered by a program such as:
+
+~~~
+export fn constant_bool_eval(reg u64 y) -> reg u64 {
+  reg u64 x = 0;
+  inline int i = 5;
+  reg bool b;
+
+  b = i != 0;
+  x = y if b;
+  return x;
+}
+~~~
+
+`i` is an inline variable, so `i != 0` is a comparison between two compile-time constants. There is no machine instruction that compares an immediate against an immediate and writes the result into a flag register, so the compiler has nothing to allocate `b` to.
+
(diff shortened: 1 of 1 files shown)


vbgl:
Do you have an example not involving a constant boolean?

Here is a simpler example that is more to the point:

~~~
export fn example() {
  #[keep] reg bool b = true;
}
~~~

iccuevas:
Hi Vincent, sorry for the delayed response.

I don't have such example, I came across this error message when coding and wanted to add to the docs.

Do you have any suggestions to have a complete (and correct) documentation of this compilation error?

vbgl:
The current implementation of register allocation does not allocate boolean variables: it relies on architectural constraints (e.g, the value of an “equal” condition can only be stored in the “zero” flag) to completely solve the allocation problem and checks that this forced solution is consistent with liveness constraints. That is to say, if some variable is free, the register allocation does not do any arbitrary choice such as: “let’s assign this variable to the carry flag”. In case there are not enough constraints, some variables remain non-allocated, and an error message is produced.

There are two known programming errors that lead to this compile-time error.

1. Non-initialized variable

When a variable only occurs in conditions (thus in particular is never assigned), it may be unconstrained. For instance:

~~~
export fn uninit(reg u32 x) -> reg u32 {
  reg bool b; // ← Not initialized!
  if b { x += 1; }
  return x;
}
~~~

In any case, programmers should pay attention to the safety of their programs.

2. Constant booleans

The compiler does not generate code to compute constant boolean values. If you need to put a constant value in some flag, please tell the compiler how to realize it (compare a value to itself, add zero to some value and extract the carry, use the `#STC` instruction on x86, etc.). Here is an example program triggering the problem:

~~~
export fn constant(reg u32 x) -> reg u32 {
  reg bool b = true; // ← Constant value: how should it be computed?
  if b { x += 1; }
  return x;
}
~~~

iccuevas:
Thank you for the detailed response :)
I've updated the docs following your comment

vbgl (approved):

vbgl:
Note that this error could be removed or changed into a warning and let the compiler fail later. In the examples given above we would get:

~~~
compilation error in function uninit:
asmgen: Invalid name for rflag (check initialization?) : b
~~~

and

~~~
compilation error in function constant:
linearization: The following assignment remains:
b = true
Is there an instruction in the target architecture that can implement it?
More information may be found online: https://jasmin-lang.readthedocs.io/en/stable/misc/faq.html#linearization
~~~

PR #1563, opened 2026-09-11
Fix a few bugs in EC extraction

### LINKED ISSUE #1555: Ill-typed extraction of syscalls with warray
Extraction of the following program with `--array-model=warray` produces an ill-typed EasyCrypt file.

~~~
export fn main() -> reg u64 {
  stack u32[2] s;
  s = #randombytes(s);
  reg u64 r = s[:u64 0];
  return r;
}
~~~

Here are the relevant parts of the output.

~~~
module type Syscall_t = {
  proc randombytes_8 (_:W8.t Array8.t) : W8.t Array8.t
}.

module M(SC:Syscall_t) = {
  proc main () : W64.t = {
    var r:W64.t;
    var s:W32.t Array2.t;
    s <- witness;
    s <@ SC.randombytes_8 (s);
    r <- (ArrayAccessCastW64_2W32.get_cast s 0);
    return r;
  }
}.
~~~

The random sampling procedure is declared with byte-arrays but used with 32-bit word-arrays.

### LINKED ISSUE #1557: Extraction generates incomplete ArrayWords files
When extracting the program below using `--array-model=warray`, a file `ArrayWords2W32.ec` is generated with missing `require` directives.

Jasmin source:

~~~
export fn main(reg u64 x) -> reg u64 {
  stack u32[2] s;
  s[:u64 0] = x;
  x = s[:u64 0];
  return x;
}
~~~

ArrayWords2W32.ec:

~~~
require import Int.

require import Array2 WArray8.

clone export ArrayWords as ArrayWords2W32  with op sizeW <- 4, op sizeA <- 2,
                                                theory Word <= W32,
                                                theory ArrayN <= Array2,
                                                theory WArrayN <= WArray8.
~~~

eponier:
# Description

This PR fixes two reported bugs (#1555 and #1557), plus a missing insertion of casts in returns.
EDIT: it also fixes a type issue in the presence of global arrays of length 0 in the Jasmin source.

Fixes #1555
Fixes #1557

I added unit tests in `extraction-unit-tests` but the files in this folder are tested only with the default array model `barray`, so the fixes about model `warray` are actually not tested. EDIT: postponed to https://github.com/jasmin-lang/jasmin/issues/1573

The chosen fix for `randombytes` may feel a bit strange in the case of model `barray`, since the extraction may produce two functions `randombytes_2W32` and `randombytes_1W64` that are actually the same. EDIT: fixed

# Checklist

### DIFF
--- a/changes/02-bugfix/1563-fix-ec-extraction.md
+++ b/changes/02-bugfix/1563-fix-ec-extraction.md
@@ -1,0 +1,2 @@
+- Various fixes to the EasyCrypt extraction
+  ([PR 1563](https://github.com/jasmin-lang/jasmin/pull/1563)).
--- a/compiler/examples/extraction-unit-tests/bug_1555.jazz
+++ b/compiler/examples/extraction-unit-tests/bug_1555.jazz
@@ -1,0 +1,6 @@
+export fn main() -> reg u64 {
+  stack u32[2] s;
+  s = #randombytes(s);
+  reg u64 r = s[:u64 0];
+  return r;
+}
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -2016,8 +2016,9 @@ struct
           |> List.map (fun x -> ESasgn ([LvIdent [ec_vars env x]], ec_ident "witness"))
       in
       let ret =
-          let ec_var x = ec_vari env (L.unloc x) in
-          match ec_leak_ret env (List.map ec_var f.f_ret) with
+          let ret = List.map (fun x -> Pvar (gkvar x)) f.f_ret in
+          let ret = List.map (toec_cast env) (List.combine f.f_tyout ret) in
+          match ec_leak_ret env ret with
           | [x] -> ESreturn x
           | xs -> ESreturn (Etuple xs)
       in
--- a/compiler/src/toEC.ml
+++ b/compiler/src/toEC.ml
@@ -83,6 +83,13 @@ let gsign_of_annot annot =
 
 let ws2bytes ws = (int_of_ws ws) / 8
 
+module Ws = struct
+  type t = wsize
+  let compare = compare
+end
+
+module WsInt = Set.Make2(Ws)(BatInt)
+module Swi = WsInt.Product
 
 (* ------------------------------------------------------------------- *)
 (* Array theories in eclib *)
@@ -442,7 +449,7 @@ module type EnvT = sig
   val pd: t -> Wsize.wsize
   val msfsz: t -> Wsize.wsize
   val arch: t -> architecture
-  val randombytes: t -> int list
+  val randombytes: t -> (wsize * int) list
   val set_fun: t -> ('a, 'b) func -> t
   val add_Array: t -> int -> unit
   val add_WArray: t -> int -> unit
@@ -453,7 +460,7 @@ module type EnvT = sig
   val add_SubArrayDirect: t -> int -> int -> int -> unit
   val add_SubArrayCast: t -> int -> int -> int -> int -> unit
   val add_ArrayAccessCast: t -> int -> int -> int -> unit
-  val add_randombytes: t -> int -> unit
+  val add_randombytes: t -> wsize -> int -> unit
   val add_jarray: t -> Wsize.wsize -> int -> unit
   val empty: architecture -> Wsize.wsize -> Wsize.wsize -> Sarraytheory.t ref -> t
   val create_name: t -> string -> string
@@ -495,7 +502,7 @@ module Env: EnvT = struct
         *)
       auxv: string BatVect.t Mpty.t ref;
       mutable count: int Mpty.t;
-      randombytes: Sint.t ref;
+      randombytes: Swi.t ref;
     }
 
   let vars env = env.vars
@@ -506,7 +513,7 @@ module Env: EnvT = struct
 
   let arch env = env.arch
 
-  let randombytes env = Sint.elements !(env.randombytes)
+  let randombytes env = Swi.elements !(env.randombytes)
 
   let array_theories env = !(env.array_theories)
 
@@ -548,7 +555,7 @@ module Env: EnvT = struct
     add_ArrayWords env sizewb sizeb;
     env.array_theories := Sarraytheory.add (ArrayAccessCast {sizews; sizewb; sizeb}) !(env.array_theories)
 
-  let add_randombytes env n = env.randombytes := Sint.add n !(env.randombytes)
+  let add_randombytes env ws n = env.randombytes := Swi.add (ws, n) !(env.randombytes)
 
   let add_jarray env ws n =
     let ats = Sarraytheory.add (Array n) !(env.array_theories) in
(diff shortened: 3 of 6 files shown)


bgregoir:
Do you think there is a way to fix "The chosen fix for randombytes may feel a bit strange in the case of model barray, since the extraction may produce two functions randombytes_2W32 and randombytes_1W64 that are actually the same."? Maybe by delegating how to do this part to the used array model.

compiler/src/toEC.ml
```diff
@@ -2014,8 +2016,9 @@ struct
           |> List.map (fun x -> ESasgn ([LvIdent [ec_vars env x]], ec_ident "witness"))
       in
       let ret =
-          let ec_var x = ec_vari env (L.unloc x) in
-          match ec_leak_ret env (List.map ec_var f.f_ret) with
+          let ret = List.map (fun x -> Pvar (gkvar x)) f.f_ret in
+          let ret = List.map (toec_cast env) (List.combine f.f_tyout ret) in
```
bgregoir, line 2020: This two List.map cannot be merged into a single one? It is not very important, upto you.

vbgl:
Regarding the fact that the `warray` mode is not tested, it would be nice to either extend the test suite or deprecate that mode. The fact that I encountered two fatal bugs as soon as I tried to use this mode (as opposed to `old`) suggests that it is not wildly used…

I worked around the issues by using the `barray` mode and the migration was essentially a cleaning.

eponier:
> Do you think there is a way to fix "The chosen fix for randombytes may feel a bit strange in the case of model barray, since the extraction may produce two functions randombytes_2W32 and randombytes_1W64 that are actually the same."? Maybe by delegating how to do this part to the used array model.

I pushed a new commit with a patch for this. I added a function `randombytes_suffix` that has to be implemented by each model, and that normalizes array lengths in model `barray`, thus producing only one implementation per array length. This works, but I'm not completely satisfied by this patch, because the knowledge that there can be only one implementation per length in model `barray` is scattered in different parts of the file (meaning that if some change is made to the file, one has to be cautious to update consistently the related parts, which leads to hard-to-maintain code in the long run). But I cannot think of a simpler patch.

eponier:
> Regarding the fact that the `warray` mode is not tested, it would be nice to either extend the test suite or deprecate that mode. The fact that I encountered two fatal bugs as soon as I tried to use this mode (as opposed to `old`) suggests that it is not wildly used…

I don't know how to test this without making EC jobs longer, and those are already the longest jobs...

vbgl:
> I don't know how to test this without making EC jobs longer, and those are already the longest jobs...

The `check-proofs` jobs are not that bad.

compiler/src/toEC.ml
eponier: I wrote it this way to copy the `List.map (toec_cast env)` done in other places.

eponier:
> > I don't know how to test this without making EC jobs longer, and those are already the longest jobs...
> 
> The `check-proofs` jobs are not that bad.

What do you think is better? Check systematically each file in the 3 models, or only specific files manipulating arrays?

eponier:
Out of curiosity, I created two pipelines to test models [old](https://gitlab.com/jasmin-lang/jasmin/-/pipelines/2851006162) and [warray](https://gitlab.com/jasmin-lang/jasmin/-/pipelines/2850996269) (the `test-extract-to-ec` jobs).

eponier:
> Out of curiosity, I created two pipelines to test models [old](https://gitlab.com/jasmin-lang/jasmin/-/pipelines/2851006162) and [warray](https://gitlab.com/jasmin-lang/jasmin/-/pipelines/2850996269) (the `test-extract-to-ec` jobs).

This found only one bug, when we define an array of globals of length 0. I pushed a commit fixing that.

compiler/src/toEC.ml
```diff
@@ -1070,14 +1071,18 @@ let toec_ty onarray env ty = match ty with
...
   Format.sprintf "%s.t %s.t" (fmt_Wsz ws) (ec_Array env n)
 
-let of_list_dfl env _ws n =
-  Eapp (Eident [ec_Array env n; "of_list"], [ec_ident "witness"])
+let of_list_dfl env ws n =
+  let witness = Format.asprintf "witness<:%s.t>" (fmt_Wsz ws) in
+  Eapp (Eident [ec_Array env n; "of_list"], [ec_ident witness])
+
+let randombytes_suffix_dfl ws n =
+  Format.asprintf "%a%s" pp_length n (fmt_Wsz ws)
```
bgregoir: I suspect that the bug was not discover, not because the models are not used but because randombyte was never used with argument/destination different from byte arrays. Since you now have something to print the randombyte_suffix. Can you use a dfl suffix compatible with the original one for the case (ws = u8).  I hope it will allows to not break existing dev.

compiler/examples/extraction-unit-tests/bug_1555.jazz
```diff
@@ -0,0 +1,7 @@
+export fn main() -> reg u64 {
+  stack u32[2] s;
+  s = #randombytes(s);
+  s[u64 0:1] = #randombytes(s[u64 0:1]);
```
vbgl:
```suggestion
  s[:u64 0:1] = #randombytes(s[:u64 0:1]);
```

The syntax without colon is deprecated.

eponier: Indeed, thanks!

eponier: (correctly reported by `jasminc`, but not `jasmin2ec`, is this normal?)

vbgl: Don’t know. I let you investigate.

eponier: Oh you have to use `--warn` to get warnings, by default they are turned off. So nothing unexpected. (no color for the warning though, but this is a different problem)

compiler/src/toEC.ml
eponier: I changed `randombytes_suffix_dfl` in the `U8` case.

compiler/examples/extraction-unit-tests/bug_1555.jazz
eponier: done!

eponier:
I think this is ready. I postponed testing other array models to [an issue](https://github.com/jasmin-lang/jasmin/issues/1573).

changes/02-bugfix/1563-fix-ec-extraction.md
```diff
@@ -0,0 +1,2 @@
+- Various fixes to the EasyCrypt extraction
+  ([PR 1563](https://github.com/jasmin-lang/jasmin/pull/1563)).
```
bgregoir: Maybe, I have a bad understanding of the interest of the changelog. For me the main interest is for users. I don't know if they want to read the PR to know what has been fixed. So maybe a little bit more of informations will be good.

eponier: I pushed a better changelog.

eponier:
(One last concern I have is that in `extraction-unit-tests`, the policy until now was to add a statement in `proofs.ec` it seems, is it ok to add other files and not use them in `proofs.ec`?)

bgregoir (approved):

PR #1564, opened 2026-09-12
Total semantics of the expression operators

bgregoir:
# Description
The goal of this PR is to prepare some changes for the integration of the safety checker. 

The typed semantics of the operators (sem_sop1_typed, sem_sop2_typed,
sem_opN_typed) is now built from a total semantics (sem_op_total.v) and a
list of safety conditions on the arguments (op_semi.v): the operator
succeeds iff its conditions hold, and then computes its total version.
The conditions are terms of a small language (acond, in sopn_semi.v),
evaluated with the total semantics, so that they can later be turned into
assertions.

The first commit is a move. The second commit adds the total semantics, the conditions and the proof
that the guarded total semantics equals the existing one, for every
operator. The lastcommit makes it the definition and adapts the
downstream proofs (mostly the lowering proofs, where a total operator now
unfolds directly to its result).

Fixes # (issue) 

# Checklist


proofs/compiler/constant_prop.v
```diff
@@ -283,7 +283,7 @@ Definition s_op2 o e1 e2 :=
   | _       => ssem_sop2 o e1 e2
   end.
 
-Definition app_sopn := app_sopn of_expr.
+Definition app_sopn := sem_type.app_sopn of_expr.
```
eponier, line 286: Is this change intended?

proofs/lang/sem_op_typed.v
```diff
@@ -1,223 +1,268 @@
...
-  | Ovsub ve ws     => mk_sem_sop2 (sem_vsub ve)
-  | Ovmul ve ws     => mk_sem_sop2 (sem_vmul ve)
-  | Ovlsr ve ws     => mk_sem_sop2 (sem_vshr ve)
-  | Ovlsl ve ws     => mk_sem_sop2 (sem_vshl ve)
-  | Ovasr ve ws     => mk_sem_sop2 (sem_vsar ve)
-
-  | Owi2 s sz o => sem_wiop2_typed s sz o
-  end.
+(* Sharper form: under the conditions the semantics is exactly the total
+   one. *)
```
eponier, line 131: Do we need the version with an exists, if we have this stronger version?

proofs/lang/sem_op_typed.v
```diff
@@ -1,223 +1,268 @@
...
+have htr : mapM2 ErrType truncate_val [:: eval_atype (type_of_op1 o).1] [:: v1]
+             = ok [:: to_val x1].
++ by rewrite /= /truncate_val hof.
+have {}hall : all (acond_b [:: to_val x1]) (op1_safe o).
++ by rewrite (all_acond_b_truncate (op1_safe_ok o) htr).
+by rewrite /sem_sop1_typed mk_sem_op1E (check_safe_ok ErrArith hall).
+Qed.
 
-Arguments sem_sop2_typed : clear implicits.
+Lemma sem_sop2_typed_safeE (o : sop2) (v1 v2 : value) x1 x2 :
```
eponier, line 146: idem

proofs/lang/sem_op_typed.v
```diff
@@ -1,223 +1,268 @@
...
+          (tin := [:: eval_atype (type_of_op2 o).1.1; eval_atype (type_of_op2 o).1.2])
+          (t := eval_atype (type_of_op2 o).2) ErrArith (sem_sop2_total o) x1 x2).
+Qed.
 
-Definition sem_wiop2_typed (sign : signedness) (sz : wsize) ( o : wiop2) :
-  let t := type_of_wiop2 sz o in
-  let t := (eval_atype t.1.1, eval_atype t.1.2, eval_atype t.2) in
-  sem_t t.1.1 → sem_t t.1.2 → exec (sem_t t.2) :=
-  match o with
+Lemma sem_sop1_typed_total o :
```
eponier, line 83: Do we need this, we have `sem_sop1_typed_totalE` that is just stronger.

proofs/lang/sem_op_typed.v
```diff
@@ -1,223 +1,268 @@
...
+  op1_total o -> forall x, exists y, sem_sop1_typed o x = ok y.
+Proof. by move=> h x; exists (sem_sop1_total o x); apply: sem_sop1_typed_totalE. Qed.
 
-  | WIadd => @mk_sem_wiop2 sign sz Z.add
-  | WImul => @mk_sem_wiop2 sign sz Z.mul
-  | WIsub => @mk_sem_wiop2 sign sz Z.sub
-  | WIdiv => @mk_sem_divmod sign sz (signed wdiv wdivi sign)
-  | WImod => @mk_sem_divmod sign sz (signed wmod wmodi sign)
+Lemma sem_sop2_typed_total o :
```
eponier, line 87: idem

proofs/_CoqProject
```diff
@@ -205,12 +205,15 @@ lang/psem_facts.v
...
 lang/sem_one_varmap.v
-lang/sem_op_typed.v
+lang/sem_op_total.v
 lang/sem_one_varmap_facts.v
 lang/sem_params.v
 lang/sem_type.v
 lang/sha256.v
 lang/shift_kind.v
+lang/sopn_semi.v
+lang/op_semi.v
```
eponier, line 215: normally we just follow alphabetical order in this file

proofs/lang/sopn_semi.v
```diff
@@ -0,0 +1,709 @@
...
+Proof. by move=> h; elim: tin vs f => /= [vs f | t tin ih vs f v]; [apply h | apply ih]. Qed.
+
+(* Without condition, [mk_sem_op] is the total semantics. *)
+Lemma mk_sem_op_nil tin t err f :
+  sem_prod_eq tin (@mk_sem_op tin t [::] err f) (sem_prod_ok tin f).
+Proof. by apply: mk_semi_aux_id. Qed.
+
+(* [mk_sem_op] reduces on a concrete signature: this is what makes the
+   equality proofs provable by [reflexivity]. *)
+Lemma mk_sem_op_reduces_example :
```
eponier, line 705: Do we want to keep this? Maybe use Example of Fact instead of Lemma if this is case.

proofs/lang/sopn_semi.v
```diff
@@ -0,0 +1,709 @@
...
+  mapM2 ErrType truncate_val tin vs = ok vs' ->
+  ac_wt tin c ->
+  interp_acond vs' c = interp_acond vs c.
+Proof.
+move=> htr /eqP hwt; have := interp_acond_truncate htr hwt.
+case: (interp_acond vs c) => [v | e] //=.
+by move=> [v'' -> /truncate_val_typeE [b -> ->]].
+Qed.
+
+Lemma acond_b_truncate tin vs vs' c :
```
eponier, line 417: It would be better if the order of the hypotheses was consistent with the previous lemma.

proofs/lang/sem_op_typed.v
```diff
@@ -1,223 +1,268 @@
...
+apply: wint_range_eq; first by rewrite /= !truncate_word_u.
+have h2 : (0 <= int_of_word Unsigned w2)%Z.
++ by rewrite /int_of_word /signed; have := wunsigned_range w2; Lia.lia.
+rewrite /= /zlsl; case: ZleP => [_ | hlt];
+  last by have := wunsigned_range w2; Lia.lia.
+by rewrite /sem_shl /sem_shift wrepr_mul wrepr_int_of_word (wshl_sem _ h2)
+           GRing.mulrC.
+Qed.
 
+Lemma sem_sop2_typed_wishrE sg sz (w1 : word sz) (w2 : word U8) :
```
eponier, line 265: We prove in `wishr_eq` a simpler expression. Maybe it's better to keep the current lemma for uniformity, but maybe add a comment here pointing at `wishr_eq` for better discoverability.

eponier (commented):
I think I understood the motivation, but at some point in the review, I got lost in the definitions, so I may have missed something.

I don't like the name `sopn_semi.v`. Reading that name, I would guess that it includes `op_semi.v` but it's the other way around.

And should we keep the commit messages as they are? No strong opinion.

proofs/lang/sopn_semi.v
bgregoir: I think we can remove it

proofs/lang/sopn_semi.v
bgregoir: I will fix that

proofs/lang/sem_op_typed.v
bgregoir: I think the only interest of this lemma is to recovers the old semantic and to not break existing proof (or to be able to restore them easily).
Maybe we can remove this lemma and fix the only proof using it (but maybe this will require a bigger change in the proof). I will try.

proofs/compiler/constant_prop.v
bgregoir: No, I will try to fix that

bgregoir: In fact it is now necessary. There is two function app_sopn, one in sem_type the other in values (Definition app_sopn := (*sem_type.*)app_sopn of_val). Some changes in the exported files require to have the long name  now.

bgregoir: I will try to see if I can remove some "Require export values" to fix that.

proofs/lang/sem_op_typed.v
bgregoir: I will remove them

proofs/compiler/constant_prop.v
bgregoir: I think it's work.

bgregoir: It is fixed

proofs/lang/sem_op_typed.v
bgregoir: done

proofs/lang/sem_op_typed.v
bgregoir: done

proofs/lang/sem_op_typed.v
bgregoir: done

proofs/lang/sem_op_typed.v
bgregoir: done

proofs/lang/sopn_semi.v
bgregoir: done

proofs/lang/sopn_semi.v
bgregoir: done

proofs/lang/sem_op_typed.v
bgregoir: Most of the lemmas of that family have been removed

proofs/_CoqProject
bgregoir: done

bgregoir:
For the name of the files, I am also a little be lost (but this is already the case in main). In any case, it is something that can be easily changed.

PR #1570, opened 2026-09-16
Deprecate (int) cast without explicit signedness

### LINKED ISSUE #1568: Missing implicit sint→int conversion
When **unsigned** word-sized integers are used as array indices, they are *implicitly* converted to integers, as illustrated by the following function:

~~~
export fn with_ui() -> reg u8 {
  stack u8[8] data;
  data[:u64 0] = 0;
  reg ui64 idx = 1;
  reg u8 res = data[idx];
  return res;
}
~~~

However, **signed** word-sized integers do not enjoy this facility and the following program is (wrongly) rejected (with ““typing error: can not implicitly cast si64 into int”):

~~~
export fn with_si() -> reg u8 {
  stack u8[8] data;
  data[:u64 0] = 0;
  reg si64 idx = 1;
  reg u8 res = data[idx];
  return res;
}
~~~

A possible workaround is to add an explicit conversion: `data[(int)idx]`.

vbgl:
# Description

Small cleaning in pretyping, enabling the use of signed integers as array indices.

Fixes #1568.

# Checklist


changes/02-bugfix/1569-sint-conversion.md
```diff
@@ -0,0 +1,5 @@
+- The cast operator `(int)` has been deprecated. Use `(sint)`, `(uint)`, or
+  nothing instead. Signed word-sized integers can be used as array indices.
```
bgregoir, line 2: unsigned word sized integers cannot be used as array indices ? Or it was already possible before ?

vbgl: See #1568.

bgregoir: You are right and furthermore we discuss of this before. My point is that I found the changelog confusing. 
I will prefers "Signed/Unsigned word-sized integers can be used as array indices". But it is upto you.

bgregoir (approved):

bgregoir:
Also we have nothing in the documentation that say that for array indices, cast to int is automatically inserted. If we have we should say that this work for word/signed word/ unsigned word.

bgregoir:
I have my answer no documentation on that.

eponier:
Testing
```
  reg si64 w = 0;
  stack u64[2] s;
  s[0] = 0;
  reg u64 res = s[(uint)w];
  return res;
}
```
gives
```
line 5 (24-25):
typing error: can not implicitly cast si64 into u64
```
Before it was
```
typing error: can not implicitly cast si64 into int
```
which was admittedly confusing, but are we happy with the new one?

vbgl:
A correct fix is to add an explicit cast to `u64`: `s[(uint)(64u)w]`.

